PR #48, opened 2015-05-29
CBA UI - params support in code exec

SzwedzikPL:
Support for params in cba ui with backwards compatibility

Example:
```
[
    ["Menu Caption", "flexiMenu resource dialog", "optional icon folder", menuStayOpenUponSelect],
    [
        [
            "caption",
            [name player, {hint format ["my name is %1", _this];}],
            "icon",
            "tooltip",
            {"submenu"|["menuName", "", {0/1}]},
            -1,
            {0|1/"0"|"1"/false|true},
            {-1|0|1/"-1"|"0"|"1"/false|true}]
    ]
]
```

### DIFF
--- a/addons/ui/flexiMenu/fnc_execute.sqf
+++ b/addons/ui/flexiMenu/fnc_execute.sqf
@@ -4,11 +4,14 @@
 //-----------------------------------------------------------------------------
 #include "\x\cba\addons\ui\script_component.hpp"
 
-private["_action", "_subMenu", "_multiReselect", "_useListBox", "_subMenuSource", "_params", "_pathName"];
+private["_arrayID", "_actionData", "_action", "_subMenu", "_multiReselect", "_useListBox", "_subMenuSource", "_params", "_pathName", "_actionCode", "_actionParams"];
 
-_action = _this select 0;
-_subMenu = _this select 1;
-_multiReselect = _this select 2;
+_arrayID = _this;
+_actionData = GVAR(menuActionData) select _arrayID;
+
+_action = _actionData select 0;
+_subMenu = _actionData select 1;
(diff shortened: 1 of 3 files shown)


Dr-Eyeball:
This task needs further clarification.
Where are the param's required to be used/applied, just the **"code"** line?

* AFAIK, param's already work with the 2nd "code" line, since all code lines are compiled and executed at the time of selection of the menu option. 
 * (So even if the player name, or the variable used in the code line, should change in the meantime while the menu is displayed, it will already use the latest name/variable value at the time of execution.)

Eg: |
--- |
{hint format ["my name is %1", name player}, |
{hint format ["my name is %1", TAG_playerNameVar}, |

Unless you are requesting for **real-time processing of parameters,** for real-time refreshing, so it:
* refreshes the menu option captions in real time, even after the menu is drawn the first time.
* refreshes the menu title caption in real time.
* real-time enabling/disabling of menu options.
* real-time visibility of menu options.
 * (This could be tricky, unless maybe it uses the "-1" style of skipping hidden options, which might be cleaner.)

SzwedzikPL:
It was made to add support for local variables in actions. It's "must have" for me now when im using scripts to generate "dynamic" menu for some objects. Now script can pass params to action like this:
```
[[_paramOne, _paramTwo], myFnc]
```

Real-time processing of parameters would be nice also :) It can solve problem when some option which was active when menu is opened shouldnt be active i.e. after 5 sec becouse the situation has changed. For now im solving this with closing and reopening menu but its a bit not comfortable.

Dr-Eyeball:
I didn't even know local var's could continue to be used outside of their scope like that. Interesting.
I guess they must continue to persist due to their remaining reference. Thanks for info.

PR #53, opened 2015-05-31
Restore XEH fired EH for Heli_Attack_01_base_F

Killswitch00:
Enable XEH for Helicopter_Base_F and disable its fired EH, just like the original config. Also, make sure the descendant Heli_Attack_01_base_F class has the stock BIS fired EH restored.


ViperMaul:
@jaynus please test with ACE Helos

ViperMaul:
The configs looks good though so far.

PR #55, opened 2015-06-01
Restore XEH init EH for the BLUFOR FIA units.

Killswitch00:
The west FIA units were missing the XEH init event handler. This caused problems such as missing self-interaction functionality in combination with ACE.


ViperMaul:
Hotfix for testing v1.1.22.150601 - http://bit.ly/1FpOy2Y

PR #58, opened 2015-06-03
Created optional addon to enable an automatic 'add XEH to addons that…

Killswitch00:
This lets users have the old, pre-RC7 behaviour back where XEH will add XEH event processing to units and vehicles that do not have it out of the box.


ViperMaul:
:+1: 
Worked like beloved magic for me. 
Tested with [PG_Services Mod] (http://www.armaholic.com/page.php?id=21616)
And [EuroForce Mod](http://www.armaholic.com/page.php?id=28004)

ViperMaul:
For testing purposes http://bit.ly/1FYrfzE

reoxdk:
ive just testet it with DDAM on our server and it works there thanks great job

PR #59, opened 2015-06-04
Added LSTRING and CSTRING with External variants Macros

jonpas:
As discussed with @ViperMaul this adds newly added Macros from ACE3 to CBA, with `#ifndef` for compatibility.

@ViperMaul should I be adding `#ifndef` on ACE3's side as well?

### DIFF
--- a/addons/main/script_macros_common.hpp
+++ b/addons/main/script_macros_common.hpp
@@ -953,7 +953,7 @@ Author:
     Jonpas
 ------------------------------------------- */
 #ifndef STRING_MACROS_GUARD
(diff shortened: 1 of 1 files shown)


Glowbal:
Could instead use a single guard.

```sqf
#ifndef STRING_MACROS_GUARD
#def STRING_MACROS_GUARD

// define macros here

#endif
```

jonpas:
That's pretty slick, thank you!

ViperMaul:
:+1:

jonpas:
@ViperMaul this also goes into 2.0. :)

PR #60, opened 2015-06-05
Replaced set count for pushback

dixon13:
As title says...


Killswitch00:
Looks OK to be merged.

PR #62, opened 2015-06-05
Changed set to pushBack

dixon13:
Both are doing the same thing in this function but pushBack is faster.


Killswitch00:
That will not work if CBA_fnc_filter is asked to do an in-place filtering of the input array.

esteldunedain:
The perf overhead of set only happens if the array is resized, which is not the case here (the size is already set at the beginning). Therefor this patch makes no sense.

Issue #65, opened 2015-06-08
All 17 BIS C_Hatchback_ class cars overwrite XEH Init


ViperMaul:
All 17 BIS C_Hatchback_ class cars overwrite XEH Init
C_Hatchback_01_F and the other 16 seen in this reference.
Reference: http://gyazo.com/4f0995c9ec7d32963f23b93b05f4cab2

NOTE: 
Using the optional **cba_enable_auto_xeh.pbo** does solve this issue.

PabstMirror:
```
diag_log text "-----------------------------";
x3 = configProperties [(configfile >> "CfgVehicles"), "(configName _x) isKindOf 'Land'"];
{
    _initConfig = configProperties [(_x), "(configName _x) == 'EventHandlers'", false];
    x4 =_initConfig;
    if ((count _initConfig) == 1) then {
        _initEH = getText ((_initConfig select 0) >> "init");
        if (_initEH != "_this call SLX_XEH_EH_Init") then {
            diag_log format ["%1 = %2", (configName _x), _initEH];
            systemChat format ["%1 = %2", (configName _x), _initEH];
        };
    };
} forEach x3;
```
This code should dump all of the problem entries (just the parents):
```
"VirtualMan_F = _this call bis_fnc_initVirtualUnit;"
"I_G_Story_Protagonist_F = "
"I_G_resistanceLeader_F = "
"Static_Designator_01_base_F = (_this select 0) execVM "\A3\Static_F_Mark\Designator_01\Scripts\init.sqf""
"Static_Designator_02_base_F = (_this select 0) execVM "\A3\Static_F_Mark\Designator_02\Scripts\init.sqf""
"MBT_03_base_F = if (local (_this select 0)) then {{(_this select 0) animate [_x,(random 1)]} forEach ['HideHull','HideTurret']}"
"C_Offroad_01_F = if (local (_this select 0)) then {[(_this select 0), "", [], true] call bis_fnc_initVehicle;};"
"Offroad_01_repair_base_F = if (local (_this select 0)) then {[(_this select 0), "", [], true] call bis_fnc_initVehicle;};"
"B_G_Offroad_01_repair_F = if (local (_this select 0)) then {[(_this select 0), "", [], true] call bis_fnc_initVehicle;};"
"I_G_Offroad_01_F = if (local (_this select 0)) then {[(_this select 0), "", [], true] call bis_fnc_initVehicle;};"
[...]

PR #70, opened 2015-06-10
Enable XEH for a few more BIS units and vehicles

Killswitch00:
Fixes the vehicles that PabstMirror reported (https://github.com/CBATeam/CBA_A3/issues/65#issuecomment-110485462)


ViperMaul:
:+1:   Looks good to go!

PR #78, opened 2015-06-19
Restore the CBA keybinding list in the briefing. Fixes #44

### LINKED ISSUE #44: CBA Keybindings List in Briefing
Restore the Keybindings list after keybind rewrite.

Killswitch00:
This will display all key binds that use the CBA key binding system in the briefing, under the CBA->Keybindings diary record section.

### DIFF
--- a/addons/help/XEH_postClientInit.sqf
+++ b/addons/help/XEH_postClientInit.sqf
@@ -20,10 +20,10 @@ _pkeynam = {
     _shift = if(_shift) then {42} else {0};
     _ctrl = if(_ctrl) then {56} else {0};
     _alt = if(_alt) then {29} else {0};
-    _keys = [_shift,_ctrl,_alt,_key];
+    _keys = [_shift,_ctrl,_alt,_dikKey];
     _keystrg = "^";
     {
-        _knaml = call compile format["format['%2',%1]",(keyName _x),"%1"];
+        _knaml = [cba_keybinding_dikDecToStringTable, format ["%1", _x]] call BIS_fnc_getFromPairs;
         _knaml = [_knaml, " "] call (uiNamespace getVariable "CBA_fnc_split");
         _knam = "^";
         {_k = _x; _knam = _knam + " " + _k} forEach _knaml;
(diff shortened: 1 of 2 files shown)


ViperMaul:
Looks like a great start. 
Currently it is hard to differentiate between the keybinds and the feature. 
http://gyazo.com/9abde0f5f090f029996bc9949e34536f

Perhaps we can give the keybinds some color. Or some kind of contrasting solution.

ViperMaul:
The different colors look great! 

Two opportunities for improvement. 
1. Keybind Combo Display 
  When you scan the keybinds fast and most people do, it is the spaces that normally separates a key combo. And dashes and pluses that combines a single key combo. 
  So ```Left Shift-Left Alt-+ [NUM]``` looks like 4 keys instead 3 keys. 

2. The correct key is + key but, which plus key, well the NUM-+ key. However since the + is displayed before the [NUM] it loses that meaning. 
Look at this


Compared to this



One thought would be to just 

1. Abbreviate the Left and Right with no space in between.
so Left Shift would be come LShift and Right Shift would become RShift. Now it looks like one key instead of two separate keys at a quick glance. 

2. Same with NUM pad handling it should be shown as Num+ or [Num+] as this is closer to the defacto standard way of representing this key.

3. Remove the dash as the key separator and just since we employed the above. However keys like Caps Lock will still look like two keys even though most players will understand it is one key. Example:   LAlt Caps Lock

4. We could experiment what it looks like to enclose each key inside { } 
Example: { LAlt }{ Caps Lock }
or 
Example { RShift }{ [ } or would that be { RShift }{ LBracket }

We have some experimenting to do.

Killswitch00:
Let's try dba7707 on for size and see how that style feels.

ViperMaul:
Looks much improved. 
The next thought to mind is the LAlt LShift etc looks wordy and redundant. 
I think it would now look cleaner if we Omit the L for all left SHIFT, CTRL, ALT keys. But continue to use the rare RShift, RCTRL, RAlt 

Reference:

ViperMaul:
Looks much cleaner. 
image

ViperMaul:
:+1:

PR #81, opened 2015-06-21
Key combo rewrite - Phase 1

ViperMaul:
- [x] Adding the ability to SET a single modifier key as a keybind. Example RSHIFT, RCTRL, RALT
- [ ] Adding the ability to SET a modifier key within a key combo. Example RCTRL+LSHIFT+P
- [x] Adding the ability to SET a numpad key
- [ ] Adding the ability to USE a single modifier key as a keybind
- [x] Adding the ability to USE a modifier key within a key combo
- [x] Adding the ability to USE a numpad key


ViperMaul:
Issues remain, the ability to set RCTRL works but is currently ignored when attempting to activate the code assigned to it. Will look at that in my next session.

PR #91, opened 2015-07-06
New CBA_fnc_randPosArea

SilentSpike:
Adds a new function (CBA_fnc_randPosArea) to return a random position within a given marker/trigger:

- Supports ellipses and rectangles (and rotation)
- Distribution should be uniform
- Optional boolen parameter can be used to return only positions on the perimeter
- Returns empty array if non-area is passed in (either "ICON" marker or a non-trigger object)

### DIFF
--- a/addons/common/CfgFunctions.hpp
+++ b/addons/common/CfgFunctions.hpp
@@ -354,8 +354,14 @@ class CfgFunctions
             // CBA_fnc_randPos
             class randPos
             {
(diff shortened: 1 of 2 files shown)


SilentSpike:
There we go, that's it finalized.

I took cue from CBA_fnc_inArea, but I notice that it will throw errors if a non-trigger object is passed. I can copy the bad input handling from this new function over to there if desired.

PR #93, opened 2015-07-08
Updating EXPLODE macros with 1.48 commands

SilentSpike:
Unfortunately the new commands don't quite match up with the way that the CBA macros are structured so I had to redefine both the `EXPLODE_N_SYS` macros and the `EXPLODE_N_PVT` macros.

This should give a retroactive performance upgrade to all `EXPLODE` and `PARAMS` macros.


SilentSpike:
Updated `DEFAULT_PARAM` to also make use of the new commands.

SilentSpike:
Also updated `CBA_fnc_defaultParam` to match the macro

jonpas:
:+1:

PR #99, opened 2015-07-27
Fixed broken macro

SilentSpike:
Found this one thanks to szwedzikpl in slack. All broken because of a typo :cry:


ViperMaul:
+1

PR #101, opened 2015-07-27
Simplifying the "simplify angle" functions

SilentSpike:
Removes while loops from the simplify angle functions (bad for unscheduled environment usage) and replaces with pure maths (also results in less operations)


SilentSpike:
Noticed my maths in the simplifyAngle180 function is off, fixing now

SilentSpike:
Fixed :smile:

PR #102, opened 2015-08-03
Optimize String Functions/add "Real" subString

jokoho48:
Add "Real" substring (CBA_fnc_substring_new)
Improve Performance in capitalize,letfTrim,rightTrim,substring

### DIFF
--- a/addons/strings/fnc_substring.sqf
+++ b/addons/strings/fnc_substring.sqf
@@ -2,7 +2,7 @@
 Function: CBA_fnc_substring
 
 Description:
-    Extracts the index-based substring from a string.
+    Extracts the index-based substring from a string.(deprec)
 
 Parameters:
     _string - String to make replacement in [String]
@@ -20,6 +20,7 @@ Example:
 
 Author:
     jaynus
+    joko // Jonas
 --------------------------------------------------------------------------- */
 
 #include "script_component.hpp"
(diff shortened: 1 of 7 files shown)


addons/strings/fnc_substring.sqf
```diff
@@ -28,18 +29,12 @@ SCRIPT(substring);
...
-private ["_arr", "_ret", "_i", "_x"];
-_arr = toArray _string;
-_ret = [];
+params["_string","_startIndex","_endIndex"];
+// Check if _start is Larger than _endIndex to Prevent Issues
+if (_startIndex > _endIndex) exitWith {""};
+// Calculate Differenz between _start and _end for select lenth value
+_endIndex = _endIndex + 1 - _startIndex;
+// Cut out String
+_ret = _string select [_startIndex, _endIndex];
```
SilentSpike: You can cut out the middleman here and just return `_string select [_startIndex, _endIndex]`

Killswitch00:
Please split this into two separate pull requests - one for fixing existing functions. The other can suggest a new string funcion, including the reasoning why the new one is "real".

jokoho48:
Why its "real" is because substring normaly work with start Index and lenth and not end index. as Example Programming language you can bring here C#, SQL, XML, C++

PR #109, opened 2015-08-07
Removing depreciated PARAMS macros

SilentSpike:
Replaces all PARAMS_N macros with the new `params` command.
Also condenses some appropriate cases of DEFAULT_PARAM and replaces the single instance of EXPLODE_N_PVT.

This was all done with regex and I don't have a test build setup at the moment so it's possible that there could be some minor syntax errors in here. However they should be few and far between if at all. Suggest running dev build when merged to check for any script errors.


ViperMaul:
@SilentSpike, @Killswitch00:

My first test of this PR with ACE yielded an error divide by zero message in CBA hashGet function. 

<pre>
11:59:35 Error in expression <f (_index >= 0) then
{
(_hash select 2) select _index; 
} else {
if (isNil "_def>
11:59:35   Error position: <select _index; 
} else {
if (isNil "_def>
11:59:35   Error Zero divisor
11:59:35 File x\cba\addons\hashes\fnc_hashGet.sqf, line 15
</pre>

References:
[Source L30-L37](https://github.com/SilentSpike/CBA_A3/blob/c6f93afc4824c5b90d198b90efe13f5c16ca73b2/addons/hashes/fnc_hashGet.sqf#L30-L37)
[commited change](https://github.com/SilentSpike/CBA_A3/commit/c6f93afc4824c5b90d198b90efe13f5c16ca73b2)

However running Unit Tests:
[] call CBA_fnc_testHash  yielded no errors. 

ACE Interaction menus do not work with this CBA version. I suspect it is due to the error above. 
Switching back to the CBA master branch, I get no errors and the ACE Interaction menu is working again. 

It looks like we have to be careful about merging this. 
Must investigate.

SilentSpike:
:+1: Will check it out, thanks for testing. Is the CBA dev environment just the same as the ace one? I should really set up my own proper one sometime

ViperMaul:
Yes, retty much; With the exception that the PboPrefix is 'x' instead of 'z'

SilentSpike:
Ah, I see why. It's because of this:

`params ["_hash","_key", ["_default",_hash select HASH_DEFAULT_VALUE]];`

The default value is referencing one of the other values in the same command. There might be other cases of this, will need to be updated so that the default value is an actual value and then an if statement sets it to `_hash select HASH_DEFAULT_VALUE` if it's the default value.

SilentSpike:
hashGet should be fixed, I'll do a quick sweep of all the params that include default values to check for command self-reference

ViperMaul:
We should update the Unit Test for Hashes to include this case, I would think.

SilentSpike:
I'm note entirely sure what those are ;)

All other instances of that problem in this commit are now fixed.

jonpas:
> Is the CBA dev environment just the same as the ace one?

As @ViperMaul said, I even ported setup.py and build.py and with @ViperMaul updated make.py so it's exactly the same with the exception of different prefix. :banana: for you! :grin:

ViperMaul:
:+1:

PR #125, opened 2015-08-17
Fix existing AI functions

SilentSpike:
- Add waypoint had an extra meaningless parameter (house position) that didn't match it's header
- taskPatrol added one too many waypoints

### DIFF
--- a/addons/ai/fnc_addWaypoint.sqf
+++ b/addons/ai/fnc_addWaypoint.sqf
@@ -32,32 +32,32 @@ Author:
 
 ---------------------------------------------------------------------------- */
 #include "script_component.hpp"
-private ["_group","_position"];
-_group = (_this select 0) call CBA_fnc_getgroup;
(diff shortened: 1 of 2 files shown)


Killswitch00:
What if the extra house position is just undocumented? If we remove that and someone out there is actually using it...

SilentSpike:
@Killswitch00: The waypoint is never attached to a house though, so it doesn't do anything :stuck_out_tongue:

Killswitch00:
Interesting thing, this fnc_addWaypoint. As you say, it was never actually possible to use this function to set a waypoint to an object (such as a house) since an object input parameter is converted to a position by the call to `CBA_fnc_getPos`.

Further, if people have actually provided a waypoint completion radius to the function according to the documentation, that has never worked (a call will be made to `setWaypointHousePosition` with what the user thinks is the WP completion radius)

The fix you suggest would correct the completion radius problem and thus make the function work as documented.

I'll buy that. :+1:

PR #131, opened 2015-08-25
Add 4 New Array Functions

jokoho48:
Add 4 New Array Functions to Find First Entry of TypeOf, TypeName, Null or Nil in Array


Killswitch00:
Since the "find" script command finds the **first** array element or string-within-string, these function names could be shortened somewhat. Also, "empty" in BIS scripting parlance is "nil". How about calling them

 * `CBA_fnc_findNil` finds the first _nil_ entry
 * `CBA_fnc_findNull` finds the first _null_ entry
 * `CBA_fnc_findTypeOf` finds the first entry of a certain type.

MikeMatrix:
I like @Killswitch00's names a lot more. They're very similar, to the point and short to type.

jokoho48:
is everything fine?

Killswitch00:
The unit tests for the three new functios are missing and the added unit test file could be renamed to, say, "test_find.sqf".

jokoho48:
my question is did i need a test function? or not

addons/arrays/test_find.sqf
```diff
@@ -0,0 +1,185 @@
...
+TEST_OP(str _original,==,str _expected,_fn);
+TEST_OP(str _result,==,str _expected,_fn);
+
+
+
+
+_fn = "CBA_fnc_findTypOf";
+LOG("Testing " + _fn);
+
+TEST_DEFINED("CBA_fnc_findTypOf","");
```
Killswitch00: I can see that this unit test file hasn't been tested at all. Run it using
```
[] call compile preprocessFileLineNumbers "\x\cba\addons\arrays\test_find.sqf";
```

addons/arrays/test_find.sqf
```diff
@@ -0,0 +1,185 @@
...
+TEST_DEFINED("CBA_fnc_findNil","");
+
+// Find First Nil Entry
+_original = ["", player, "", player, nil, "", player, nil];
+_result = ["", player, "", player, nil, "", player, nil] call CBA_fnc_findNil;
+_expected = 4;
+TEST_OP(str _original,==,str _expected,_fn);
+TEST_OP(str _result,==,str _expected,_fn);
+
+
```
Killswitch00: The three tests below all do the same thing as the first test above - look for the first literal `nil` element in the test arrays. Keep the first one, and change the three redundant tests to use
 * an array with a variable in it somewhere that is nil
 * an array with a code element in it that returns nil
 * an empty input array ([])

addons/arrays/test_find.sqf
```diff
@@ -0,0 +1,185 @@
...
+_original = ["", player, "", player, "", "", player, displayNull];
+_result = ["", player, "", player, "", "", player, displayNull] call CBA_fnc_findNull;
+_expected = 7;
+TEST_OP(str _original,==,str _expected,_fn);
+TEST_OP(str _result,==,str _expected,_fn);
+
```
Killswitch00: Good - these four tests above test for different kinds of null values (but a test with controlNull is missing). It also needs tests for
* an input array with a variable in it somewhere that is null-valued (for example objNull).
* an input array with a code element in it that returns one of the null values.
* an input array that is empty.

addons/arrays/test_findNil.sqf
```diff
@@ -0,0 +1,41 @@
...
+// ----------------------------------------------------------------------------
+
+private ["_expected", "_result", "_fn"];
+
+_fn = "CBA_fnc_findNil";
+LOG("Testing " + _fn);
+
+TEST_DEFINED("CBA_fnc_findNil","");
+
+// Use of embeded nil
```
Killswitch00: "embedded"

addons/arrays/test_findNull.sqf
```diff
@@ -0,0 +1,41 @@
...
+LOG("Testing " + _fn);
+
+TEST_DEFINED("CBA_fnc_findNull","");
+
+// Use of embedded null
+_result = ["", objNull, player, 3] call CBA_fnc_findNull;
+_expected = 1;
+TEST_OP(_result,==,_expected,_fn);
+
+// Only find first Null
```
Killswitch00: "null", not "Null"

addons/arrays/fnc_findTypeOf.sqf
```diff
@@ -0,0 +1,43 @@
...
+---------------------------------------------------------------------------- */
+#include "script_component.hpp"
+
+scopeName "main";
+
+params [["_array", [], [[]]], ["_typeOf", nil, [objNull, ""]]];
+
+if (isNil "_typeOf" || {_array isEqualTo []}) exitWith {-1};
+
+if (IS_OBJECT(_typeOf) then {
```
Killswitch00: I see that this hasn't been tested in-game at all.

addons/arrays/fnc_findTypeName.sqf
```diff
@@ -0,0 +1,48 @@
+/* ----------------------------------------------------------------------------
+Function: CBA_fnc_findTypeName
+
+Description:
+    A function that returns the index of the first entry of the given type in an array.
+
+Parameters:
+    0: Array
+    1: TypeName, if parameter is a string, that contains a case insensitive typename, it will be used. Otherwies typename of the variable will be used.
```
Killswitch00: "Otherwise"

MikeMatrix: God dammit :D

addons/arrays/test_findTypeOf.sqf
```diff
@@ -0,0 +1,61 @@
...
+_result = [["", player, 5], player] call CBA_fnc_findTypeOf;
+_expected = 1;
+TEST_OP(_result,==,_expected,_fn);
+
+// Find first only
+_result = [["", player, 5, objNull, player], typeOf player] call CBA_fnc_findTypeOf;
+_expected = 1;
+TEST_OP(_result,==,_expected,_fn);
+
+// Finf first only, mixed solutions in array
```
Killswitch00: "Finf" --> "Find"

addons/arrays/test_findTypeOf.sqf
```diff
@@ -0,0 +1,61 @@
...
+_result = [["", typeOf player, 5, objNull, player], typeOf player] call CBA_fnc_findTypeOf;
+_expected = 1;
+TEST_OP(_result,==,_expected,_fn);
+
+// Return not found, when no matching elemnt is in array
```
Killswitch00: "element"

addons/arrays/fnc_findTypeName.sqf
```diff
@@ -0,0 +1,48 @@
...
+
+#define TYPENAMES ["ARRAY", "BOOL", "CODE", "CONFIG", "CONTROL", "DISPLAY", "GROUP", "LOCATION", "OBJECT", "SCALAR", "SCRIPT", "SIDE", "STRING", "TASK", "TEXT", "TEAM_MEMBER", "NAMESPACE"]
+
+scopeName "main";
+
+params [["_array", [], [[]]], "_typeName"];
+
+if (isNil "_typeName" || {_array isEqualTo []}) exitWith {-1};
+
+// If a string is given, tansform to uppercase for type matching
```
Killswitch00: "transform"

ViperMaul: A better word for "transform" in this context is to use "convert"

PR #134, opened 2015-08-27
Wrong return value for separator strings longer than the input string…

### LINKED ISSUE #133: CBA_fnc_split - wrong return
Returns the wrong value if the seperator is longer than the String.

[Line 41](../blob/master/addons/strings/fnc_split.sqf#L41)
Is:
`exitwith {toArray _input};`

should be:
`exitwith {[_input]};`

test with:
`["a","test"] call CBA_fnc_split;`
returns `[97]`instead of `["a"]`

Killswitch00:
…. Fixes #133

### DIFF
--- a/addons/strings/fnc_split.sqf
+++ b/addons/strings/fnc_split.sqf
@@ -31,40 +31,6 @@ SCRIPT(split);
 
 // ----------------------------------------------------------------------------
 
-private ["_split", "_index", "_inputCount", "_separatorCount", "_find", "_lastWasSeperator"];
-params ["_input", ["_separator",""]];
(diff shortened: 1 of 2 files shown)


Killswitch00:
Note to self: the suggested refactoring of CBA_fnc_split won't work. See https://github.com/CBATeam/CBA_A3/issues/133#issuecomment-135393761

Killswitch00:
A simple conversion using splitString is not possible due to differences in how the separator argument is interpreted - for CBA_fnc_split, it is a separator string. For splitString, it is a set of separator characters. PR will be abandoned.

PR #139, opened 2015-08-27
Improved logging system

MikeMatrix:
Improved logging system with a more flexible approach to handling when to log, how the log looks and what to log.
This can be used in pretty much any situation and can be easily expanded by either internal code or third party code.

The intention of this is to standardize logging not only for RPT logs, make it readable and add enough metadata to be useful in any situation.

As it stands right now, the code allows you to dynamically hook log writers, that will actually perform the output. This means you can easily create a logger that outputs a formatted log to the RPT as well as systemChat, on every client's RPT, etc.
At this point it only allows you to put copies of the function in, which will be expanded into being able to reference functions with Strings, so you can call them directly instead of passing in a copy of them.

I feel the logging format right now seems acceptable, but it is still being worked on and tweaked for better readability.

Still WIP.

### DIFF
--- a/addons/log/config.cpp
+++ b/addons/log/config.cpp
@@ -1,0 +1,15 @@
+#include "script_component.hpp"
+
+class CfgPatches {
+    class ADDON {
+        units[] = {};
+        weapons[] = {};
+        requiredVersion = REQUIRED_VERSION;
+        requiredAddons[] = { "cba_common" };
+        version = VERSION;
+        author[] = { "MikeMatrix" };
+        authorUrl = "http://dev-heaven.net/projects/cca";
+    };
+};
+
+#include "CfgFunctions.hpp"
--- a/addons/diagnostic/CfgFunctions.hpp
+++ b/addons/diagnostic/CfgFunctions.hpp
@@ -45,6 +45,42 @@ class CfgFunctions
                 description = "Runs unit tests for an addon or component.";
                 file = "\x\cba\addons\diagnostic\fnc_test.sqf";
             };
+            // CBA_fnc_addLogLevel
+            class addLogLevel
+            {
+                description = "Add a system wide named log level.";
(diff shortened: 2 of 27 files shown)


addons/log/config.cpp
```diff
@@ -0,0 +1,15 @@
...
+
+class CfgPatches {
+    class ADDON {
+        units[] = {};
+        weapons[] = {};
+        requiredVersion = REQUIRED_VERSION;
+        requiredAddons[] = { "cba_common" };
+        version = VERSION;
+        author[] = { "MikeMatrix" };
+        authorUrl = "http://dev-heaven.net/projects/cca";
```
Killswitch00: The project URL is https://github.com/CBATeam/CBA_A3 nowadays.

MikeMatrix:
Will change that with the next changes. Still have some improvements to do.
I'll probably add a log level stack instead, so you can temporary force a different log level for a specific part without losing the previous log level.

addons/log/config.cpp
MikeMatrix: Will change that with the next changes.

PabstMirror:
Problem with this kind of dynamic logging is it turns macros that would normally be compiled out into function calls.

`TRACE_1("test",_keys)`
becomes 
```
["cba", "hashes", format ['%1:  _keys=%2', "test", if (isNil{ _keys}) then {nil} else { _keys}], 10, ["CBA_fnc_diagLogWriter"], "x\cba\addons\hashes\fnc_hashCreate.sqf", 42] call CBA_fnc_logDynamic;
```
instead of just becoming `/* disabled */`- and having no effect.

We gain the advantage of allowing someone to debug, without needing to have raw files, or mess with filepatching; but there is going to be a performance cost.  We can still add dynamic debuing, but I think they need to be on new macros and not modify the old ones.

PR #153, opened 2015-09-08
Fix XEH EventHandler inheritance for chickens

PabstMirror:
RPT: Cannot delete class EventHandlers, it is referenced somewhere (used
as a base class probably).


jonpas:
Nice find! :+1:

PR #155, opened 2015-09-14
Fix Keybinding Issue with Controllers

### LINKED ISSUE #113: Keybinding: Unknown Code 327778
#### This was reported to me in a Skype PM but they do not know how to reproduce it. 

2015-07-02_00001_2_zpsoga01hvs

#### If you have seen this before, perhaps you can give us a reproducible case for this so we can fix it.

jokoho48:
3 Bugs fixed in this PR
1. Private issues
2. Unkown Key by Controller Inputs(for button 1 to 10 used by normal xbox controller + digi kross)
3. deleting Keys by having a controller plugedin reason for that is that 2 onKeyDown events called perframe and overwrite its selfe and brake the system
4. add a Ignore Keys for ignoring Key that not useable(controll axis at the controller because xinput need a 25% dead zone that is not fixable without BI return the value how much a controller Float button is pressed)
5. some params rewrite
fixes #113


addons/keybinding/script_component.hpp
```diff
@@ -2,7 +2,7 @@
 #include "\x\cba\addons\main\script_mod.hpp"
 #include "\x\cba\addons\main\script_macros.hpp"
 
-//#define DEBUG_ENABLED_KEYBINDING
+#define DEBUG_ENABLED_KEYBINDING
```
Killswitch00: Debug should not be enabled in release builds

jokoho48:
@Killswitch00 fixed

jokoho48:
@ViperMaul Milestone missing

PR #165, opened 2015-10-06
Add removal of specific backpack and item from cargo functions

jonpas:
This PR adds 2 (4 including global variants) new functions:
- `removeBackpackCargo(Global)` - removes a specific type and number of backpack(s) from cargo space
- `removeItemCargo(Global)` - removes a specific type and number of item(s) from cargo space 

There are already `removeMagazineCargo(Global)` and `removeWeaponCargo(Global)` functions, I will be updating those with new commands as well in a separate PR.

### DIFF
--- a/addons/common/CfgFunctions.hpp
+++ b/addons/common/CfgFunctions.hpp
@@ -372,25 +372,25 @@ class CfgFunctions
             // CBA_fnc_removeBackpackCargo
             class removeBackpackCargo
             {
-                description = "Function to remove specific backpacks from local cargo space.";
+                description = "Removes specific backpack(s) from local cargo space.";
                 file = "\x\cba\addons\common\fnc_removeBackpackCargo.sqf";
             };
             // CBA_fnc_removeBackpackCargoGlobal
             class removeBackpackCargoGlobal
             {
-                description = "Function to remove specific backpacks from global cargo space. MP synchronized.";
+                description = "Remove specific backpack(s) from global cargo space. MP synchronized.";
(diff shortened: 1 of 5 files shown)


PabstMirror:
You are not doing anything, I just want to point out a limitation of the game engine with backpacks.

Backpacks are containers, so removing all and re-adding by classname will lose their inventory information.

E.G.  Container with 2 packs: "A" with a bandage and "B" empty.  You remove backpack B, and A will be readded empty.

Also, some backpacks come with cargo pre-added (`"B_AssaultPack_mcamo_AT"`) (same BS as those weapons with attachments).  You re-add those backpacks and they come back full of AT rockets.

Command is great for running at init, care needs to be used when running mid-mission.

jonpas:
Very good point. I can add support for non-empty backpacks, but not sure about the backpacks with pre-added contents.

PabstMirror:
`CBA_fnc_removeWeaponCargoGlobal` has the same problem with restoring/removing weapon attachments

`CBA_fnc_removeMagazineCargoGlobal` will reset ammo count on all magazines (also resets magazine unique ID)

I wouldn't worry at all, maybe just add a note to header.

jonpas:
Sounds good, I'll do that.

jonpas:
Status on this and #167 ?

Killswitch00:
I've been busy and haven't had time to review the functions properly. I'm glad to see the activity in here. Will try to review them soon(tm).

jonpas:
Cheers, was just looking for an update, no rush. :)

And indeed, there are so many things I want to do, finding the time is the hard one though.

Issue #169, opened 2015-10-09
removePerframeHandler: Undefined Variable in expression: _x


ViperMaul:
<pre> 6:01:17 Error in expression <ewHandle", "_return"];
_return = false;
_x params ["", "", "", "", "", "_publicH>
 6:01:17   Error position: <_x params ["", "", "", "", "", "_publicH>
 6:01:17   Error Undefined variable in expression: _x
 6:01:17 File x\cba\addons\common\fnc_removePerFrameHandler.sqf, line 17</pre>
Reported by the CityLifeRPG.com dev team: ArMaTeC

ViperMaul:
> I think its this http://pastebin.com/36xK8cZb  -ArMaTeC

jokoho48:
found it posible open

ViperMaul:
> I can reproduce error with https://gist.github.com/PabstMirror/ef3e12712269ec63d989 - pabstmirror

PR #170, opened 2015-10-09
posible fix for #169

jokoho48:
posible fix for #169

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,6 +16,3 @@ Thumbs.db
 
 *~
 *#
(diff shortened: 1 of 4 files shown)


.gitignore
```diff
@@ -15,4 +15,4 @@ Thumbs.db
 /addons/CBA_Project.sublime-workspace
 
 *~
-*#
+*#
```
jonpas: I thought you used Atom, newlines! :banana:

PR #171, opened 2015-10-09
fix RTP issue in PFH main Loop

jokoho48:


addons/common/fnc_removePerFrameHandler.sqf
```diff
@@ -26,7 +26,9 @@ Author:
 
 params ["_publicHandle"];
 private "_handle";
-if (isNil "_publicHandle" || (_publicHandle < 0)) exitWith {}; // Nil/no handle, nil action
+if ((isNil "_publicHandle" || (_publicHandle < 0)) && {(count GVAR(PFHhandles)) < _publicHandle}) exitWith { // Nil/no handle, nil action or handle is larger than Public Handle Array
+    WARNING("PFH ID allready Removed or handle Not Exist or handle is smaller than 0");
```
jonpas: PFH ID already removed or handle does not exist or handle is smaller than 0

MikeMatrix: I'd go with `Invalid or not existing PFH ID.`

jonpas: That's even better.

jokoho48:
ok i and @jonpas tested it no issues found

jonpas:
:+1:

addons/common/fnc_removePerFrameHandler.sqf
```diff
@@ -25,8 +25,12 @@ Author:
 #include "script_component.hpp"
 
 params ["_publicHandle"];
+
+if (isNil "_publicHandle" || {(_publicHandle < 0)} || {!((count GVAR(PFHhandles)) >= _publicHandle)}) exitWith { // Nil/no handle, nil action or handle is larger than Public Handle Array
```
PabstMirror: Should this be `{(count GVAR(PFHhandles) <= _publicHandle)}`?

If the count is 0, and the _publicHandle is 0, you want it to be true and exit.

ViperMaul:
@PabstMirror 
Let's make a new pull request and make that change. This PR is already merged.

PR #173, opened 2015-10-10
Restore equivalent position handling in taskDefend

SilentSpike:
Previously if no position was given the position of the group was used instead.

When I updated the function to use `params` and then retroactively check if the default position value was used I forget to make the `_position` variable use the *position* of the group rather than just the group itself.

### DIFF
--- a/addons/ai/fnc_taskDefend.sqf
+++ b/addons/ai/fnc_taskDefend.sqf
@@ -32,6 +32,7 @@ _group = [_group] call CBA_fnc_getGroup;
 if !(local _group) exitWith {}; // Don't create waypoints on each machine
 
 _position = [_position,_group] select (_position isEqualTo []);
+_position = _position call CBA_fnc_getPos;
 
 _group enableattack false;
 


addons/ai/fnc_taskDefend.sqf
```diff
@@ -32,6 +32,7 @@ _group = [_group] call CBA_fnc_getGroup;
 if !(local _group) exitWith {}; // Don't create waypoints on each machine
 
 _position = [_position,_group] select (_position isEqualTo []);
+_position = _position call CBA_fnc_getPos;
```
Killswitch00: What about having that in the second element of the array above. Like so:
`_position = [_position, _group call CBA_fnc_getPos] select (_position isEqualTo []);

SilentSpike:
I believe taskPatrol is fine, because it just passes the group over to the addWaypoint function

SilentSpike:
It should also remain as is because the `_position` isn't necessarily a raw position, it also accepts locations, objects, markers, etc. (which are handled by the getPos function)

PR #179, opened 2015-10-11
Split string v152

Killswitch00:
Use splitString for single-character splits


Killswitch00:
Abandoning this for now. Reason: `splitString` behaves differently from `CBA_fnc_split` if the input string is the empty string. The former returns an empty array, the CBA function returns an array with an empty string.

Issue #94, opened 2015-07-08
Synch fireteam colours


BaerMitUmlaut:
This is kind of fixing a BI "feature", but I think CBA should take care of this since it's one of the most used addons out there.

Currently fireteam colours are not synched in MP. This causes addons having to do workarounds that watch the assigned team colour on the team leaders client and manually synch it. If you use more than one addon that tries to do something like that, you can easily cause issues, and if you use none, the fireteam colours won't get synched at all.

Killswitch00:
We accept submissions for new functionality in CBA. If you have a solution for the problem, feel free to submit it for review and possible inclusion in CBA.

BaerMitUmlaut:
My solution would be adding a PFH on the squad leader's client. He has the authority and can assign team colors, so if the color of a team member gets changed on his client this change should get broadcasted to all squad members. Mods would have to change the team color on the squad leaders client.

I have no idea what CBA's internal structure is though, so if you can give me a hint where I could add such a PFH that would be great.

jonpas:
> I have no idea what CBA's internal structure is though, so if you can give me a hint where I could add such a PFH that would be great.

It's same as ACE3's except functions are in the root of the module folder and NOT in a `functions` subfolder.

BaerMitUmlaut:
I'm more interested in what module it should be. Networking maybe?

PR #185, opened 2015-10-18
Added team color synchronization

BaerMitUmlaut:
This synchronizes fireteam colors across the network. Other addons can either call the `"CBA_teamColorChanged"` event or indirectly trigger that event by remote calling `assignTeam` on the squad leaders machine.

See issue #94.

### DIFF
--- a/addons/common/fnc_onTeamColorChanged.sqf
+++ b/addons/common/fnc_onTeamColorChanged.sqf
@@ -1,0 +1,22 @@
+/* ----------------------------------------------------------------------------
+Internal Function: CBA_fnc_onTeamColorChanged
+
+Description:
+    Assigns the units team color if it changed on another machine.
+
+Parameters:
+    _unit - unit [OBJECT]
+    _team - team the unit got assigned to [STRING]
+
+Returns:
+    Nothing
+
+Author:
+    BaerMitUmlaut
+---------------------------------------------------------------------------- */
+
+#include "script_component.hpp"
+params ["_unit", "_team"];
+
+_unit assignTeam _team;
+_unit setVariable [QGVAR(synchedTeam), _team];
--- a/addons/common/fnc_onTeamColorChanged.sqf
+++ b/addons/common/fnc_onTeamColorChanged.sqf
@@ -2,7 +2,7 @@
 Internal Function: CBA_fnc_onTeamColorChanged
 
 Description:
-    Assigns the units team color if it changed on another machine.
+    Assigns the units team color if it changed on the squad leaders machine.
 
 Parameters:
     _unit - unit [OBJECT]
--- a/addons/common/fnc_onTeamColorChanged.sqf
+++ b/addons/common/fnc_onTeamColorChanged.sqf
@@ -19,6 +19,6 @@ Author:
 params ["_unit", "_team"];
 
 _unit assignTeam _team;
-if (isServer) then {
+if (local (leader _unit)) then {
     _unit setVariable [QGVAR(synchedTeam), _team, true];
 };
--- a/addons/common/fnc_onTeamColorChanged.sqf
+++ b/addons/common/fnc_onTeamColorChanged.sqf
@@ -19,4 +19,6 @@ Author:
 params ["_unit", "_team"];
 
 _unit assignTeam _team;
-_unit setVariable [QGVAR(synchedTeam), _team];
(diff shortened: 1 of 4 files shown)


PabstMirror:
In my opinion, checking each frame might be overkill, 1-3 seconds should be fine.

Why not have the group leader call `setVariable`.  I think if there is network traffic delays, the client could raise that event multiple times before it gets the updated variable back from the server. (really only an issue if checking each frame)

This doesn't throw script error, so nothing probably needs to be done, just an oddity to be aware of.
When zeusing into a unit, `assignedTeam player` can return nil:
```
{
    _assignedTeam = assignedTeam _x;
    diag_log text format ["%1's team is [%2](%3)", _x, _assignedTeam, (isNil "_assignedTeam")];
    if (_assignedTeam != "MAIN") then {diag_log text "A";} else {diag_log text "B";};
    true
} count units player;
//Returns `zee's team is [<null>](true)` //Note neither A or B run
```
Debug:
```
[player, ace_player] = [zee,pp_2]
assignedTeam zee = nil
assignedTeam pp_2 = "RED"
```

BaerMitUmlaut:
>Why not have the group leader call `setVariable`

Because then if I call the event through another addon the setVariable would never be called with the third parameter being true, thus it wouldn't be JIP safe. Unless I missed something?

PabstMirror:
Start Local MP Server
Change group member's team to "RED"
Event will fire each frame forever on all machines besides the server.
The variable will never be set because `remoteEvent` doesn't happen locally and all other machines will fail the isServer check.

What I am thinking of is changing remote to `globalEvent`
And then in the event do:
`if (local (leader _unit)) then {`
Which should be the group leader who triggered the event.

BaerMitUmlaut:
Fixed. I also set the PFH to run every second instead of every frame, but I could increase the delay further. ACE3 synchs them every 5 seconds, however I think that is a bit too much. It's not a very demanding function.

addons/common/fnc_synchTeamColors.sqf
```diff
@@ -0,0 +1,27 @@
...
+    BaerMitUmlaut
+---------------------------------------------------------------------------- */
+
+#include "script_component.hpp"
+
+if (leader player == player) then {
+    {
+        if ((assignedTeam _x) != (_x getVariable [QGVAR(synchedTeam), "MAIN"])) then {
+            //Local team != currently synched team, so we need to synchronize them again
+            ["CBA_teamColorChanged", [_x, assignedTeam _x]] call CBA_fnc_globalEvent;
```
Glowbal: Why a `globalEvent` and not a `targetEvent`? Is this color necessary to be synced on all clients?

addons/common/fnc_synchTeamColors.sqf
```diff
@@ -0,0 +1,27 @@
+/* ----------------------------------------------------------------------------
+Internal Function: CBA_fnc_synchTeamColors
+
+Description:
+    Synchs the team colors every second. Does not need to be called manually.
```
Glowbal: There is no time delay in this function, but rather in the PFH calling this function. I'd remove the `every second` part from the description.

addons/common/fnc_onTeamColorChanged.sqf
```diff
@@ -0,0 +1,24 @@
+/* ----------------------------------------------------------------------------
+Internal Function: CBA_fnc_onTeamColorChanged
+
+Description:
+    Assigns the units team color if it changed on another machine.
```
Glowbal: Only if it changed on the `leaders` machine?

addons/common/fnc_synchTeamColors.sqf
BaerMitUmlaut: Yes, in case another player joins the squad he needs to know the assigned colors as well.

Glowbal: Doesn't that mean you want to sync them up upon joining the team instead?

BaerMitUmlaut: That's possible too, but would require extension of the PFH (unless there's a CBA event for it? there's no doc for default events and I didn't see any either) and it would ofc have a delay. Without that delay it would be great for triggering a CBA event though.

Either options aren't perfect in my opinion. Choose your poison...

PR #191, opened 2015-11-13
-JR: implement solution for a cleaner RPT

robalo:
delete slot classes before updating their parents


commy2:
:+1: 
can confirm that this removes all updating base class warnings about cba_jr

PR #192, opened 2015-11-13
-fix misplaced ] in addPFHlogic params

robalo:
follow up on #190


jonpas:
Urgh, just write it multiple lines, issues gone.

PR #196, opened 2015-11-15
'unbrittle' CBA_fnc_getPos, close #187

### LINKED ISSUE #187: getPos is a brittle function
Same deal as fix #183 for #182.

`CBA_fnc_getPos` uses brittle code that can't handle input in array form because it deals directly with variable `_this`, however it's a bit more problematic to fix and retain backwards compatibility since the parameters passed can be an array of the form `[x,y,z]`.

commy2:


Killswitch00:
Looks good. Please "camelCase" the BIS commands to make the file more readable. (For example, it's "getMarkerPos", not "getmarkerpos")

commy2:
Done. Much better.

PR #197, opened 2015-11-16
Change local keyword to private.

NouberNou:
Fixes error in current dev branch releases where keyword is already changed.

Should not be merged until A3 release updates.

### DIFF
--- a/addons/class_eventhandlers/fnc_addClassEventHandler.sqf
+++ b/addons/class_eventhandlers/fnc_addClassEventHandler.sqf
@@ -49,15 +49,15 @@ if (_event != "init" && {!(_event in SUPPORTED_EH)}) exitWith {false};
 } count (missionNamespace getVariable [QGVAR(entities), []]);
 
 // define for units that are created later
-private "_events"; _events = EVENTHANDLERS(_event,_type);
+private _events = EVENTHANDLERS(_event,_type);
 
 _events pushBack _code;
 
 SETEVENTHANDLERS(_event,_type,_events);
 
 // set flag for this eventhandler to be used on this class. reduces overhead on init.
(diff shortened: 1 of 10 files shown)


jonpas:
:+1:

commy2:
remove extra spaces

Killswitch00:
The migration away from using the `local` keyword could be done in two stages while keeping compatibility with 1.52:

Stage 1: `local _var = <value>` -> `private "_var"; _var = <value>`
Stage 2: `private "_var"; _var = <value>` -> `private _var = <value>`

After stage 1, one could do a patch release of CBA before A3 1.54 is out.

commy2:
step 1: https://github.com/CBATeam/CBA_A3/pull/203
step 2: https://github.com/CBATeam/CBA_A3/pull/204

jonpas:
Seems like Arma update will be released this week, so this might not really be worth it.

Killswitch00:
The value of doing it would be doing proactive maintenance to make the transition to 1.54 trouble-free for users of CBA. 

Not doing it means we will have to be reactive and issue a fixed CBA after 1.54 breaks CBA.

Consider the public reaction to the two alternatives and the fact that getting a "fix" out ight take some time after 1.54 hits Steam.

jonpas:
Yes, but how is stage 1 going to help this in the first place if you don't update CBA before 1.54 hits?

Not saying there is anything wrong with it, it would work perfectly, but current CBA won't work with 1.54 due to `local` keyword being used in the first place, and if you don't update it until 1.54 it will be broken either way.

Killswitch00:
#203 is now merged, so a pre-1.54 CBA release is now possible.

PabstMirror:
Can probably close and just use #204

Killswitch00:
Yeah, #203 and later #204 will do what this one does.

PR #199, opened 2015-11-19
Handle missionConfig problem in 1.55 dev

### LINKED ISSUE #198: CBA Interaction with Arma Eden editor can cause a crash
Hey guys, hope you can get something from this.

Been looking into a DevBrach Eden Editor crash and have been able to reproduce it with a blank map, with only one player object, and only CBA running (have 2.1.2.151016) 

This issue occurs with a Saved map, after you preview, you will get a popup and a warning like the following in the rpt:
0:01:45 Warning Message: '/' is not a class ('Extended_PostInit_EventHandlers' accessed)
0:01:45 Warning Message: '/' is not a class ('Extended_InitPost_EventHandlers' accessed)

At this point if you attempt to preview again odds are it will crash, or for sure if you try and exit the editor it will crash. 

Did report this to the Bohemia guys (http://feedback.arma3.com/view.php?id=26572) so see where that goes. But I did not see anything logged around this, so wanted to point it out incase you can handle things on your end. 

Let me know if I can grab any data for you.

PabstMirror:
Fix #198

-make SLX_XEH_CONFIG_FILES a macro 
**edit** - I guess something else could write to this array?? (like another place to put XEH configs?) This seems unlikely but possible.
-use new getMissionConfig command

Also fix a problem where the description.ext `respawn` capitalization had an effect.


Killswitch00:
The `_respawn` part could be refactored to use [getMissionConfigValue](https://community.bistudio.com/wiki/getMissionConfigValue)

PabstMirror:
`getMissionConfig` / `getMissionConfigValue` are still things we can look at, but they aren't needed to fix the crash, and they are not in 1.54RC.

The changes made here should allow a CBA that is compatible with both 1.54 and dev branch.

PR #205, opened 2015-11-29
Show when a keybind is unbound

PabstMirror:
Before it would show all unbound keys as duplicates (red)

### DIFF
--- a/addons/keybinding/gui/fnc_updateGUI.sqf
+++ b/addons/keybinding/gui/fnc_updateGUI.sqf
@@ -84,7 +84,7 @@ if !(isNull _display) then {
                     TRACE_1("",_keyString);
                     // Search the handler array for any other keybinds using this key.
                     _isDuplicated = false;
-                    if(_dikCode != 0) then {
+                    if(_dikCode > 0) then {
                         _dupeActionName = "";
                         {
                             private ["_sActionId", "_dupeAction"];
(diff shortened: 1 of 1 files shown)


addons/keybinding/gui/fnc_updateGUI.sqf
```diff
@@ -106,6 +106,12 @@ if !(isNull _display) then {
                         };
                     };
 
+                    private _isUnbound = false;
```
Killswitch00: Won't work for 1.52

addons/keybinding/gui/fnc_updateGUI.sqf
```diff
@@ -106,6 +106,13 @@ if !(isNull _display) then {
                         };
                     };
 
+                    private "_isUnbound";
+                    _isUnbound = false;
+                    if (_dikCode in [-1, 0]) then {
```
commy2: Thats a pretty sophisticated way to write `<= 0` in this case. :)

Killswitch00:
I can not reproduce the problem that this PR supposedly solves. An unbound key is shown as...well, nothing in the "Assigned keys" column in the configuration dialog.

PabstMirror:
The keys on the bottom are all "unbound", but show up as duplicates in red. I think explicitly showing that they are unbound should be much clearer to users.

jonpas:
If there is only one key unbound in the category it will show up empty (this is fine), but due to empty actually being it's own "key" it will show up as a duplicate bind when having more empty binds.

Killswitch00:
Thanks for the explanation and screenshot.

jonpas:
:+1:

Killswitch00:
As mentioned elsewhere, how about displaying unbound keys as they are displayed in the normal game key assingment dialog, ie with the assigned key field empty? (As PabstMirror said, it would save us the trouble of translating the [UNBOUND] string)

PR #211, opened 2015-12-03
Fix split having extra separator at end

### LINKED ISSUE #210: CBA_fnc_splitString returns wrong result for a particular input
Reproduce with: `["babab", "bab"] call CBA_fnc_split`

Actual result: `["", "ab", ""]`

Expected result: `["", "ab"]`

PabstMirror:
Fix #210
Also use `select [_index]` when selecting entire right side of a string.

### DIFF
--- a/addons/strings/fnc_split.sqf
+++ b/addons/strings/fnc_split.sqf
@@ -45,7 +45,7 @@ if (_input == _separator) exitWith {["",""]};
 
 _lastWasSeperator = true;
 while {_index < _inputCount} do {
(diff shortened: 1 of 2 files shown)


commy2:
BbabTabAbabbabababab

Killswitch00:
It passes the units tests now. commy2, have you found another input tuple that causes problem? If not, what is meant by your comment?

commy2:
Nope. I was just joking around. I just like the unit test pabst added.

PR #214, opened 2015-12-09
Fix FormatNumber Decimals with Negative Numbers

PabstMirror:
Before:
`[-12.75,0,3] call CBA_fnc_formatNumber;` = "-12.-8-50"
`[-0.12345,1,5] call CBA_fnc_formatNumber;` = "-0.-2-3-4-5-5"


commy2:
:+1: 

It's a bit strange how it rounds upwards negative numbers ending with 5, but that seems to be consistent with what the game does anyway.
```
[-1.25,0,1] call cba_fnc_formatNumber
-> "-1.2"

`round -12.5
-> -12
```

PR #215, opened 2015-12-09
MACRO: IS_x updated

Dorbedo:
updated the "IS_x" marcos because of the new "isEqualType" command


Dorbedo:
I have choosen "IsEqualType" because it was slightly faster than "IsEqualTo".

addons/main/script_macros_common.hpp
```diff
@@ -829,21 +829,21 @@ Macros: IS_x()
...
+#define IS_DISPLAY(VAR)  IS_META_SYS(VAR,displayNull)
+#define IS_GROUP(VAR)    IS_META_SYS(VAR,grpNull)
+#define IS_OBJECT(VAR)   IS_META_SYS(VAR,objNull)
+#define IS_SCALAR(VAR)   IS_META_SYS(VAR,0)
+#define IS_SCRIPT(VAR)   IS_META_SYS(VAR,scriptNull)
+#define IS_SIDE(VAR)     IS_META_SYS(VAR,west)
 #define IS_STRING(VAR)   IS_META_SYS(VAR,"STRING")
-#define IS_TEXT(VAR)     IS_META_SYS(VAR,"TEXT")
-#define IS_LOCATION(VAR) IS_META_SYS(VAR,"LOCATION")
+#define IS_TEXT(VAR)     IS_META_SYS(VAR,parseText "--</br>--")
```
commy2: Should be:
`IS_META_SYS(VAR,text "")`
imo, which is ~3 times faster.

Dorbedo: thx, I have changed it.

commy2:
:+1:

PR #220, opened 2015-12-15
Remove unecessary entities call

BaerMitUmlaut:
No need to call `entities ""` twice.


commy2:
This might actually be slower on average. In the majority of iterations the array will be the same, so there is no need to store it in a variable.
I however only compared this with the random animal spawn while jogging when the polling was done every frame. Now it's only ten times a second, so the percentage of iterations with changes in `entities ""` is significantly higher.

PR #228, opened 2016-01-01
Fix chinook init XEH

PabstMirror:
Fix bad init event handler on chinook.

Must have been added in 1.54


commy2:
:+1: but would be no longer needed with #226

PR #231, opened 2016-01-09
Make isRecompileEnabled generic, warn FiredBIS

PabstMirror:
- Make isRecompileEnabled generic (works on `xeh`, `compile` and `functions`)

 - `["CAManBase", "firedBis", {diag_log text format ["FiredBIS %1", _this];}] call CBA_fnc_addClassEventHandler;` returned true, but had no effect.  I see this as being a possible source of confustion as it takes `fired` and returns the normal (addEventHandler) `fired` params, but converting from XEH, you might expect the event name to be `firedBIS`,

### DIFF
--- a/addons/common/init_functionsModule.sqf
+++ b/addons/common/init_functionsModule.sqf
@@ -10,6 +10,7 @@ scriptName "CBA\common\init_functionsModule";
 private ["_recompile"];
 _recompile = (count _this) > 0;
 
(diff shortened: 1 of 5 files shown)


commy2:
:+1: 
`_return = (isNumber _config && {getNumber _config == 0}) || {isNumber _missionConfig && {getNumber _missionConfig == 0}};`
can eventually be simplified with https://community.bistudio.com/wiki/getMissionConfigValue

Issue #230, opened 2016-01-09
Script error with CBA_fnc_addPerFrameHandler when used in scheduled.


commy2:
While debuging PFHs under extreme conditions I found a problem with the function when used together with the scheduler.

`0 spawn {while {true} do {systemChat str ([{[_this select 1] call CBA_fnc_removePerFramehandler}, 0, []] call cba_fnc_addPerFrameHandler)}}`

Causes these to pop up randomly:
```
20:00:08 Error in expression <, nil];
cba_common_perFrameHandlerArray set [_handle, nil];
_newArray = [];

cba>
20:00:08   Error position: <set [_handle, nil];
_newArray = [];

cba>
20:00:08   Error Zero divisor
20:00:08 File x\cba\addons\common\fnc_removePerFrameHandler.sqf, line 15
```

The reason seems to be the scheduler stoping between these lines:

```
    _publicHandle = GVAR(PFHhandles) pushback _handle;
    _data = [_func, _delay, 0, diag_tickTime, _params, _publicHandle];
    GVAR(perFrameHandlerArray) pushBack _data;
```

Causing the OEF event to iterate through `GVAR(PFHhandles)` before `GVAR(perFrameHandlerArray)` is set.

Killswitch00:
Try making the "critical section" execute nonscheduled in both addPerFrameHandler and removePFH.
(There appears to be a `CBA_fnc_directCall` in CBA common)

commy2:
That would be a solution, though direct call doesn't seem to be very optimized:

```
FUNC(directCall) = {
    private "_o";
    params ["_params","_code"];
    _o = SLX_XEH_DUMMY createVehicleLocal [0, 0, 0];
    ["CBA_DC", "onEachFrame", {
        params ["_args", "_code", "_o"];
        _args call _code;
        _o setDamage 1;
        ["CBA_DC", "onEachFrame"] call BIS_fnc_removeStackedEventHandler;
    }, [_params, _code, _o]] call BIS_fnc_addStackedEventHandler;
    _o
};
```

I know that there is a different method with way less overhead now, so I'll probably rewrite `directCall` too.

Killswitch00:
Yeah, `directCall` is an old remnant and could use an overhaul.

PR #234, opened 2016-01-10
add CBA_fnc_directCall

commy2:
This is the best I can come up with.

required for #230 

~~Makes the call atomic by reducing it to a `pushBack`.~~

can be confirmed working with:
```
TEST = {systemChat str (call CBA_fnc_isScheduled)};

call TEST;
0 spawn {
    call TEST;
    TEST call CBA_fnc_directCall;
};

-> false
-> true
-> false
```
Name does not conflict, because the old one is named `CBA_common_fnc_directCall`


commy2:
Performance test:

new:
`cba_fnc_directCall`
10000/10000
0.0880981 ms

old:
`cba_common_fnc_directCall`
1175/10000
1.49186 ms

The old one also breaks when used twice in the same frame. (Only the last directCall would be used)

PR #235, opened 2016-01-10
Dequeue logs in a thread-safe fashion

Killswitch00:
Adding and removing log messages from the queue of log
messages (CBA_LOG_ARRAY) need to be "thread-safe".

Log additions use pushBack, but the log removals were
less safe from scheduling issues.

### DIFF
--- a/addons/diagnostic/fnc_log.sqf
+++ b/addons/diagnostic/fnc_log.sqf
@@ -46,21 +46,9 @@ SCRIPT(log);
                     _file + ":"+str(_lineNum + 1), _message];
             };
 
-            while {count CBA_LOG_ARRAY > 0} do
+            while {_selected = CBA_LOG_ARRAY deleteAt 0; !isNil "_selected"} do
             {
-                _selected = CBA_LOG_ARRAY select 0;
                 _selected call _fnc_log;
-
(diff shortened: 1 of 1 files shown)


PabstMirror:
I think `_selected` would undefined because of scope

addons/diagnostic/fnc_log.sqf
```diff
@@ -46,21 +46,10 @@ SCRIPT(log);
                     _file + ":"+str(_lineNum + 1), _message];
             };
 
-            while {count CBA_LOG_ARRAY > 0} do
+            _selected = "";
```
Glowbal: Missing `private`

PabstMirror: Do privates matter in spawned code?

commy2: Not in the lowest scope I guess.

Glowbal: Good point. Doesn't matter in this context.

Killswitch00: So...good enough to be merged then?

commy2: Yes

Issue #229, opened 2016-01-09
 cba_common_PFHhandles, Error Max array size would be reached.


commy2:
>Error Max array size would be reached. Current size: 999999, wanted size: 1000000.

Another one of these:
https://forums.bistudio.com/topic/168277-cba-community-base-addons-arma-3/?p=2962614

Apparently happens only after long sessions. Might be caused by misuse of `CBA_fnc_addPerFrameHandler` (could also be something in ACE, or some different addon entirely)

I think this needs some debuging messages when the 1kk array size is reached to figure out what the cause is.

Related:
```
_publicHandle = cba_common_PFHhandles pushback _handle;
_data = [_func, _delay>
 7:10:18   Error position: <pushback _handle;
_data = [_func, _delay>
 7:10:18   Error Max array size would be reached. Current size: 999999, wanted size: 1000000.
 7:10:18 File x\cba\addons\common\fnc_addPerFrameHandler.sqf, line 16
 7:10:18 Error in expression <
```

No idea if related, but this error was spammed before PFH started:
```
if (alive btr_1) then {
btr_1 setpos [ getmarkerPo>
 3:57:43   Error position: <btr_1) then {
btr_1 setpos [ getmarkerPo>
 3:57:43   Error Undefined variable in expression: btr_1
 3:57:43 File mpmissions\__cur_mp.Woodland_ACR\mission\vehicle_fix.sqf, line 3
 6:44:23 Error in expression <e_fix.sqf"
while {true} do {
```

jokoho48:
i think this "bug" is not fixable without changing the system completly

PabstMirror:
I think it should be fixed by the people calling the code, adding PFEH each frame should not be happening.  

otherwise I see two options:

 - `GVAR(PFHhandles)` would become an array of arrays.  E.G. 0-100,000 in first array, 0-200,000 in 2nd array, and so on. Downside is increased complexity, and at some point we're going still going to have floating point precision problem of huge handle numbers.

 - Reuse old handles. biggest problem is you can get unexpected results. It breaks the promise we give that handles will always be unique.  It would only kick in after there is a problem so it's more about keeping the game working. Made a quick (untested) prototype here: https://github.com/CBATeam/CBA_A3/commit/cf8babfdeba9f89d4d8af7382806c89f1817d37a

commy2:
My first step would simply adding debug into the RPT to see which function(s) are added so much. Both cases here used ACE, so maybe we can fix it ourselves.

[...]

PR #236, opened 2016-01-10
Fix issues with PerFrameHandlers

commy2:
ref: #229, #230 

- makes CBA_fnc_addPerFrameHandler and CBA_fnc_removePerFrameHandler safe to be used in scheduled environment.
- moves stuff around in `FUNC(onFrame)` to reduce overhead
- adds debug log when more than 1kk PFHs were added per mission.
- adds PFH save game compatibility

I'm currently on 1.55 dev and there seems to be no array size limit, so I can't really test #229.

### DIFF
--- a/addons/common/fnc_removePerFrameHandler.sqf
+++ b/addons/common/fnc_removePerFrameHandler.sqf
@@ -24,7 +24,7 @@ Author:
 
 params [["_handle", -1, [0]]];
 
-if (_handle < 0 || {_handle > GVAR(lastPFHid)}) exitWith {};
+if (_handle < 0 || {_handle >= count GVAR(PFHhandles)}) exitWith {};
 
 [{
     params ["_handle"];
--- a/addons/common/fnc_removePerFrameHandler.sqf
+++ b/addons/common/fnc_removePerFrameHandler.sqf
@@ -2,10 +2,10 @@
 Function: CBA_fnc_removePerFrameHandler
 
 Description:
-    Remove a handler that you have added using CBA_fnc_addPerFrameHandler
+    Remove a handler that you have added using CBA_fnc_addPerFrameHandler.
 
 Parameters:
-    _handle - The function handle you wish to remove
+    _handle - The function handle you wish to remove. <NUMBER>
 
 Returns:
     None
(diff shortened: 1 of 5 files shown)


jokoho48:
@commy2 why you remove the clean up of the perFrameHandlerArray again? that remove every performance improvement i did?

commy2:
Can you explain what you mean? There are no checks added in this PR, only stuff removed or moved around.

jokoho48:
you remove https://github.com/CBATeam/CBA_A3/pull/236/files#diff-173548a824c3a52bc2c10f270fd8dc72L44 this completly and that cleanup the perFrameHandlerArray to improve performance and remove nil from this perFrameHandlerArray array so that dont run throw the Basic Frame handler

commy2:
There are no `nil` entries in `GVAR(perFrameHandlerArray)`, so this `isNil` is redudndant.

Try this:
```
count cba_common_perFrameHandlerArray
-> 3

0 call CBA_fnc_removePerFrameHandler

count cba_common_perFrameHandlerArray
-> 2
```

jokoho48:
ok now i see what you did :+1:

commy2:
Yeah. You might have been thinking of `GVAR(PFHhandles)`.

```
cba_common_PFHhandles
-> [0,1,2]

1 call CBA_fnc_removePerFrameHandler

cba_common_PFHhandles
-> [0,any,2]
```

addons/common/fnc_removePerFrameHandler.sqf
```diff
@@ -18,34 +18,24 @@ Examples:
...
+[{
+    params ["_handle"];
+
+    GVAR(perFrameHandlerArray) deleteAt (GVAR(PFHhandles) select _handle);
+    GVAR(PFHhandles) set [_handle, nil];
+
+    {
+        _x params ["", "", "", "", "", "_handle"];
+        GVAR(PFHhandles) set [_handle, _forEachIndex];
+    } forEach GVAR(perFrameHandlerArray);
```
PabstMirror: This forEach loop needs to be above line 32 or it will fail on 2nd call to :
```
x_A = [{      diag_log text format ["---%1 - A Running %2", diag_tickTime, (_this select 1)];  }, 5, []] call CBA_fnc_addPerFrameHandler;
x_B = [{      diag_log text format ["---%1 - B Running %2", diag_tickTime, (_this select 1)];  }, 5, []] call CBA_fnc_addPerFrameHandler;
[x_A] call CBA_fnc_removePerFrameHandler;
[x_B] call CBA_fnc_removePerFrameHandler;  
```

PabstMirror: All we really need is to find index of the top level `_handle` in the `perFrameHandlerArray` array, so, we could add an exit to the loop and skip the set I think.

jokoho48:
i knew something look wrong hmm

addons/common/fnc_removePerFrameHandler.sqf
commy2: Using a `find` would be slow when `GVAR(PFHhandles)` get's big. With this method you only have to iterate through the PFHs that are still active. No idea which one is better. This is what already was there.

commy2: Derp. The problem is that the __public__ handle is pushed to `GVAR(PFHhandles)` (which is for the internal handles, those who can change).
Thats why removing the PFH twice "works". The loop auto corrects the internal handle...

commy2:
Should be fixed, @PabstMirror

commy2:
I think we should use this PR to figure out what PFH is added more than a million times in missions running longer than 4 hours. Maybe we can fix that (if it's a PFH in ACE) and we don't need to change handles at all.

Killswitch00:
I agree - the debug printout should  be useful in identifying the PFH overruns

PR #239, opened 2016-01-15
Fix include file name capitalizations

Killswitch00:
The on-disk filenames for the event handler definition
header files is "CfgEventHandlers.hpp". The include statement
needs to specify exactly that for builds in Linux to work.
(Unix file systems have case-significant file names)


commy2:
:+1: 

>The include statement
needs to specify exactly that for builds in Linux to work.

I didn't know that this also applies to Arma configs. Interesting.

PR #241, opened 2016-01-16
Fix XEH_ENABLED macro

PabstMirror:
Explosive triggers aren't working in ace, which are a vehicle that that inherits `XEH_DISABLED` from Thing.

XEH_ENABLED macro use of `delete` does not seem to have any effect.  I believe we need to zero out the number instead.

Without this:
```
    class ACE_Explosives_Place: Items_base_F {
        XEH_ENABLED;
```
`getNumber (configfile >> "CfgVehicles" >> "ACE_Explosives_Place" >> "SLX_XEH_DISABLED")` = 1


commy2:
I can confirm the same happening with wheels.

PR #243, opened 2016-01-19
Remove XEH Caching

PabstMirror:
With the new XEH overhaul, we were baking the `isServer` checks into the XEH code for server events. 

By eliminating the cache and building the code at runtime, we can add only valid XEH based on the machine type.
With CBA/ACE, compiling configFile's XEH takes less than a 1/4 second on my machine, with a full modset, it's less than 1/2 a second, so we weren't gaining much by caching.

Performance gains are **very** minimal, but `isServer` `!isDedicated` and the `forEach` iteration all take some time. I think it's better to spend a little more time at mission load than during gameplay.

I've also added a rpt warning for Composite XEH that don't have any matching text (ace's cargo was doing this).

Example:
```
class Extended_Killed_EventHandlers {
    class All {
        class test0 { killed = "diag_log text format ['test0 %1', _this];"; };
        class test1 { serverKilled = "diag_log text format ['test1 %1', _this];"; };
        class test2 { clientKilled = "diag_log text format ['test2 %1', _this];"; };
        class test3 { init = "diag_log text format ['test3 %1', _this];"; }; //text does not match event, this won't do anything
    };
};
```
Checking `cursorTarget getVariable "cba_xeh_killed"`
```
[ //Before:
{_this call ace_attach_fnc_handleKilled;},
{call ace_cargo_fnc_handleDestroyed;},
{diag_log text format ['test0 %1', _this];;},
{if (isServer) then {diag_log text format ['test1 %1', _this];};},
{if (!isDedicated) then {diag_log text format ['test2 %1', _this];};},
{}
]

[ //After:
{_this call ace_attach_fnc_handleKilled;},
{call ace_cargo_fnc_handleDestroyed;},
{diag_log text format ['test0 %1', _this];;},
{diag_log text format ['test1 %1', _this];;},
{diag_log text format ['test2 %1', _this];;}
]
```

And on dedicated we don't add clientXEH and clients don't add serverXEH.

### DIFF
--- a/addons/xeh/fnc_compileEventHandlers.sqf
+++ b/addons/xeh/fnc_compileEventHandlers.sqf
@@ -64,11 +64,8 @@ private _resultNames = [];
             };
         };
 
-        // only add event on machines where it exists
-        if !(_eventFunc isEqualTo "") then {
-            _result pushBack ["", _eventName, compile _eventFunc];
-            _resultNames pushBack _customName;
-        };
+        _result pushBack ["", _eventName, compile _eventFunc];
+        _resultNames pushBack _customName;
     } forEach configProperties [_baseConfig >> XEH_FORMAT_CONFIG_NAME(_eventName)];
 } forEach ["preStart", "preInit", "postInit"]; 
 
@@ -170,11 +167,25 @@ private _resultNames = [];
 
             // only add event on machines where it exists
             if !(_eventFunc isEqualTo _eventFuncBase) then {
-                _result pushBack [_className, _eventName, compile _eventFunc, _allowInheritance, _excludedClasses];
-                _resultNames pushBack _customName;
+                _eventFunc = compile _eventFunc;
+            } else {
+                _eventFunc = {};
             };
+
+            _result pushBack [_className, _eventName, _eventFunc, _allowInheritance, _excludedClasses];
+            _resultNames pushBack _customName;
         } forEach configProperties [_x];
     } forEach configProperties [_baseConfig >> XEH_FORMAT_CONFIG_NAME(_eventName), "isClass _x"];
 } forEach [XEH_EVENTS];
 
-_result
+//_result select {!((_x select 2) isEqualTo {})} @todo 1.55 dev, delete everything below
+
+private _return = [];
+
+{
+    if !((_x select 2) isEqualTo {}) then {
+        _return pushBack _x;
+    };
+} forEach _result;
+
+_return
--- a/addons/xeh/fnc_compileEventHandlers.sqf
+++ b/addons/xeh/fnc_compileEventHandlers.sqf
@@ -45,17 +45,21 @@ private _resultNames = [];
             };
 
             // client only events
-            private _entryClient = _x >> "clientInit";
+            if (!isDedicated) then {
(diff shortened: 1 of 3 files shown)


esteldunedain:
:+1:

addons/xeh/fnc_compileEventHandlers.sqf
```diff
@@ -165,7 +174,7 @@ private _resultNames = [];
                 };
             } forEach _resultNames;
 
-            _result pushBack [_className, _eventName, compile _eventFunc, _allowInheritance, _excludedClasses];
+            _result pushBack [_className, _eventName, compile _eventFunc, _allowInheritance, _excludedClasses, _doesSomething];
```
commy2: There is no need to even push them to the results if they don't match.

Glowbal: Agreed. Would also make for cleaner code.

addons/xeh/fnc_preInit.sqf
```diff
@@ -61,15 +61,20 @@ GVAR(EventsLowercase) = [];
...
+// always recompile extended event handlers
+#ifdef DEBUG_MODE_FULL
+    XEH_LOG("XEH: Compiling XEH START");
+#endif
+
+GVAR(allEventHandlers) = [];
+
+{
+    GVAR(allEventHandlers) append (_x call CBA_fnc_compileEventHandlers);
+} forEach [configFile, campaignConfigFile, missionConfigFile];
```
esteldunedain: You changed the order from:

`configFile -> missionConfigFile -> campaignConfigFile`

to 

`configFile -> campaignConfigFile -> missionConfigFile`

Could that affect some existing content?

commy2: Originally it was `configFile -> campaignConfigFile -> missionConfigFile` too (copied this order from the old branch). Shouldn't make any difference though.

commy2:
Found a problem. WIP

addons/xeh/fnc_preInit.sqf
esteldunedain: :+1: + it makes more sense

commy2:
Done. Can be merged.
Has potential to be simplified post v1.55 dev, which I put inside a comment.

addons/xeh/fnc_compileEventHandlers.sqf
```diff
@@ -165,10 +181,29 @@ private _resultNames = [];
...
             } forEach _resultNames;
 
-            _result pushBack [_className, _eventName, compile _eventFunc, _allowInheritance, _excludedClasses];
+            // only add event on machines where it exists
+            if !(_eventFunc isEqualTo _eventFuncBase) then {
+                _eventFunc = compile _eventFunc;
+                TRACE_3("does something",_customName,_className,_eventName);
+            } else {
+                _eventFunc = {};
+                TRACE_3("does NOT do something",_customName,_className,_eventName);
```
commy2: Like this? @PabstMirror

PabstMirror: Looks good, thanks commy

PR #246, opened 2016-01-22
JR: Do not blacklist long scopes on the short mg rails anymore, move …

### LINKED ISSUE #159: High zoom optics broken on all LMGs
CBA3 Version: v200150817-fixed

Mods:
    @CBA_A3

Description:

Most high zoom optics can not be used on any LMGs.
List:
- LMG_Zafir_F
- LMG_Mk200_F
- MMG_02_sand_F

When attempting to equip the scope in VA or via script, it won't appear.
Broken Scopes include:
- AMS
- DMS (optic_DMS)
- Kahila
- LRPS 
- MOS

Steps to reproduce:

Start Virtual Arsenal
Equip any lmg from list above 
Equip any of the optics above

Where did the issue occur?

    "Multiplayer", "Editor"

RPT log file:
n/a

robalo:
…that to optional pbo instead. Fixes #159.

### DIFF
--- a/addons/jr/jr_classes.hpp
+++ b/addons/jr/jr_classes.hpp
@@ -87,6 +87,7 @@ class asdg_OpticRail1913_short: asdg_OpticRail1913 { // attachments that fit on 
 
 class asdg_OpticRail1913_short_MG: asdg_OpticRail1913_short { // attachments that would not work well on machine guns
     class compatibleItems: compatibleItems {
+/* // moved to optional addon: jr_disable_long_scopes_on_short_mg_rail
         optic_SOS = 0;
         optic_DMS = 0;
         optic_LRPS = 0;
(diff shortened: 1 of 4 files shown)


commy2:
Works :+1: Only problem is that these attachments are still shown in the arsenal, but are not useable (at least in current 1.55 dev).
I think that is a problem of the VA though.

Killswitch00:
commy2, does the same no-usability issue happen if you start 1.55dev without CBA? If not, it is worth investigating further.

robalo:
That behavior would be when the scope is incompatible (will happen with the optional pbo) and it's indeed a problem in VA which does not filter compatible items properly: http://feedback.arma3.com/view.php?id=23111

Killswitch00:
Ah, OK - "good" news (read: it's not CBA)

Killswitch00:
Hmm...that did not go where it should...

robalo:
Might be my bad, did PR from diff between 2 of my branches.

PR #247, opened 2016-01-23
replace PUSH() Macro with pushback

jokoho48:


addons/common/fnc_getGroupIndex.sqf
```diff
@@ -42,7 +42,7 @@ _number = [];
 // Format of player label is "<groupName>:<groupNumber> <playerName>"
 for "_i" from (_groupLabelLen + 1) to ((count _labelArray) - 1) do {
     if ((_labelArray select _i) == ASCII_SPACE) exitWith {};
-    PUSH(_number,_labelArray select _i);
+    _number pushBack _labelArray select _i;
```
commy2: `_number pushBack (_labelArray select _i);

addons/common/XEH_postInit.sqf
```diff
@@ -96,7 +96,7 @@ if !(isDedicated) then {
                 [GVAR(actionList), {
                     TRACE_3("Inside the code for the hashPair",(vehicle player),GVAR(actionIndexes), _value);
                     if (!isNil "_value" && typeName(_value) == "ARRAY") then {
-                        PUSH(GVAR(actionIndexes), (vehicle player) addAction _value)
+                        GVAR(actionIndexes) pushBack (vehicle player) addAction _value
```
Killswitch00: Does this need safeguarding with
` GVAR(actionIndexes) pushBack ((vehicle player) addAction _value)` ?

esteldunedain: When in doubt its better to add it anyway

commy2: Yeah. This is missing the brackets too.
Evaluations of binary commands are done from left to right:
`GVAR(actionIndexes) pushBack (vehicle player) addAction _value`
->
`<index/number> addAction _value`
->
`Error Number exected object`

The brackets around `(vehicle player)` on the other hand are optional.

Correct and minimalistic:
`GVAR(actionIndexes) pushBack (vehicle player addAction _value)

I personally think that it's best to use as few brackets as possbile. Only exception would be
`(unaryCommand _var1) binaryCommand _var2`, so in the end exactly what killswitch proposed.

esteldunedain: > I personally think that it's best to use as few brackets as possbile

I know you do :wink:.

Just to ring the opposite bell, I think it's usefull for the code to be as self-obvious as possible, so reviewers/maintainers don't have to wonder if a certain line is being interpreted correctly or not.

commy2: >I think it's usefull for the code to be as self-obvious as possible
True. But everyone reviewing SQF should know the basic principles. And too many brackets hurt readability.
That discussion doesn't belong here though. The line is currently wrong and errors out.

jokoho48:
ok change/fixed

commy2:
:+1:
looks good

PR #251, opened 2016-02-04
Cleanup Vector Functions

PabstMirror:
- Add unit test
 - Cleanup, new privates, and use vector commands
 - Fix errors with `vectMagn`/`vectMagn2d` (I have no clue why they were calling simplifyAngle. `vectMagn2d` always returned nil as the function didn't exist)


Killswitch00:
Neat. Please add the new unit test to the `CATEGORIES` macro in `main\test.sqf` aswell to have it run if/when one runs the whole test suite.

Issue #255, opened 2016-02-06
Versioning system shows "long" version numbers wrong


Killswitch00:
The ALiVE version string is of the form `major.minor.patchlevel.build`, where `build` is a 7-digit number of the form YYMMDDR (ie the date followed by a release increment number). The build number causes the CBA version output to look "funny":
`10:47:14 [782,39.253,0,"CBA_VERSIONING: cba=2.3.0.160128, alive_main=0.9.10.1.51013e+006, "]`

Fix the string conversion so that the resulting version string looks proper, ie so that 
`alive_main=0.9.10.1510131` is shown.

PR #256, opened 2016-02-06
Fix CBA versioning string formatting

Killswitch00:
Version number components that are too large (over six digits?)
will be formatted into a scientific notation by the default
game string conversions. (Eg 1234567 will become "1.23457e+006")

This change uses a filter function to convert the version number
components to the expected format.


commy2:
:+1: works

PabstMirror:
Will close #255

PR #260, opened 2016-02-11
add CBA_fnc_createNamespace and CBA_fnc_deleteNamespace

commy2:
This is a faster alternative to creating pseudo-namespaces compared to `createVehicleLocal` with `Logic` or similar. It also never runs into any limitations regarding group count and is strictly local.
It is based on creating a location outside of the map which is used as container for variables.

Perf test:
`call CBA_fnc_createNamespace;`
0.0207031 ms
(10000/10000)

`"Logic" createVehicleLocal [-1000,-1000,0];`
0.138744 ms
(6649/10000)

Usage:
```
test_namespace = call CBA_fnc_createNamespace;
test_namespace setVariable ["test_value", 123];
```
```
test_namespace getVariable "test_value"
-> 123
```
```
allVariables test_namespace
["test_value"]
```
```
test_namespace call CBA_fnc_deleteNamespace;
```

Only downside is, that the `getVariable ARRAY` syntax does not work it.

### DIFF
--- a/addons/common/fnc_createNamespace.sqf
+++ b/addons/common/fnc_createNamespace.sqf
@@ -20,5 +20,6 @@ Author:
     commy2
 ---------------------------------------------------------------------------- */
 #include "script_component.hpp"
+SCRIPT(createNamespace);
 
 createLocation ["CBA_NamespaceDummy", [-1000, -1000, 0], 0, 0]
--- a/addons/common/fnc_createNamespace.sqf
+++ b/addons/common/fnc_createNamespace.sqf
@@ -1,0 +1,24 @@
+/* ----------------------------------------------------------------------------
+Function: CBA_fnc_createNamespace
+
+Description:
+    Creates a namespace. Used to store and read variables via setVariable and getVariable.
+    The Namespace is destroyed after the mission ends. getVariable ARRAY is not supported.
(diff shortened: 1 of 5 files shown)


addons/common/fnc_createNamespace.sqf
```diff
@@ -0,0 +1,24 @@
...
+Examples:
+    (begin example)
+        _namespace = call CBA_fnc_createNamespace;
+    (end)
+
+Author:
+    commy2
+---------------------------------------------------------------------------- */
+#include "script_component.hpp"
+
```
Killswitch00: `SCRIPT(createNamespace);`

addons/common/fnc_deleteNamespace.sqf
```diff
@@ -0,0 +1,25 @@
...
+Examples:
+    (begin example)
+        _namespace call CBA_fnc_deleteNamespace;
+    (end)
+
+Author:
+    commy2
+---------------------------------------------------------------------------- */
+#include "script_component.hpp"
+
```
Killswitch00: `SCRIPT(deleteNamespace);`

esteldunedain:
Perfect for ACE. Thanks

PR #261, opened 2016-02-11
Add Caching to CBA_fnc_compatibleItems

jokoho48:
adding Caching for config values based on #260


PabstMirror:
There never is a call to `setVariable` to save the results in the cache.
The filter only runs on a cache hit, so would be ignored on the first call.
There is no proper return on a cache miss, it always just returned `0` from the last `count` statement.
There is an extra `private ["_compatibleItems"];` inside the `if (isNil "_compatibleItems") then {` so there would be scope issues on that variable.

@jokoho48 look at d69a0973eb22f49b34a512ddcf212a09f7c065ba for fixes and a test case.
Can probably just PR that branch if you are ok with it.

PR #262, opened 2016-02-17
Delay installing PlayerActions PFEH until first use

PabstMirror:
Saves installing a PFEH if the feature is not used.


commy2:
:+1: tested and works

Glowbal:
Nice :+1:

Killswitch00:
This needs to be rebased due to the recent PUSH macro removal PR having been merged (the problematic file is addons/common/XEH_postInit.sqf, which, in this PR is different from the same file in master currently)

Issue #244, opened 2016-01-21
Conflict between CBA_Events and CBA_Help prevents mission from loading


commy2:
`CBA_fnc_readKeyFromConfig` is intended to work with number entries (https://github.com/CBATeam/CBA_A3/blob/master/addons/events/fnc_readKeyFromConfig.sqf#L27), but the `help` module (https://github.com/CBATeam/CBA_A3/blob/master/addons/help/XEH_preClientInit.sqf#L62-L90) conflicts with that preventing the mission to start, because it assumes all entries of the sub classes in `CfgSettings\CBA\Events\` are classes themselves.

Error message:
`'action1/' is not a class ('key' accessed)`

```
class CfgSettings {
    class CBA {
        class Events {
            class TestAddon {
                action1 = 15; // tab

                class action2 { // ctrl T
                    key = 20;
                    shift = 0;
                    ctrl = 1;
                    alt = 0;
                };
            };
        };
    };
};
```

commy2:
```
a1 = ["TestAddon", "action1", { systemChat str _this }] call CBA_fnc_addKeyHandlerFromConfig
a2 = ["TestAddon", "action2", { systemChat str _this }] call CBA_fnc_addKeyHandlerFromConfig
```

PR #265, opened 2016-02-18
improve / fix help module, fix #258 and #244

### LINKED ISSUE #258: Missing keybind in menu and script error on mission start
This was reported multiple times already, but no one bothered to post a RPT. The root cause for it was already fixed, but there are still people out there that occasionally have the issue, due to these entries remaining in the profileNamespace.

```
15:27:29 Error in expression <_mod = _keyBind select 1;
_shift = _mod select 0;
_ctrl = _mod select 1;
_alt = >
15:27:29   Error position: <select 0;
_ctrl = _mod select 1;
_alt = >
15:27:29   Error select: Type Number, expected Array,String,Config entry
15:27:29 File x\cba\addons\help\XEH_postClientInit.sqf, line 64
```

commy2:
- fixes #258 (hopefully) and #244
- fixes: mods in main menu now auto cycle even when `-world="empty"`
- fixes: addKeyinding keys don't overwrite the help diary entry of all config based key handlers anymore
- uses new scripting commands (`private` keyword, `configClasses`, `params` etc.)
- uses faster `createNamespace` instead of hashes
- removes header files that are in Arma already and includes them instead
- removes `CBA_help_fnc_describe`, because it's never used, not documented and does no longer work sufficiently in A3
- adds function headers, uses macros where appropriate, etc.


addons/help/config.cpp
```diff
@@ -14,3 +14,5 @@ class CfgPatches {
 
 #include "CfgEventhandlers.hpp"
 #include "CfgRscStd.hpp"
+
+//#include "test_key.hpp"
```
Killswitch00: Is this needed?

addons/help/stringtable.xml
```diff
@@ -1,60 +1,62 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<Project name="Arma2"><Package name="CBA">
+<Project name="Arma2">
```
Killswitch00: Should this be bumped to Arma3?

PR #270, opened 2016-02-21
Add the ability to disable fnc_globalExecute

### LINKED ISSUE #269: Exile devs wont use CBA_A3 because:
The Exile Devs complain that CBA : 

> Using CBA on any Arma server will open a big door for hackers. Actually, it will make it very easy for them. To show you how simple it is; guess what this does:
~snip~

> If you would like cars flying around, all people dropped to the same place, all bases destroyed, all players killed, endless money for everyone... implement it on your own. But be aware of the consequences (And tell me the server address, so I can convince you it's bad ;)).

Killswitch00:
Mission makers and/or server managers can disable the use of
CBA_fnc_globalExecute either by using a mission setting or
an optional addon. When disabled, calling fnc_globalExecute
does nothing. Fixes #269

### DIFF
--- a/addons/network/XEH_preInit.sqf
+++ b/addons/network/XEH_preInit.sqf
@@ -94,10 +94,10 @@ if (SLX_XEH_MACHINE select 3) then {
 };
 
 // CBA_fnc_globalExecute is normally enabled
-_disableGE = false;
+private _disableGE = false;
 
 // Check the CfgSettings >> "CBA" >> "network" >> "disableGlobalExecute" value
-_entry = (CFGSETTINGSS(CBA,COMPONENT) >> "disableGlobalExecute");
+private _entry = (CFGSETTINGSS(CBA,COMPONENT) >> "disableGlobalExecute");
 if (isNumber _entry) then {
     _disableGE = (getNumber _entry == 1);
 } else {
@@ -107,7 +107,7 @@ if (isNumber _entry) then {
 };
 
 // Check the mission configuration
-_dGE = getMissionConfigValue ["CBA_disableGlobalExecute", -1];
+private _dGE = getMissionConfigValue ["CBA_disableGlobalExecute", -1];
 if (_dGE isEqualType 0) then {
     if (_dGE in [0, 1]) then {
         _disableGE = (_dGE == 1);
--- a/addons/network/XEH_preInit.sqf
+++ b/addons/network/XEH_preInit.sqf
@@ -46,8 +46,6 @@ ADDON = false;
 ISNIL(timeSync_Disabled,true); // deprecated
 ISNIL(weatherSync_Disabled,true);
 
-PREP(exec);
-
 DEPRECATE(fnc_remoteExecute,fnc_globalExecute);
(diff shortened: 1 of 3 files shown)


addons/network/XEH_preInit.sqf
```diff
@@ -95,8 +93,38 @@ if (SLX_XEH_MACHINE select 3) then {
     };
 };
 
+// CBA_fnc_globalExecute is normally enabled
+_disableGE = false;
```
jonpas: `private`

addons/network/XEH_preInit.sqf
```diff
@@ -95,8 +93,38 @@ if (SLX_XEH_MACHINE select 3) then {
     };
 };
 
+// CBA_fnc_globalExecute is normally enabled
+_disableGE = false;
+
+// Check the CfgSettings >> "CBA" >> "network" >> "disableGlobalExecute" value
+_entry = (CFGSETTINGSS(CBA,COMPONENT) >> "disableGlobalExecute");
```
jonpas: `private`

addons/network/XEH_preInit.sqf
```diff
@@ -95,8 +93,38 @@ if (SLX_XEH_MACHINE select 3) then {
...
+if (isNumber _entry) then {
+    _disableGE = (getNumber _entry == 1);
+} else {
+    if (isText _entry) then {
+        _disableGE = (getText _entry == "true");
+    };
+};
+
+// Check the mission configuration
+_dGE = getMissionConfigValue ["CBA_disableGlobalExecute", -1];
```
jonpas: `private`

esteldunedain:
:+1: for the optional pbo

addons/network/XEH_preInit.sqf
```diff
@@ -95,8 +93,38 @@ if (SLX_XEH_MACHINE select 3) then {
...
+    };
+} else {
+    if (_dGE isEqualType "") then {
+        _disableGE = (_dGE == "true");
+    };
+};
 
-[QUOTE(GVAR(cmd)), { if (GVAR(init)) then { _this spawn FUNC(exec) } }] call (uiNamespace getVariable "CBA_fnc_addEventHandler");
+if (_disableGE) then {
+    uiNamespace setVariable [QGVAR(exec), compileFinal ""];
```
PabstMirror: Should this be `QUOTE(DFUNC(exec))`?

I don't think we want to use uiNameSpace / caching for this function at all.
Otherwise I can create a mission with it disabled, and it will compileFinal the function to `{}` in UI
then join a server with it enabled and PREP will still pull the `{}`.

Killswitch00: Good point - the idea was to make sure "FUNC(exec)" cannot be used nor restored (by non-CBA means) while globalExecute is disabled.

commy2:
looks good :+1:

ViperMaul:
Good. Let's update the optionals section of README.md

Killswitch00:
This needs more work to handle a few more...issues. (Discussed elsewhere)

PR #273, opened 2016-02-23
add CBA_missionTime

commy2:
BI seems to be unable to make a proper time command that is synched in MP and is correctly resumed in save games, JIP and all that.

This PR adds the `CBA_missionTime` variable.

pros:

- based on tickTime
- set every frame
- synched in MP/JIP
- resumes correct value when loading save games
- multiplied by accTime (SP only)
- paused when game is paused (ESC menu, SP only)

cons:

- -1 on clients during preInit (correct value on and after postInit)
- not linked to ingame day time (faster)
- ignores `skipTime`, `setDate`, `timeMultiplier` etc. (ingame day time)

Test:
https://gist.github.com/commy2/c01e4885b8961363ef7e
https://gist.github.com/commy2/06de40913799965ba8e5

`0 = 0 spawn {waitUntil {diag_log CBA_missionTime; systemChat str CBA_missionTime; false}};`

I ran it for 2hrs and the values didn't diverge. (Although it was two game instances on the same machine)

### DIFF
--- a/addons/common/init_perFrameHandler.sqf
+++ b/addons/common/init_perFrameHandler.sqf
@@ -8,8 +8,8 @@
 GVAR(perFrameHandlerArray) = [];
 GVAR(fpsCount) = 0;
 GVAR(lastCount) = -1;
-GVAR(lastFrameRender) = 0;
-GVAR(lastTickTime) = 0;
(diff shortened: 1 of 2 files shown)


esteldunedain:
Looks good to me. This is critical to support things that require timing across save/load games.

PR #275, opened 2016-02-24
Use remote exec in network

commy2:
alternative solution for #269 and #270 

This way we can completely ignore all security things as they are handled by `CfgRemoteExec` and BI's / Battleyes security framework.

Also changes `CBA_fnc_globalSay` and `CBA_fnc_globalSay3d` to use `remoteExecCall`

test:
```
private _fnc_test = {sleep 1; systemChat str _this};

[-2, _fnc_test, "All"] call CBA_fnc_globalExecute;
[-1, _fnc_test, "ClientOnly"] call CBA_fnc_globalExecute;
[ 0, _fnc_test, "ServerOnly"] call CBA_fnc_globalExecute;
```

```
[[player], "Alarm"] call CBA_fnc_globalSay;
[[player], ["Alarm", 500, 1]] call CBA_fnc_globalSay;
[player, "Alarm"] call CBA_fnc_globalSay;
```

```
[player, "Alarm"] call CBA_fnc_globalSay3d;
[player, "Alarm", 500] call CBA_fnc_globalSay3d;
```

The function header of `CBA_fnc_globalSay3d` claims that you can use an array of objects as targets (similar to `CBA_fnc_globalSay`, probably copy paste ...). That was broken and caused a script error. Fixed now.

```
[[player], "Alarm"] call CBA_fnc_globalSay3d;
```


Killswitch00:
Much better than the hack I cooked up.

PR #238, opened 2016-01-15
improve some Hash Functions

jokoho48:
improve some hash functions

simplfy and cleanup

WIP

PR #278, opened 2016-02-25
Hashcleanup

commy2:
replaces #238 
everyone should be able to edit this now

### DIFF
--- a/addons/hashes/fnc_hashCreate.sqf
+++ b/addons/hashes/fnc_hashCreate.sqf
@@ -32,23 +32,10 @@ Author:
 SCRIPT(hashCreate);
 
 // -----------------------------------------------------------------------------
+params [["_array", [], [[]]], "_defaultValue"];
 
-DEFAULT_PARAM(0,_array,[]);
-DEFAULT_PARAM(1,_defaultValue,nil);
-private ["_keys", "_values"];
-
-_keys = [];
-_values = [];
-
-_keys resize (count _array);
-_values resize (count _array);
-
-for "_i" from 0 to ((count _array) - 1) do
-{
-    _keys set [_i, (_array select _i) select 0];
-    _values set [_i, (_array select _i) select 1];
-};
+private _keys = _array apply {_x select 0};
+private _values = _array apply {_x select 1};
 
 // Return.
-[TYPE_HASH, _keys, _values,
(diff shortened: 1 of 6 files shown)


esteldunedain:
@commy2, what about changing the underlying structure to use `CBA_fnc_createNamespace`? `fnc_hashEachPair` could be based on `allVariables` which support locations.

commy2:
hashes can use key different from strings, which is a benefit.
We can internally replace all hashes that only use strings as keys with createNamespace though (as long as they are not used in ui, i.e. the main menu)

esteldunedain:
> hashes can use key different from strings, which is a benefit.

oh, that's true. although using `str(_object)`as key could also work for almost all cases.

> We can internally replace all hashes that only use strings as keys with createNamespace though (as long as they are not used in ui, i.e. the main menu)

Yes, that's a much safer alternative, although it's obviously much more work.

esteldunedain:
Another possible alternative would be to add a second set of functions for "FastHashes", that only support string keys.

esteldunedain:
> Another possible alternative would be to add a second set of functions for "FastHashes", that only support string keys.

Besides, for the first time in CBA those would be REAL hashes

commy2:
`str(_object)` is not reliable. It changes when the vehicle var name changes or when the object switches locality (#remote tag)

PabstMirror:
location hashes are local only (I don't think you can publicVariable a location)
I think ACRE has some stuff to serialze them, but it's not trivial.

commy2:
ed was trying to convert objects to strings.

esteldunedain:
Yeah, you are right. Porting current CBA Hashes to `CBA_fnc_createNamespace` doesn't seem viable. Creating API calls for real hashes that coexist with the current ones would be a better solution.

commy2:
Real hashes as in using LOCATIONS with string keys? Converting objects to strings?

Two LOCATIONS can do the trick making the object key the value of a string key LOCATION.

esteldunedain:
> Real hashes as in using LOCATIONS with string keys? 

yes

> Converting objects to strings? Two LOCATIONS can do the trick making the object key the value of a string key LOCATION.

I don't think that'd work reliably

commy2:
PR still has the same problems as the other one, tag WIP

PR #284, opened 2016-02-29
CBA events rewrite

commy2:
Oh boy, here we go.

##### CBA event system

- use 1 (2 on server) PVEH instead of 4-5
- use params, private keyword, all these other handy new commands, macros and drastically simplify
- use `CBA_fnc_createNamespace` instead of hashes
- also use `CBA_fnc_createNamespace` instead of two local game logics (2x less `createVehicleLocal "logic"`) per machine
- use hashes to keep event arrays small. -> less overhead when many events were removed.

add `CBA_fnc_targetEvent`
- similar to `CBA_fnc_globalEvent` in use, but can specify machines that recieve events by passing objects, groups and arrays of objects and groups as third paramter.
- event will be executed on the machine(s) where the object(s) or/and group(s) is local

add `CBA_fnc_serverEvent`
- similar to `CBA_fnc_globalEvent`, but will only execute event on server machine

add return value to `CBA_fnc_localEvent`.
- calling the event will return the return value of the last event added. (Could be used like "handleDamage" event, where the last return value does stuff)

____

deprecate:
`CBA_fnc_addClientToServerEventhandler`, `CBA_fnc_clientToServerEvent`, `CBA_fnc_removeClientToServerEvent`
- emulated with `CBA_fnc_serverEvent`

`CBA_fnc_addReceiverOnlyEventhandler`, `CBA_fnc_receiverOnlyEvent`, `CBA_fnc_removeReceiverOnlyEvent`
- emulated with `CBA_fnc_targetEvent`

`CBA_fnc_addLocalEventHandler`, `CBA_fnc_whereLocalEvent`, `CBA_fnc_removeLocalEventHandler`
- emulated with `CBA_fnc_targetEvent`

- delete internal functions no longer used, associated with those above
- events are now guaranteed to work during postInit instead of after postInit.

____

#####  CBA display handler

`CBA_fnc_addDisplayHandler`
- now properly supports all possible events on the mission display (46)
- now correctly handles restarted missions. events will not carry over. duplicates are removed on mission start.
- now correctly handles resumed missions. mission events will be re added on main display load (using a `DisplayLoad` XEH)

- removes strange `createTrigger` (present even in MP!) solution that would reassign all key handlers when no keys were pressed for 5 seconds
- loading a mission will therefore immediately re add the display handlers now, instead of having to idle a few seconds

____

##### CBA keyhandler

- reduce amount SQF by rewriting and new command usage
- replace 5 different hashes used onKeyDown (executed every frame when at least one key is held down) with `CBA_fnc_createNamespace` or arrays
- split keyHandler in keydown keyUpHandler

^ these should drastically reduce the amount of overhead for the CBA keybinding module

- fixes a bug where calling `CBA_fnc_addKeyHandler` with the same hashkey twice would result in the last added event to be executed twice.
- fixes a bug where `CBA_fnc_addKeyHandlerFromConfig` would prevent a mission from loading when the added event used the simple config syntax without modifiers
- uses BIs DIK table as #include to replace magic numbers in code

____

##### Misc.
- log moved from preInit to postInit so `didJIP` correctly reports https://community.bistudio.com/wiki/didJIP
- log new `missionVersion` command on postInit (detects 3den vs 2d mission)
- adds `CBA_isHeadlessClient`, loged on postInit

- don't overload onLoad for `RscDisplayCurator`. uses `DisplayLoad` XEH instead
- removes unnecessary internal `CBA_MISSION_START` event
- removes unused undocumented (internal?) `CBA_playerSpawn` event. Would report if the player respawned, but unreliable due to `teamSwitch` and `selectPlayer`. It was also a never ending waitUntil loop on headless clients since 1.56.
- fixes a bug were CBA would delete all keyDown and keyUp events by other mods / missions when no additional sleep was added after the usual `waitUntil {!isNull findDisplay 46}`. very annoying.


esteldunedain:
:+1: This PR makes the backbone of CBA events system very similar to that of ACE events, originally written by @NouberNou. That framework has been extensively used over the last year, and aside from a few key improvements, which @commy2 ported, it really probed to be very adecuate from the get go.

In particular, having only one type of event handler, which can be invoqued by local or remote events, gives greater flexibility and makes it easier to use.

PR #285, opened 2016-03-01
Skip some unused compiles in XEH

commy2:
- don't compile events handlers except preStart when starting the game. They are discarded anyway, so this needlesly slows down starting the game.

- don't compile preStart event handlers in missions. They are never used while the game is running.


PabstMirror:
Yes!, I had been thinking of doing something like this. Faster at pre-start, faster at mission load.

PR #288, opened 2016-03-02
Port ACE delayed execution functions to CBA

esteldunedain:
This PR ports the following functions from ACE, with some optimizations. All of them run in non sched environment and are handled by the onFrame loop that handles PFH.

- `CBA_fnc_waitAndExecute`: Executes a code once in non sched environment with a given game time delay. It' based on a queue, so only the bare minimum of elements are evaluated each frame. This is a very efficient way of delaying execution, certainly much more than a PFH, specially for long delays. It's savegame compatible due to the use of `CBA_missionTime`.

- `CBA_fnc_execNextFrame`: A specialized version of `CBA_fnc_waitAndExecute` that delays execution a single frame.

- `CBA_fnc_waitUntilAndExecute`: Evaluates a condition code on each frame, and finally launches a sepparate statement code once when the condition returns true. This behaves like BIS `waitUntil`, but in non sched environment. Due to all condition being evaluated each frame, this function should be used only when necessary.

All of this are heavily used in ACE. We'll certainly switch to CBA's versions if this is merged.

### DIFF
--- a/addons/common/CfgFunctions.hpp
+++ b/addons/common/CfgFunctions.hpp
@@ -141,6 +141,12 @@ class CfgFunctions
                 description = "Drops a weapon.";
                 file = "\x\cba\addons\common\fnc_dropWeapon.sqf";
             };
+            // CBA_fnc_execNextFrame
+            class execNextFrame
+            {
+                description = "Add piece of code that will execute once on the next frame.";
+                file = "\x\cba\addons\common\fnc_execNextFrame.sqf";
+            };
             // CBA_fnc_findEntity
             class findEntity
             {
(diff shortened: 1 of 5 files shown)


addons/common/CfgFunctions.hpp
```diff
@@ -141,6 +141,12 @@ class CfgFunctions
                 description = "Drops a weapon.";
                 file = "\x\cba\addons\common\fnc_dropWeapon.sqf";
             };
+            // CBA_fnc_execNextFrame
+            class execNextFrame
+            {
```
jonpas: Opening brace on same line as class. I know it's not used in others in this file but it has to get converted anyways.

jonpas:
Totally awesome! :+1:

addons/common/CfgFunctions.hpp
esteldunedain: Ït's not used at all in cba_common, so I prefer to maintain  consistency

jonpas: @commy2 is changing it, he is currently going through entire cba_common. But sure.

esteldunedain:
Thanks!

PR #290, opened 2016-03-02
Fix versioning

PabstMirror:
Fix versioning system
`setVariable` doesn't seem to work on local variables.

### DIFF
--- a/addons/versioning/XEH_postInit.sqf
+++ b/addons/versioning/XEH_postInit.sqf
@@ -28,8 +28,8 @@ if (!SLX_XEH_DisableLogging) then {
         _class = (configFile >> "CfgPatches" >> (_dependencyInfo select 0));
         private _dependencyIsPresent = call compile format ["%1", (_dependencyInfo select 2)];
         if ((isNil "_dependencyIsPresent") || {!(_dependencyIsPresent isEqualType false)}) then {
-            //https://dev.withsix.com/issues/74516 - The code could return non-bool, if "true" is converted to "1" durring binerization
-            WARNING("Versioning Conditional return is bad" + str _x);
+            //https://dev.withsix.com/issues/74516 - The code could return non-bool, if "true" is converted to "1" durring binarization
+            WARNING("Versioning conditional return is bad" + str _x);
             _dependencyIsPresent = true;
         };
         if (_dependencyIsPresent) then {
--- a/addons/versioning/XEH_postInit.sqf
+++ b/addons/versioning/XEH_postInit.sqf
@@ -26,7 +26,7 @@ if (!SLX_XEH_DisableLogging) then {
         _mod = _x select 0;
         _dependencyInfo = _x select 1;
         _class = (configFile >> "CfgPatches" >> (_dependencyInfo select 0));
-        private _dependencyIsPresent = call compile (_dependencyInfo select 2);
+        private _dependencyIsPresent = call compile format ["%1", (_dependencyInfo select 2)];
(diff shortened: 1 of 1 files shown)


commy2:
:+1:

PabstMirror:
```
missionNamespace setVariable ["_dependencyIsPresent", _dependencyInfo select 2];
if (_dependencyIsPresent) then {
```
Old code was doing this, as far as I know we still need the call compile to convert the string into a bool.

commy2:
>I believe there is another method other than call compile, right?

`missionNamespace getVariable` can not replace all `call compile`. It only works when you have the name of a (global) variable.
`missionNamespace getVariable "CBA_Gvar" == call compile "CBA_Gvar" == CBA_Gvar`

It does not work when you compile a whole expression (like here) instead of just a var name.

NouberNou:
OK I think I need a better explanation of what those lines are doing, because I am confused as hell... what is `_dependencyInfo select 2`?

PabstMirror:
@NouberNou 
`_dependecyInfo` should be something like
```
["cba_main_a3",[1,0,0],"isClass(configFile >> 'CfgPatches' >> 'A3_Map_Stratis')"]
//but someone could put "true" which could mistakenly compile to `1` and get
["cba_xeh",[1,0,0],1]],["CBA",["cba_main",[1,0,0],1]
```

@ViperMaul `setVariable` won't work here because we need to convert the conditional string
```
"isClass(configFile >> 'CfgPatches' >> 'A3_Map_Stratis')"
```
into a boolean for the `if`, setVariable would just store the raw string into the var, it needs to be evaluated with `call`

Tested with a dependency of `XEH[] = {"cba_xeh", {6,0,0}, 1};` which was handled without script error, but will write warning to rpt.

addons/versioning/XEH_postInit.sqf
```diff
@@ -26,7 +26,12 @@ if (!SLX_XEH_DisableLogging) then {
         _mod = _x select 0;
         _dependencyInfo = _x select 1;
         _class = (configFile >> "CfgPatches" >> (_dependencyInfo select 0));
-        missionNamespace setVariable ["_dependencyIsPresent", _dependencyInfo select 2];
+        private _dependencyIsPresent = call compile format ["%1", (_dependencyInfo select 2)];
+        if ((isNil "_dependencyIsPresent") || {!(_dependencyIsPresent isEqualType false)}) then {
+            //https://dev.withsix.com/issues/74516 - The code could return non-bool, if "true" is converted to "1" durring binerization
```
Killswitch00: binarization

addons/versioning/XEH_postInit.sqf
```diff
@@ -26,7 +26,12 @@ if (!SLX_XEH_DisableLogging) then {
         _mod = _x select 0;
         _dependencyInfo = _x select 1;
         _class = (configFile >> "CfgPatches" >> (_dependencyInfo select 0));
-        missionNamespace setVariable ["_dependencyIsPresent", _dependencyInfo select 2];
+        private _dependencyIsPresent = call compile format ["%1", (_dependencyInfo select 2)];
+        if ((isNil "_dependencyIsPresent") || {!(_dependencyIsPresent isEqualType false)}) then {
+            //https://dev.withsix.com/issues/74516 - The code could return non-bool, if "true" is converted to "1" durring binerization
+            WARNING("Versioning Conditional return is bad" + str _x);
```
Killswitch00: conditional

PR #291, opened 2016-03-04
rework common module

commy2:
same as the other one, but friendlier to review (I hope)


commy2:
- move `CBA_fnc_parseYAML` to Hashes module. It returns a hash and Common is already pretty full.

commy2:
- rewrite `CBA_fnc_addMagazine`, `CBA_fnc_removeWeapon` and `CBA_fnc_removeMagazine`to work on remote units. (still won't work with `CBA_fnc_addWeapon` unfortunately)
- remove broken function `CBA_fnc_addMagazineVerified`. (doesn't and never did work with A3 inventory system)
- as replacement, add optional parameter to `CBA_fnc_addWeapon` and `CBA_fnc_addMagazine`; put item in nearby ground weapon holder if the inventory full
- add another optional parameter to `CBA_fnc_addMagazine`. Can add a magazine with a specific ammo count
- add optional parameter to `CBA_fnc_removeMagazine`. Can now remove a magazine with a specified ammo count.
- add `CBA_fnc_addItem` and `CBA_fnc_removeItem` to complement other functions (can't use `addWeapon` on first aid kits)

commy2:
- deprecate `CBA_fnc_addXCargoGlobal` and `CBA_fnc_addXCargoGlobal` functions. The local inventory space is no longer supported in A3. Instead, the local variants of the commands only work when the target object is local, but the command has global effect.
- add optional parameter to `CBA_fnc_addWeaponCargo` and `CBA_fnc_addMagazineCargo`; put items in nearby ground weapon holder if cargo is full
- add `CBA_fnc_addItemCargo` and `CBA_fnc_addBackpackCargo`

commy2:
- add `CBA_fnc_getItemConfig`. returns config path of any item class name given. (works with CfgWeapons, CfgMagaines and CfgGlasses)
- add `CBA_fnc_getObjectConfig`. returns config path of any object or object class name. (works with any OBJECT type in the game, CfgVehicles, CfgAmmo, CfgNonAIVehicles)
- remove broken function `CBA_fnc_determineMuzzles`, (wouldn't report muzzles correctly, probably due to incompatibility with A3)
- add replacement `CBA_fnc_getMuzzles` instead. returns all muzzle names of a weapon.
- add `CBA_fnc_getWeaponModes`. returns weapon modes of a weapon. optional parameter to include AI only modes.
- improve `CBA_fnc_getTurret` with new SQF commands. should be faster now. also supports invalid turret paths now and returns `<config null>` for them. also supports [-1] turret designation for driver turrets, which is occasionally used in A3 (old function would error)

commy2:
- add `CBA_fnc_vehicleRole`. returns the role of a soldier inside his current vehicle. (driver, gunner, commander, turret, cargo; empty string when on foot). the `assignedVehicleRole` SQF command is meant for AI and fails on clients on remote units in MP
- add `CBA_fnc_turretPath`. returns the current turret path of a soldier when inside a vehicle (reverse version of `turretUnit` SQF command
- add `CBA_fnc_turretPathWeapon`. returns the turret path of a weapon of a vehicle (reverse version of `weaponsTurret` SQF command.
- rewrite `CBA_fnc_getFirer`. Slightly faster. Now handles modified weapons instead of reling on config defaults. Now correctly handles soldiers and helicopters (would always return the same turret on them). also handles manual fire (will report the pilot instead of objNull, while keeping the weapons turret path).

commy2:
- sort functions in different categories, so they are easier to handle in the ingame functions viewer
- params and headers for other functions. small improvements with new commands and formatting
- remove team color synch handling internal functions from functions viewer by using PREP
- remove unreachable or no longer working, commented out code and delete no longer valid comments
- deprecate:  
`CBA_common_fnc_directCall` with way faster `CBA_fnc_directCall`  
`CBA_fnc_intToString` with (now) way faster `str`  
`CBA_fnc_defaultParam` which is now just a copy of a simple `param` command  
`CBA_fnc_systemChat` with `systemChat` (added at the end of A2 OA)  
`CBA_fnc_getPistol` with `handgunWeapon` from A3  
`CBA_fnc_createCenter` as `createCenter` is no longer needed in A3  
`CBA_fnc_players` with `allPlayers` (v1.44 A3)  
`CBA_fnc_locked` as it was for compatibility between TOH and A2, which are no longer supported  
- remove functions that were deprecated (with RPT log) 6+ years ago

commy2:
- improve `CBA_fnc_getGroupIndex` with A3 string handling commands and make it safe to work in scheduled environment (vehicle var name could be lost)
- add `CBA_fnc_getMagazineIndex` and `CBA_fnc_currentMagazineIndex` to return the magazine ID(s) of a magazine type / the current magazine of a unit

commy2:
- remove obsolete minit_functionsModule. No longer needed in A3 as functions are guaranteed to be set before preInit.

addons/common/backwards_comp.sqf
```diff
@@ -0,0 +1,78 @@
+
+CBA_fnc_createCenter = {
+    WARNING('Deprecated function used: OLD_FUNCTION');
```
Killswitch00: `OLD_FUNCTION` -> CBa_fnc_createCenter

addons/common/fnc_addMagazine.sqf
```diff
@@ -8,50 +8,75 @@ Description:
...
+    _return
 };
-if !(isClass (__cfg >> _item)) exitWith {
-    TRACE_2("Item not exist in Config",_unit,_item);
-    false
+
+if (_verify) then {
+    if (_unit canAdd _item) then {
+        if (_ammo < 0) then {
+            _unit addMagazine [_item, 1E6]; // addMagazine STRING is not synched when used on remote units. addMagazine ARRAY is.
```
Killswitch00: What's the purpose of adding one million magazines to a unit?

addons/common/fnc_addMagazine.sqf
```diff
@@ -8,50 +8,75 @@ Description:
...
+        if (isNull _weaponHolder || {_unit distance _weaponHolder > 2}) then {
+            _weaponHolder = createVehicle ["GroundWeaponHolder", [0,0,0], [], 0, "NONE"];
+            _weaponHolder setPosASL getPosASL _unit;
+        };
+
+        [_weaponHolder, _item] call CBA_fnc_addMagazineCargo;
+    };
+} else {
+    if (_ammo < 0) then {
+        _unit addMagazine [_item, 1E6]; // addMagazine STRING is not synched when used on remote units. addMagazine ARRAY is.
```
Killswitch00: Same here

addons/common/fnc_deleteEntity.sqf
```diff
@@ -17,36 +17,39 @@ Returns:
...
         };
-        deleteVehicle _this;
+        deleteVehicle _entity;
     };
     case "GROUP" : {
-        (units _this) call CBA_fnc_deleteEntity;
-        {deleteWaypoint _x} foreach (wayPoints _this);
-        deleteGroup _this;
+        (units _entity) call CBA_fnc_deleteEntity;
+        {deleteWaypoint _x} foreach (wayPoints _entity);
```
Killswitch00: `forEach`

addons/common/fnc_deleteEntity.sqf
```diff
@@ -17,36 +17,39 @@ Returns:
...
+
+[_this] params [["_entity", objNull, [objNull, grpNull, locationNull, "", []]]];
 
-switch (typeName _this) do {
+switch (typeName _entity) do {
     case "ARRAY" : {
         {
             _x call CBA_fnc_deleteEntity;
-        } foreach _this;
+        } foreach _entity;
```
Killswitch00: forEach

addons/common/fnc_deleteEntity.sqf
Killswitch00: forEach

addons/common/fnc_getAlive.sqf
```diff
@@ -5,42 +5,30 @@ Description:
...
     Rommel
-
 ---------------------------------------------------------------------------- */
-
 #include "script_component.hpp"
+SCRIPT(getAlive);
+
+[_this] params [["_entities", [], [objNull, grpNull, []]]];
+
+if (_entities isEqualType objNull) exitwith {
```
Killswitch00: exitWith

addons/common/fnc_getNearest.sqf
```diff
@@ -45,4 +48,5 @@ _return = if (count _this > 2) then {[]} else {objNull};
         };
     };
 } foreach _array;
```
Killswitch00: forEach

addons/common/fnc_isAlive.sqf
```diff
@@ -5,37 +5,45 @@ Description:
...
+        private _return = false;
+
         {
-            if (_x call CBA_fnc_isalive) exitwith {true};
-            false;
-        } foreach _this;
+            if (_x call CBA_fnc_isAlive) exitwith {
+                _return = true;
+            };
+        } foreach _entity;
```
Killswitch00: forEach

addons/common/fnc_setPos.sqf
```diff
@@ -17,51 +19,43 @@ Returns:
...
+
         {
-            private ["_txyz","_tx", "_ty", "_tz"];
-            _txyz = _x worldtomodel _ldp;
-            _tx = _dx + (_txyz select 0);
-            _ty = _dy + (_txyz select 1);
-            _tz = _dz + (_txyz select 2);
-            _x setpos [_tx,_ty,_tz];
+            _x setpos (_position vectorAdd (_x worldToModel _leaderPos));
         } foreach (units _entity);
```
Killswitch00: forEach

commy2: and setPos  ^^

Killswitch00: and worldToModel :-)

addons/common/fnc_addMagazine.sqf
commy2: It's not a million magazines, but one magazine with full ammo.
See: https://community.bistudio.com/wiki/addMagazine_array
`addMagazine STRING` fails on remote units. The alternate syntax `addMagazine ARRAY` does work on remote units.
I've tested this in local hosted MP and it's true. `addMagazine` does desynch the inventory causing weird issues when used on remote units. This is part of the fix to get this function working on all units regardless of locality.

addons/common/fnc_setPos.sqf
commy2: Already is. That's the deleted line

Killswitch00: Ah yes, my mistake.

addons/common/fnc_addMagazine.sqf
commy2: And yes. I've seen magazines with such ammo counts. One of these DayZ axes had 10.000 ammo in it's dummy magazine.

PR #302, opened 2016-03-12
Fix mikro converting true to 1

PabstMirror:
Ref #290
Trivial fix for mikero's dll 5.24 converting `"true"` to `1` in arrays, which will now causes minor rpt warning:

`"x\cba\addons\versioning\XEH_postInit.sqf:32","WARNING: Versioning conditional return is bad["CBA",["cba_main",[2,3,1],1]]"]`


esteldunedain:
:+1: Unfortunately this is needed

PR #305, opened 2016-03-30
Use cba event for cba_fnc_swtichPlayer

### LINKED ISSUE #304: Security Issue: fnc_switchPlayer
https://github.com/CBATeam/CBA_A3/blob/master/addons/common/fnc_switchPlayer.sqf#L69-L71

You are compiling code (not compile final) as a global varibale.
PublicVariable this to everyone
Then remoteexec the code

There is no real reason to have this as a function that is compiled on the fly, it doesn't change.

---------------------

In its current state its possible for hacker to compile a new function, sent it to all clients/servers.
Then remoteexec it. all the hacker just needs to worry about battleye script filters....

PabstMirror:
Close #304
Replace publicVariable and remoteExecCall with a CBA global event.

### DIFF
--- a/addons/common/CfgRemoteExec.hpp
+++ b/addons/common/CfgRemoteExec.hpp
@@ -1,9 +1,0 @@
-
-class CfgRemoteExec { // white listing currently broken! (1.56)
-    class Functions {
-        // Used by CBA_fnc_switchPlayer
-        class GVAR(setVehVarName) {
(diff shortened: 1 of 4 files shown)


commy2:
I suggest naming the event `CBA_setVehicleVarName` or maybe even `CBA_setVehicleVarNameGlobal` for concistency with the SQF command. It seems useful in other places.

PR #306, opened 2016-04-01
use mission event for PlayerConnected

commy2:
this replaces all `onPlayerConnected` stacked events with `PlayerConnected` mission events. it has the benefit that no mission or addon using a delayed `onPlayerConnected <CODE>` (SQF command) can overwrite our events.
WIP tag, because there currently is a bug, where the `PlayerConnected` mission event does not fire when `onPlayerConnected <CODE>` hasn't been used at least once. this is fixed according to the todays changelog, but the bug seems to be still present in the RC version uploaded to Steam.
I'll test if they fixed that next week.
backwards compatible with 1.54 / Linux, using the same method as in the `EachFrame` PR

### DIFF
--- a/addons/common/XEH_preInit.sqf
+++ b/addons/common/XEH_preInit.sqf
@@ -37,4 +37,9 @@ GVAR(addons) = _addons;
 // BWC
 #include "backwards_comp.sqf"
 
+// band aid - remove this once they fix PlayerConnected mission event handler
+// https://forums.bistudio.com/topic/143930-general-discussion-dev-branch/page-942#entry3003074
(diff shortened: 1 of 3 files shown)


commy2:
https://forums.bistudio.com/topic/143930-general-discussion-dev-branch/?p=3003074

esteldunedain:
This is missing the WIP label

commy2:
PlayerConnected is still broken. I could add a work around that could be easily removed should they ever fix it.

commy2:
could be merged now. maybe we can remove that band aid ... eventually

PR #312, opened 2016-04-21
Move addMissionEventHandler to postInit

PabstMirror:
Fixes semi-critical bug introduced in ~~1.60~~1.58: addMissionEventHandler ["EachFrame", has
no effect in preInit when called from singleplayer eden.

Before #303 PFEH's `FUNC(onFrame)` was installed in postInit, so this should work no differently than last release.

### DIFF
--- a/addons/common/XEH_postInit.sqf
+++ b/addons/common/XEH_postInit.sqf
@@ -1,6 +1,33 @@
 // #define DEBUG_MODE_FULL
 #include "script_component.hpp"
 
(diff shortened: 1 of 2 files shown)


commy2:
Is this in 1.58?

PabstMirror:
opps, this is for 1.58 which came out today, NOT 1.60

commy2:
roger. it did work when I made this weeks ago...

PR #314, opened 2016-04-24
CBA_fnc_globalEventJIP

commy2:
This adds `CBA_fnc_globalEventJIP` to the CBA events frame work.

It is basically the same as `CBA_fnc_globalEvent`, but all event calls (with their passed arguments) that happen during a mission are stored globally on a game logic. They are automatically called on clients that join in progress.

This is usefull for local commands that are not JIP persistent (e.g. `setIdentity`, `setFace`, `setRank`, etc. etc.)

The `_jipId` variable is an optional parameter that can be used to overwrite previous events with a new one. If no id is provided one will be automatically generated. This follows the same logic as the JIP ID in the `remoteExec` SQF command. (https://community.bistudio.com/wiki/remoteExec)

(While it uses `clientOwner`, which I consider to be broken in save games, this actually doesn't matter as the wrong owner IDs returned by it are still unique)

```
["print_test", {
    diag_log text str _this;
    _this spawn {
        sleep 5; // the chat is cleared after the loading screen in A3, so delay the message a bit
        systemChat str _this;
    };
```

`systemChat str (["print_test", "TEST_TEST_TEST"] call CBA_fnc_globalEventJIP)`


Glowbal:
Can a globalEventJIP be removed from the queue?

commy2:
No. At least not with API.

PR #317, opened 2016-04-25
CBA Settings frame work

commy2:
- Adds dynamically created "game options" that can also be set and enforced on dedicated servers without having to move around or edit any files.


addons/settings/fnc_forceSetting.sqf
```diff
@@ -0,0 +1,49 @@
...
+    profileNamespace setVariable [SETTING_KEY_SERVER_FORCE(_setting), true];
+};
+
+if (_source == "mission") then {
+    missionNamespace setVariable [SETTING_KEY_MISSION_FORCE(_setting), true, true];
+    profileNamespace setVariable [SETTING_KEY_MISSION_FORCE(_setting), true];
+};
+
+// send changed event, because the current active setting could have changed
+private _value = _setting call CBA_fnc_getSetting;
```
commy2: .

addons/settings/fnc_onButtonClick_configure.sqf
```diff
@@ -0,0 +1,125 @@
...
+    _ctrlClientButton ctrlEnable false;
+    _ctrlClientButton ctrlShow false;
+    _ctrlServerButton ctrlEnable false;
+    _ctrlServerButton ctrlShow false;
+    _ctrlMissionButton ctrlEnable false;
+    _ctrlMissionButton ctrlShow false;
+
+    // Terminate any waiting input actions
+    GVAR(waitingForInput) = false;
+    GVAR(input) = [];
```
commy2: todo, delete

commy2:
####Make a setting

####Adding Settings via addon config.

```
class CBA_Setting_Boolean_base;
class CBA_Setting_List_base;
class CBA_Setting_Slider_base;
class CBA_Setting_Color_base;

class CBA_Settings {
    class My_FancyAddon {
        displayName = "The Fanciest";

        // Checkbox type setting - can take the values true or false
        class My_ExampleSetting_1: CBA_Setting_Boolean_base {
            displayName = "Enable Custom Setting 1";
            tooltip = "";
            defaultValue = 0; // 0 - false (unchecked), 1 - true (checked)
        };

        // Slider type setting - always takes a numeric value between two extremes
        class My_ExampleSetting_2: CBA_Setting_Slider_base {
            displayName = "Custom Slider Setting 2";
            tooltip = "";
            min = 200; // low end of the slider
            max = 10000; // high end of the slider
            defaultValue = 5000; // slider defaults to this value if never touched by the player
        };

        // Listbox type setting - values can be numbers or strings
        class My_ExampleSetting_3: CBA_Setting_Slider_base {
            displayName = "Custom Listbox Setting 3";
            tooltip = "";
            values[] = {0,1}; // must be same size as  "valueNames"
            valueNames[] = {"disabled","enabled"}; // names shown in ingame settings editor
            defaultIndex = 0; // which index should be picked as default value?
        };

        // Color picker type setting
        class My_ExampleSetting_4: CBA_Setting_Color_base {
            displayName = "Custom Slider Setting 2";
            tooltip = "";
            defaultValue[] = {1,0,0}; // red - default color for this setting; 4d array means with alpha value
        };
    };
};
```

commy2:
Example - View distance slider done via init.sqf
```
init.sqf`
```
["CBA_SettingChanged", {
    params ["_setting", "_value"];

    if (_setting == "Commy_ViewDistance") then {
        setViewDistance _value;
    };

["Mission Settings", "Commy_ViewDistance", "SLIDER", "View Distance", [200, 15000, 5000, 0]] call cba_settings_fnc_create;
```

File: addons/settings/gui_initDisplay.sqf [hunk 3]
```diff
@@ -87,3 +87,9 @@ GVAR(missionSettingsTemp) = [] call CBA_fnc_createNamespace;
 
 // set this per script to avoid it being all upper case
 (_display displayCtrl IDC_TXT_FORCE) ctrlSetText localize LSTRING(force);
+
+// ----- hide presets menu
+private _ctrlPresetsGroup = _display displayCtrl IDC_PRESETS_GROUP;
+
+_ctrlPresetsGroup ctrlEnable false;
+_ctrlPresetsGroup ctrlShow false;
```
commy2: .

addons/settings/gui_initDisplay_disabled.sqf
```diff
@@ -7,10 +7,11 @@ private _ctrlToggleButton = _display displayCtrl IDC_BTN_CONFIGURE_ADDONS;
...
 private _ctrlMissionButton = _display displayCtrl IDC_BTN_MISSION;
+private _ctrlPresetsGroup = _display displayCtrl IDC_PRESETS_GROUP;
 
 _ctrlToggleButton ctrlEnable false;
 
 {
     _x ctrlEnable false;
     _x ctrlShow false;
-} forEach [_ctrlAddonsGroup, _ctrlClientButton, _ctrlServerButton, _ctrlMissionButton];
+} forEach [_ctrlAddonsGroup, _ctrlClientButton, _ctrlServerButton, _ctrlMissionButton, _ctrlPresetsGroup];
```
commy2: .

commy2:
### Creating a setting

- All settings are created with `CBA_Settings_fnc_init`. The function can be used anywhere and has local effects (function has to be executed on every machine, e.g. init.sqf or init box of a game logic etc.)

- Currently for different setting "types" are supported:
-- CHECKBOX : A checkbox. The resulting setting is a boolean.
-- SLIDER: Resulting setting is a number between a min and a max value. (e.g. view distance slider)
-- LIST: A dropdown list. Resulting value can be anything, but only one item can be selected at any time.
-- COLOR: Will create a "color picker" setting. The value will be an array representing a color. 
The array size can be 3 or 4, depending on the passed default value. The fourth element will represent the opacity ("alpha value").

#### Arguments of `CBA_Settings_fnc_init`
```
Parameters:
    _setting     - Unique setting name. Matches resulting variable name <STRING>
    _settingType - Type of setting. Can be "CHECKBOX", "LIST", "SLIDER" or "COLOR" <STRING>
    _title       - Display name or display name + tooltip (optional, default: same as setting name) <STRING, ARRAY>
    _category    - Category for the settings menu <STRING>
    _valueInfo   - Extra properties of the setting depending of _settingType. See examples below <ANY>
    _isGlobal    - true: all clients share the same state of the setting (optional, default: false) <ARRAY>
    _script      - Script to execute when setting is changed or forced. (optional) <CODE>
```

#### Example - View distance slider:
```
[
    "Commy_ViewDistance", // internal setting name, should always contain a tag!
    "SLIDER", // setting type
    "View Distance", // pretty name shown inside the ingame settings menu. can be stringtable entry.
    "My Mission Settings", // pretty name of the category where the setting can be found. can be stringtable entry.
    [200, 15000, 5000, 0], // data for this setting: [min, max, default, number of shown trailing decimals]
    nil, // "global" flag. set this to true to always have this setting synchronized between all clients in Multiplayer
    {  
        params ["_value"];
        setViewDistance _value;
    } // function that will be executed once on mission start and every time the setting is changed.
] call CBA_Settings_fnc_init;
```

esteldunedain:
:+1: For this being merged. Any estimate on when the next CBA will be released?

jonpas:
@esteldunedain When it will be ready! :laughing:

Nathan423:
Would it be possible to add an argument to be local in the sense that the value does not show up on the "server" or "mission" tab. This would be useful for adding specific options for a few people. For instance, I would like to have an option to view the groupID of units, but only for people white listed to access the function.

commy2:
@Nathan423 Please explain what you are trying to do.

Nathan423:
@commy2 Well, I have a part of a script that is checking for a said variable, the code is 
```
if (player getVariable ["uogm_mapMon_enableGroupID", false]) then {
    _marker setMarkerTextLocal format ["%1 | %2",(name _x), (groupID (group _x))];
} else {
    setMarkerTextLocal format ["%1",(name _x)];
};
```
And I would like to directly reference the variable instead of creating another variable and writing to that. For reference, here is my code to create the setting (executed locally on the client)
```
["uogm_mapMon_enableGroupID", "CHECKBOX", "Enable Group ID (Map Monitor)", "United Operations", [true], nil, {
        params ["_value"];
        player setVariable ["uogm_mapMon_enableGroupID", _value];
        systemChat str _value;
}] call CBA_Settings_fnc_init;
```

commy2:
I still don't understand. I don't see `GVAR` anywhere in your code.
Also, this is a weird place to ask this question. I don't receive mails from closed issues.

PR #319, opened 2016-04-26
Fix event error

PabstMirror:
With full modset I get nonstop errors after #314
```
Error select: Type Bool, expected Array,String,Config entry
File x\cba\addons\events\XEH_postInit.sqf, line 8
```

`allVariables cba_events_eventNamespaceJIP;` has "cba_xeh_isinitialized" and "cba_xeh_isprocessed" which are just `true`

So just a funny side effect of the XEH fallback loop running and setting unexpected variables on the logic.

### DIFF
--- a/addons/events/XEH_postInit.sqf
+++ b/addons/events/XEH_postInit.sqf
@@ -1,12 +1,13 @@
 #include "script_component.hpp"
 
 // execute JIP events after post init to guarantee execution of events added during postInit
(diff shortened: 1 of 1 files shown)


commy2:
That was exactly what the `isEqualTo EVENT_PVAR_STR` is for. I didn't even consider that there could be non arrays stored in that namespace. derp.

PR #321, opened 2016-04-30
Documentation cleanup, stage 1

Killswitch00:
Keep the first sentence of a function file's Description short
and separated from the rest of the description text. This first
sentence is used to auto-generate CfgFunctions.hpp by the
tools/funcitons_config.rb script, which is also used to generate
files needed to create the on-line HTML function documentation.


Killswitch00:
No change in functionality and no comments.

PR #324, opened 2016-05-02
Add "initRetro" event to addClassEventHandler

PabstMirror:
- Adds ability to run init code on all current and future objects via addClassEventHandler

Example usage:
```
["SettingsInitialized", {
    if (feature_enabled) then {
        ["CAManBase", "initRetro", {
            systemChat format ["Enable Feature On %1", _this]
    };
} call ace_common_fnc_addEventHandler;
````

Normal "init" would have no effect on existing units because unit's init event has already happened by the time the settingsInit event has happened.

### DIFF
--- a/addons/xeh/fnc_addClassEventHandler.sqf
+++ b/addons/xeh/fnc_addClassEventHandler.sqf
@@ -38,6 +38,13 @@ if (!GVAR(fallbackRunning) && {ISINCOMP(_className)}) then {
 // no such CfgVehicles class
 if (!isClass _config) exitWith {false};
 
+//Handle initReto event type:
+private _runRetroactiveInit = false;
+if (_eventName == "initRetro") then {
+    _runRetroactiveInit = true;
+    _eventName = "init";
+};
+
 _eventName = toLower _eventName;
 
 // no such event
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by esteldunedain** [comment]
Looks interesting

**Comment by Killswitch00** [comment]
Looks useful. I'm pondering if there are other naming possibilities for the event. Still thinking...

**Comment by commy2** [comment]
Why not make the `_runRetroactiveInit` flag an additional parameter of the function, instead of creating a new event name (that could potentially collide with something BI creates at some point even if very unlikely) ?

**Comment by esteldunedain** [comment]
I was thinking the same, but that function has two optional parameters already, and `_runRetroactiveInit` would only work for one event type.

I had two wacky alt proposals:
- Make the initRetro a wrapper function around `CBA_fnc_addClassEventHandler`, that just adds the init event and sepparately runs it for existing objects.
- Or more daringly: make initRetro behaviour the default "init" behaviour

**Comment by PabstMirror** [comment]
I was not sure about the event name; I think commy2 's idea would look like this:

Parameters:
5: _applyInitRetroactivly - Apply "init" event type on existing units that have already been initilized. [optional] <BOOLEAN> (default: false)

To Call:
```
["CAManBase", "init", {systemChat str _this}, true, [], true] call CBA_fnc_addClassEventHandler;
```
If we like this way more, I will make the changes

jonpas:
I like the optional argument more, don't really see a need for a completely separate event.

Killswitch00:
Same - I prefer to have it as an optional argument to CBA_fnc_addClassEventHandler.

PabstMirror:
Updated

commy2:
I added another check to not do weird stuff with events that are not "init" and "initPost"

esteldunedain:
As far as I can tell, it should be spelled `Retroactively`

PR #325, opened 2016-05-10
add CBA_fnc_binocularMagazine

commy2:
- Adds `CBA_fnc_binocularMagazine` which reports the units "binocular magazine" (laserbatteries in laserdesignator)
- Adds `CBA_fnc_addBinocularMagazine` which adds a magazine to the units binocular.
- Adds `CBA_fnc_removeBinocularMagazine` which removes the units "binocular magazine" (if any)

Those are the current best SQF work arounds for these missing commands. (as far as I can tell)

This follows the same logic as `primaryWeaponMagazine`, `handgunMagazine`, `removePrimaryWeaponItem`, `addSecondaryWeaponItem` etc.

### DIFF
--- a/addons/common/fnc_addBinocularMagazine.sqf
+++ b/addons/common/fnc_addBinocularMagazine.sqf
@@ -34,9 +34,6 @@ if (isNil "_ammo") then {
 };
 
 private _loadout = getUnitLoadout _unit;
-
-private _binocularInfo = _loadout param [8];
-_binocularInfo set [4, [_magazine, _ammo]];
-_loadout set [8, _binocularInfo];
+(_loadout select 8) set [4, [_magazine, _ammo]];
 
 _unit setUnitLoadout _loadout;
--- a/addons/common/fnc_addBinocularMagazine.sqf
+++ b/addons/common/fnc_addBinocularMagazine.sqf
@@ -3,6 +3,7 @@ Function: CBA_fnc_addBinocularMagazine
 
 Description:
     Adds a magazine to the units rangefinder.
+    Note that this breaks the unique magazine ids due to the usage of setUnitLoadout.
 
 Parameters:
     _unit     - A unit <OBJECT>
--- a/addons/common/fnc_addBinocularMagazine.sqf
+++ b/addons/common/fnc_addBinocularMagazine.sqf
@@ -1,0 +1,41 @@
+/* ----------------------------------------------------------------------------
+Function: CBA_fnc_addBinocularMagazine
+
+Description:
+    Adds a magazine to the units rangefinder.
+
+Parameters:
+    _unit     - A unit <OBJECT>
+    _magazine - The magazine to add <STRING>
+    _ammo     - Ammo count of the magazine (optional, default: full magazine) <NUMBER>
+
+Returns:
+    None
+
+Examples:
+    (begin example)
+        player addWeapon "Laserdesignator";
+        [player, "Laserbatteries"] call CBA_fnc_addBinocularMagazine;
+    (end)
+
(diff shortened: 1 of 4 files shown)


PabstMirror:
I was wondering if this would be any faster:
`(((getUnitLoadout  player) select 8) select 4) param [0, ""]`

but it's actually a little slower than the for loop; `getUnitLoadout` is an expensive command.

commy2:
Yeah. Was doing the same right now and had the same result.

`getUnitLoadout`/`setUnitLoadout` can be used to finally have a `addBinocularMagazine`variant

PabstMirror:
Unfortunitly that will break unique magazine ids
`player setUnitLoadout (getUnitLoadout player)`
will have different `magazinesDetail player`

:cry:

jonpas:
> Unfortunitly that will break unique magazine ids

As expected... it would be nice if we could get an optional parameter to keep magazine IDs.

commy2:
>Unfortunitly that will break unique magazine ids

hmm, true. I don't see any other way though

PabstMirror:
We could do something like a magID lookup map.
 - Player adds a unique mag `someUniqueMag:55`
 - `["someUniqueMag:55"] call CBA_fnc_getUniqueMagIndex` would return a index of 3 
 - We do the gear swap, player now has mag with ID `someUniqueMag:77`.
 - Gear swap would handle this and add entires to the map.
 - Calling `["someUniqueMag:77"] call CBA_fnc_getUniqueMagIndex`` would have same return of 3

@commy2 - What about using this for CBA_fnc_addBinocularMagazine:
```
params [["_unit", objNull, [objNull]], ["_magazine", "", [""]], ["_ammo", nil, [0]]];

if (!local _unit) exitWith {};
private _binocularWeapon = binocular _unit;
if (_binocularWeapon == "") exitWith {};

_unit removeWeapon _binocularWeapon;


if (isNil "_ammo") then {
   _ammo = getNumber (configFile >> "CfgMagazines" >> _magazine >> "count");
};
_unit addMagazine [_magazine, _ammo];


_unit addWeapon _binocularWeapon;
```

commy2:
`addMagazine` fails if you have no uniform/vest/backpack or they are full

commy2:
>We could do something like a magID lookup map.

Yeah, but anyone else using `setUnitLoadout` would still break it.

jokoho48:
i just took a look at the model and found 2 thing that you should improve, the Polycount in this parts of the model is way to high
http://puu.sh/oNmPh/b985503f8b.jpg
http://puu.sh/oNmTj/ad8adb18ac.png

commy2:
joko, I think you took a wrong turn somewhere.

jokoho48:
what the Fuck how did this work... Fu you github... i just realise what you mean dafug

jonpas:
I'll just leave this here for @jokoho48 http://static.flickr.com/3248/3285678255_1905fbfbff.jpg

addons/common/fnc_addBinocularMagazine.sqf
```diff
@@ -0,0 +1,39 @@
+/* ----------------------------------------------------------------------------
+Function: CBA_fnc_addBinocularMagazine
+
+Description:
+    Adds a magazine to the units rangefinder.
+    Note that this breaks the unique magazine ids due to the usage of setUnitLoadout.
```
Killswitch00: Add an empty line after the first sentence. The first sentence can(*) then be used to generate a `description` field for CfgFunctions entries.

(*) The old Ruby script that does this needs updating, but with the Description: paragraph in this form, automation is possible.

PR #328, opened 2016-05-14
add 'CBA_fnc_addPlayerEventHandler'

commy2:
Port (?) from ACE.
This is a framework to add event handlers based around properties of the players avatar that are pretty usefull, but unfortunately require polling.

Possible events:
        - `"unit"`       - player controlled unit changed
        - `"weapon"`     - currently selected weapon change
        - `"loadout"`    - players loadout changed
        - `"vehicle"`    - players current vehicle changed
        - `"turret"`     - position in vehicle changed
        - `"visionMode"` - player changed to normal/night/thermal view
        - `"cameraView"` - camera mode changed ("Internal", "External", "Gunner" etc.)
        - `"visibleMap"` - opened or closed map

### DIFF
--- a/addons/common/fnc_currentUnit.sqf
+++ b/addons/common/fnc_currentUnit.sqf
@@ -1,0 +1,20 @@
+/* ----------------------------------------------------------------------------
+Function: CBA_fnc_currentUnit
+
+Description:
+    Returns the currently controlled unit.
+    Different from "player" when remote controlling units via zeus.
+
+Parameters:
+    None
+
+Returns:
+    Currently controlled unit <OBJECT>
+
+Author:
+    commy2
+---------------------------------------------------------------------------- */
+#include "script_component.hpp"
+SCRIPT(currentUnit);
+
+missionNamespace getVariable ["bis_fnc_moduleRemoteControl_unit", player]
--- a/addons/common/fnc_currentUnit.sqf
+++ b/addons/common/fnc_currentUnit.sqf
@@ -2,8 +2,7 @@
 Function: CBA_fnc_currentUnit
 
 Description:
-    Returns the currently controlled unit.
-    Different from "player" when remote controlling units via zeus.
+    Returns the controlled unit. ("player" or remote controlled unit via zeus)
 
 Parameters:
     None
--- a/addons/events/fnc_addPlayerEventHandler.sqf
+++ b/addons/events/fnc_addPlayerEventHandler.sqf
@@ -34,7 +34,9 @@ SCRIPT(addPlayerEventHandler);
 
 params [["_type", "", [""]], ["_function", {}, [{}]]];
 
-private _id = switch (toLower _type) do {
+_type = toLower _type;
+
+private _id = switch (_type) do {
 case ("unit"): {
     [QGVAR(unitEvent), _function] call CBA_fnc_addEventHandler // return id
 };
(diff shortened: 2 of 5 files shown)


esteldunedain:
Mmm, I celebrate that this are getting added to CBA. However, I'm not really fond of the way it's executed for two reasons:

1- Most of the events are trivially cheap to evaluate, but instead of adding them to a single loop that always run (like in ACE), each one is tucked in it's own EH function, making them slower. This is specially true if you want to use many or all of them (like in ACE). I would instead recommend that either all of them are installed by default, or all are installed when the first is.

2- That a new installer function is required `CBA_fnc_addPlayerEventHandler` instead of `CBA_fnc_addLocalEvent`

commy2:
All checks could certainly be moved in one EachFrame event.

I don't want to have this loop running by default though. Most mods won't make use of any of these, so I want to avoid being critiqued for adding 'pointless' checks every frame. Imagine a community using CBA only for JR and CUP (Life + CUP) or even a mil sim comm with only CBA + TFAR.
A wrapper would still be needed in that case and I don't really see a problem with that.

bux578:
How about running everything in one loop and have developers "opt-in" to the events?

PabstMirror:
We could dynamically build up the polling function with `compile`.

commy2:
>How about running everything in one loop and have developers "opt-in" to the events?

Using the `CBA_fnc_addPlayerEventHandler` function once would be a way to opt-in I guess.

>We could dynamically build up the polling function with compile.
Sure. That would mean compiling 8 functions at mission start though (because if you use all, they are all added after another)

esteldunedain:
Thanks @commy2 :+1:

addons/common/fnc_currentUnit.sqf
```diff
@@ -0,0 +1,20 @@
+/* ----------------------------------------------------------------------------
+Function: CBA_fnc_currentUnit
+
+Description:
+    Returns the currently controlled unit.
+    Different from "player" when remote controlling units via zeus.
```
Killswitch00: Either shorten to one sentence or separate with empty line for possible future improvements to documentation generation.

addons/events/fnc_addPlayerEventHandler.sqf
```diff
@@ -0,0 +1,145 @@
...
+    [QGVAR(cameraViewEvent), _function] call CBA_fnc_addEventHandler // return id
+};
+case ("visiblemap"): {
+    [QGVAR(visibleMapEvent), _function] call CBA_fnc_addEventHandler // return id
+};
+default {-1};
+};
+
+// add loop for polling if it doesn't exist yet
+if (isNil QGVAR(pollingLoopAdded) && {_id != -1}) then {
```
Killswitch00: Keep track of the EachFrame handler id here instead, so that the polling function can be removed if/when all player event handlers are removed again.

addons/events/fnc_addPlayerEventHandler.sqf
```diff
@@ -0,0 +1,145 @@
...
+    GVAR(oldWeapon) = "";
+    GVAR(oldLoadout) = [];
+    GVAR(oldLoadoutNoAmmo) = [];
+    GVAR(oldVehicle) = objNull;
+    GVAR(oldTurret) = [];
+    GVAR(oldVisionMode) = -1;
+    GVAR(oldCameraView) = "";
+    GVAR(oldVisibleMap) = false;
+
+    addMissionEventHandler ["EachFrame", {
```
Killswitch00: Keep track of the EachFrame handler id so that the polling function can be removed if/when all player event handlers are removed again.

PR #329, opened 2016-05-14
Handle nil/non-bool returns for keyHandlers

PabstMirror:
When user added key handler returns bad data, arma throws a generic unhelpful error.
```
Error in expression <#line 1 "x\cba\addons\events\fnc_keyHand>
 Error position: <#line 1 "x\cba\addons\events\fnc_keyHand>
 Error Type Script, expected Bool
```

`_blockInput` is returned at bottom of our function, so if a handler returns bad data, our function will return it as well.

`LOG(` won't show under normal debug; we could swap to `WARNING(` if we want to give feedback (this is what we had before #284)?

keyHandlerUp is similar, but doesn't return the results, so it shouldn't matter there.

### DIFF
--- a/addons/events/fnc_keyHandlerDown.sqf
+++ b/addons/events/fnc_keyHandlerDown.sqf
@@ -49,10 +49,11 @@ private _blockInput = false;
 
             _blockInput = _params call _code;
 
-            if ((isNil "_blockInput") || {!(_blockInput isEqualType false)}) then {
-                LOG(PFORMAT_2("Keybind Handler returned nil or non-bool", _x, _blockInput));
-                _blockInput = false; //Set to a bool, so we don't return garbage at end
-            };
+            #ifdef DEBUG_MODE_FULL
+                if ((isNil "_blockInput") || {!(_blockInput isEqualType false)}) then {
+                    LOG(PFORMAT_2("Keybind Handler returned nil or non-bool", _x, _blockInput));
+                };
+            #endif
         };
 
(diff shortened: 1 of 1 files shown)


Killswitch00:
If a `WARNING` could show enough info to help indicate which handler (read: which addon/mod component) that's broken, that would be useful.

commy2:
I think it would be better if we replace L:70
`_blockInput`
with:
`!isNil "_blockInput" && {_blockInput isEqualType true}`

Benefits:
- it's closer to all other event handlers that use a boolean `true` to overwrite something - if you return anything else than a boolean or `nil` it is treadted just like `false`
- no RPT spam for addons that probably won't be updated (ever again or soon)
- you don't have to check this condition multiple times (for every active keybinding) every frame you hold down at least one key.

PabstMirror:
@Killswitch00 the logging would work like this:
```
["A", "B", ["C", "D"], { systemChat "x";}, {}, [16, [false, false, false]]] call cba_fnc_addKeybind;
->
...keyHandlerDown.sqf:54","Keybind Handler returned nil or non-bool:  _x=a_b_down,  _blockInput=any"]
```
`a_b` should make it easy to id the mod.  But I agree with commy2, and I don't think it's worth throwing an rpt for each keypress when an addon returns `nil`. I left the loggin in, but it would only run with `DEBUG_MODE_FULL`

Killswitch00:
Yeah, commy2 made good points and I agree with his no-spam suggestion

PR #331, opened 2016-05-17
Replace replace

BaerMitUmlaut:
This PR replaces the replace function. The old one does not work properly with some characters (for example öüäß) due to buggy BI commands:
```
("ü" splitString "") apply {toArray _x}
-> [[65533],[65533]]
"abcßdef" find "ß"
-> 3
"abcßdef" find "d"
-> 5
```
It also adds params checking and doesn't use any `toArray`/`toString` stuff anymore.

### DIFF
--- a/addons/strings/fnc_replace.sqf
+++ b/addons/strings/fnc_replace.sqf
@@ -27,14 +27,14 @@ Author:
 SCRIPT(replace);
 
 params [["_string", "", [""]], ["_find", "", [""]], ["_replace", "", [""]]];
-private ["_offset", "_index"];
-if (_find == "" || {_replace find _find != -1}) exitWith {_string};
-
-_offset = count (_find splitString "");
+if (_find == "") exitWith {_string}; // "1" find "" -> 0
+private _result = "";
+private _offset = count (_find splitString "");
 
 while {_string find _find != -1} do {
-    _index = _string find _find;
-    _string = (_string select [0, _index]) + _replace + (_string select [_index + _offset]);
+    private _index = _string find _find;
+    _result = _result + (_string select [0, _index]) + _replace;
+    _string = _string select [_index + _offset];
 };
 
-_string
+_result + _string
--- a/addons/strings/fnc_replace.sqf
+++ b/addons/strings/fnc_replace.sqf
@@ -19,35 +19,22 @@ Example:
     (end)
 
 Author:
-    jaynus
+    BaerMitUmlaut
 --------------------------------------------------------------------------- */
 
 #include "script_component.hpp"
 
 SCRIPT(replace);
 
-// ----------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


addons/strings/fnc_replace.sqf
```diff
@@ -19,35 +19,22 @@ Example:
...
 --------------------------------------------------------------------------- */
 
 #include "script_component.hpp"
 
 SCRIPT(replace);
 
-// ----------------------------------------------------------------------------
+params [["_string", "", [""]], ["_find", "", [""]], ["_replace", "", [""]]];
+private ["_offset", "_index"];
+if (_find == "" || {_replace find _find != -1}) exitWith {_string};
```
commy2: This is not needed. It's a tautology and the same as the while loop condition.

BaerMitUmlaut: This checks for infinite loops when you would do something like:
```
["Hello", "e", "ee"] call CBA_fnc_replace;
```

addons/strings/fnc_replace.sqf
```diff
@@ -19,35 +19,22 @@ Example:
...
 
-params ["_string","_pattern","_replacement"];
-private["_i", "_cp", "_findIndex", "_stringArray", "_replaceArray", "_returnArray"];
+_offset = count (_find splitString "");
```
Glowbal: Can this not be simply `count _find`?

jonpas: `count STRING` works. Except if this is to circumvent those unsupported characters.

commy2:
Before:
```
["TEST","E","EA"] call CBA_fnc_replace
"TEAST"
```

After:
```
["TEST","E","EA"] call CBA_fnc_replace
"TEST"
```

addons/strings/fnc_replace.sqf
commy2: The point of `_offset` is, to find ASCII (?) characters that all string commands have problems with.

```
`"ö" splitString ""`
-> ["�","�"]
```

BaerMitUmlaut:
This works now:
```
["TEST","E","EA"] call CBA_fnc_replace
-> "TEAST"
```

commy2:
Passes the test now
```
0 spawn compile preprocessFileLineNumbers  "\x\cba\addons\strings\test_strings.sqf"
```

The benefit of this PR:
Before:
```
["Mörser","ö","oe"] call CBA_fnc_replace
-> "Moeser"
```

After:
```
["Mörser","ö","oe"] call CBA_fnc_replace
-> "Moerser"
```

We should probably add that and `"TEAST"` to the test.

commy2:
Performance:
`["Mörser","ö","oe"] call CBA_fnc_replace`

Before: 0.0525 ms
After: 0.0432 ms

Killswitch00:
Agree - please add the two new tests to the unit test suite.

BaerMitUmlaut:
Those two tests have been added.

PR #333, opened 2016-05-19
Update AI functions

SilentSpike:
In creating some zeus AI modules for ACE I noticed that the CBA AI functions didn't behave exactly as expected and made this PR to change that.

The changes can be found in the commit descriptions, but to summarise:
- Task functions will now clear all previous waypoints before assigning the task's waypoints (a new function `fnc_clearWaypoints` was introduced to correctly clear all existing waypoints without issue).
- The `fnc_searchNearby` function can now be called and doesn't have to be spawned.
- The `fnc_taskPatrol` function now completes the patrol by placing the cycle waypoint back at the start instead of randomly.
- Functions `fnc_taskDefend` and `fnc_searchNearby` were rewritten to be both more readable and optimized.
- Function `fnc_taskDefend` now handles building positions slightly differently such that if multiple groups are tasked to defend overlapping areas the positions are shared between all of the units, not just the units within each group. Meaning that multiple units will not not occupy the same position unless all positions are occupied already.
- Function `fnc_searchNearby` now handles cases where units are killed while searching the building better. The group will collectively try to search every position in the building, whereas previously some positions could be overlooked because they were assigned to a dead unit (since the units array wasn't updating correctly). Similarly the function will handle cases where the leader is killed better, returning units to their new leader after the search is complete.

### DIFF
--- a/addons/ai/fnc_searchNearby.sqf
+++ b/addons/ai/fnc_searchNearby.sqf
@@ -64,7 +64,11 @@ if ((leader _group) distanceSqr _building > 250e3) exitwith {_group lockWP false
     };
 
     // Once units are all finished searching return to previous tasks
-    waitUntil {sleep 3; {unitReady _x} count (units _group) >= count (units _group) - 1};
+    waitUntil {
+        sleep 3;
+        private _units = (units _group) - [_leader];
+        ({unitReady _x} count _units) >= count _units
+    };
     {
         _x doFollow (leader _group); // Not using _leader in case of death
     } forEach (units _group);
--- a/addons/ai/fnc_searchNearby.sqf
+++ b/addons/ai/fnc_searchNearby.sqf
@@ -22,15 +22,20 @@ Author:
 
 params ["_group"];
 _group = _group call CBA_fnc_getGroup;
-_group lockWP true;
 
 private _building = nearestBuilding (leader _group);
-if ((leader _group) distanceSqr _building > 250e3) exitwith {_group lockWP false};
+if ((leader _group) distanceSqr _building > 250e3) exitwith {};
 
 [_group,_building] spawn {
     params ["_group","_building"];
     private _leader = leader _group;
-    private _behaviour = behaviour _leader;
+
+    // Add a waypoint to regroup after the search
+    _group lockWP true;
+    private _wp = _group addWaypoint [getPos _leader, 0, currentWaypoint _group];
+    private _cond = "({unitReady _x || !(alive _x)} count thisList) == count thisList";
+    private _comp = format ["this setFormation %1; this setBehaviour %2; deleteWaypoint [group this, currentWaypoint (group this)];",formation _group,behaviour _leader];
+    _wp setWaypointStatements [_cond,_comp];
 
     // Prepare group to search
     _group setBehaviour "Combat";
@@ -56,22 +61,10 @@ if ((leader _group) distanceSqr _building > 250e3) exitwith {_group lockWP false
             if (unitReady _x) then {
                 private _pos = _positions deleteAt 0;
                 _x commandMove _pos;
+                sleep 2;
             };
         } forEach _units;
-
-        // Provide time for orders to be carried out
-        sleep 10;
     };
 
-    // Once units are all finished searching return to previous tasks
-    waitUntil {
-        sleep 3;
-        private _units = (units _group) - [_leader];
-        ({unitReady _x} count _units) >= count _units
-    };
-    {
-        _x doFollow (leader _group); // Not using _leader in case of death
-    } forEach (units _group);
-    _group setBehaviour _behaviour;
     _group lockWP false;
 };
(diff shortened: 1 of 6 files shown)


addons/ai/fnc_taskDefend.sqf
```diff
@@ -25,78 +25,84 @@ Returns:
...
+
+                    _unit doMove _pos;
                     sleep 5;
-                    waituntil {unitready (_this select 0)};
-                    (_this select 0) disableai "move";
-                    dostop _this;
-                    waituntil {not (unitready (_this select 0))};
-                    (_this select 0) enableai "move";
+                    waituntil {unitReady _unit};
+                    _unit disableAI "move";
```
SilentSpike: Just a note that this function should definitely be updated to use the new feature added in today's [dev branch changelog](https://forums.bistudio.com/topic/140837-development-branch-changelog/?p=3027032):

`Ability to enableAI / disableAI "PATH", which stops the AI’s movement but not the target alignment`

However I have no idea if the dev branch still will appear in the next verison of arma, or the one after

Glowbal: Probably something for a future PR when we do know in which version it will be available?

addons/ai/fnc_searchNearby.sqf
```diff
@@ -9,51 +9,56 @@ Parameters:
...
+    _group setBehaviour "Combat";
+    _group setFormDir ((leader _group) getDir _building);
+
+    // Search while there are still available positions
+    private _positions = _building buildingPos -1;
+    while {!(_positions isEqualTo [])} do {
+        // Abort search if the group has no units left
+        if ((units _group) isEqualTo []) exitWith {};
+
+        // Send all available units to the next available position
```
Killswitch00: Change in behaviour: the old version does not send the first unit (leader?) into the building, but the new one does.

addons/ai/fnc_taskDefend.sqf
```diff
@@ -25,78 +25,84 @@ Returns:
...
-private ["_count", "_list", "_list2", "_units", "_i"];
-_statics = [_position, vehicles, _radius, {(_x iskindof "StaticWeapon") && {(_x emptypositions "Gunner" > 0)}}] call CBA_fnc_getnearest;
-_buildings = _position nearObjects ["building",_radius];
-_units = units _group;
-_count = count _units;
+private _statics = _position nearObjects ["StaticWeapon", _radius];
+private _buildings = _position nearObjects ["Building", _radius];
 
+// Filter out occupied statics
+_statics = _statics select { (_x emptyPositions "Gunner") > 0 };
```
Killswitch00: Not compatible with the MacOS/Linux ports, currently at v1.54. Needs workaround like commy2 did recently (the "linux.pbo" addon)

addons/ai/fnc_taskDefend.sqf
```diff
@@ -25,78 +25,84 @@ Returns:
...
-                };
-                _array = _building getvariable "CBA_taskDefend_positions";
-            };
-            if (count _array > 0) then {
-                _p = (_building getvariable "CBA_taskDefend_positions") call BIS_fnc_selectRandom;
-                _array = _array - [_p];
-                if (count _array == 0) then {
+        // 93% chance to occupy a random nearby building position
+        if ((random 1 < 0.93) && { !(_buildings isEqualto []) }) then {
+            private _building = selectRandom _buildings;
```
Killswitch00: Ports compatibiltiy issue here aswell - `selectRandom` is 1.56+

SilentSpike:
I hadn't realize CBA is now supporting two versions of arma, well that sucks :laughing:

Will amend all of the above!

addons/ai/fnc_searchNearby.sqf
SilentSpike: Good catch, though I'm going to write it such that the leader will only stay outside when the group has more than 2 units :+1:

SilentSpike: So I've noticed that the leader will continue on to the group's next waypoint if he isn't searching the building himself.

Should I change it to make him wait outside until the search is done? Otherwise he could run quite far ahead of the group.

Killswitch00: Yeah, that's better than him going on without his team.

SilentSpike: Unfortunately I can't find a reliable & safe way to do this.

I could create a waypoint - which introduces the problem of trying to safely remove the waypoint once it has served its purpose.

While `doStop` permanently stops the unit as far as I can tell - which can be avoided in future once the `disableAI "PATH"` functionality is in the release version of arma.

SilentSpike: Okay, got the waypoint solution working cleanly - it's actually preferable to using AI commands because it also takes care of spare group members too and allows me to put the post-search code into the completion statement instead of running a loop :smile:

PR #340, opened 2016-05-26
Add support for sending events to client IDs

PabstMirror:


jonpas:
Would this be better if incorporated into `targetEvent` rather?

commy2:
This has one Problem `clientOwner` reports totally wrong ID is reloaded multiplayer games.

To solve this, I suggest pubVaring the server the actual client id from a `PlayerConnected` mission event.

PabstMirror:
Now syncs clientID, probably needed to do anyway because `clientOwner` is 1.56.
We could merge this into targetEvent, but the way they work is quite different and I think it would get messy.
OwnerEvent can do client->client with no sqf running on server.

addons/events/XEH_postInit.sqf
```diff
@@ -11,3 +11,18 @@
...
+if (isServer) then {
+    GVAR(clientID) = [0, 2] select isMultiplayer;
+    addMissionEventHandler ["PlayerConnected", {
+        params ["_id", "_uid", "_name", "_jip", "_owner"];
+        TRACE_5("PlayerConnected eh",_id,_uid,_name,_jip,_owner);
+
+        if (_owner != 2) then {
+            GVAR(clientID) = _owner;
+            _owner publicVariableClient QGVAR(clientID);
+            GVAR(clientID) = [0, 2] select isMultiplayer;
```
commy2: Variable seems useful. Maybe name it "CBA_clientID" thereby make it "official" API, just like `CBA_missionTime` ?

addons/events/XEH_preInit.sqf
```diff
@@ -13,6 +16,7 @@ if (isServer) then {
 
 // can't add at preInit
 0 spawn {
+    QGVAR(clientID) addPublicVariableEventHandler {TRACE_1("clientID PVEH",_this);};
```
commy2: I don't think this is needed.

addons/events/XEH_postInit.sqf
```diff
@@ -11,3 +11,18 @@
...
+
+
+if (isServer) then {
+    CBA_clientID = [0, 2] select isMultiplayer;
+    addMissionEventHandler ["PlayerConnected", {
+        params ["_id", "_uid", "_name", "_jip", "_owner"];
+        TRACE_5("PlayerConnected eh",_id,_uid,_name,_jip,_owner);
+
+        if (_owner != 2) then {
+            CBA_clientID = _owner;
```
Killswitch00: Are `PlayerConnected` handlers executed in scheduled or unscheduled space? (If scheduled, I see a race condition with the `CBA_clientID` variable here)

jonpas:
I am still wondering why that's not incorporated into `CBA_fnc_targetEvent`. A simple type check, would make it much simpler to use.

Killswitch00:
Merge reverted - discussion on this PR had not finished yet. PabstMirror, please re-submit the function for further discussion.

PR #352, opened 2016-06-02
Add support for sending events to client IDs (redo)

PabstMirror:
Redo of #340


commy2:
Confirmed working. Tested in 3 way local MP with resumed mission.

PR #354, opened 2016-06-03
Add meta.cpp file

Killswitch00:
Metadata file used by the server browser in the A3 game launcher
to pull mods used on a server directly from the Steam Workshop.
This changeset was originally submitted by bux578 but went missing
along the way.


bux578:
:disappointed:

PR #356, opened 2016-06-04
Add override capability to fnc_taskAttack

Wired-Tiger:
When playing around with the attack, defend, and patrol functions I noticed that if the group was already set to either patrol or defend that their behavior would change but not move or change course. Probably not needed often, but still nice to have the capability of redirecting the group.

### DIFF
--- a/addons/ai/fnc_taskAttack.sqf
+++ b/addons/ai/fnc_taskAttack.sqf
@@ -31,6 +31,7 @@ params ["_group","_position", ["_radius",0]];
 _group = _group call CBA_fnc_getGroup;
 if !(local _group) exitWith {}; // Don't create waypoints on each machine
 
-[_group] call CBA_fnc_clearWaypoints; // Clear existing waypoints
+// Override any previously set waypoints
+[_group] call CBA_fnc_clearWaypoints;
 
 [_group, _position, _radius, "SAD", "COMBAT", "RED"] call CBA_fnc_addWaypoint;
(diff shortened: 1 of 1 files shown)


addons/ai/fnc_taskAttack.sqf
```diff
@@ -31,4 +31,6 @@ params ["_group","_position", ["_radius",0]];
 _group = _group call CBA_fnc_getGroup;
 if !(local _group) exitWith {}; // Don't create waypoints on each machine
 
+[_group] call CBA_fnc_clearWaypoints; // Clear existing waypoints
```
Glowbal: A comment should explain why a piece of code does what it does. Ie a very brief summary of what you wrote in the PR.

Wired-Tiger: I'm a little confused. Is that not what I did at the end with " // Clear existing waypoints" ? Should it be on top instead?

Killswitch00: `// Clear existing waypoints` tells the reader _what_ the line does, not _why_ it is being done. (and it feels a bit redundant, since the name of the function being called hints at that).

Wired-Tiger: Okay. Thanks for the clarification!

Killswitch00:
This should be optional so as to not change how the function behaves for existing uses of it. (Keep backwards compatibility)

SilentSpike:
Late to the party, but it should be noted that any AI function used after taskDefend probably won't work 100% as expected due to the use of `doStop`.

Now that CBA is supporting 1.62 I'll change that to `disableAI "PATH"` in my next PR related to the AI functions. Will see if I can perhaps make them play better together when used in sequence too.

PR #359, opened 2016-06-04
More author[] -> authors[] 1.60 fixes

Killswitch00:


optionals/cache_disable/config.cpp
```diff
@@ -7,7 +7,8 @@ class CfgPatches {
         requiredVersion = REQUIRED_VERSION;
         requiredAddons[] = {"CBA_Extended_EventHandlers", "CBA_Main"};
         version = VERSION;
-        author[] = {"Sickboy"};
+        author = "$STR_CBA_Author";
```
jonpas: `CSTRING(Author)`?

Killswitch00: That would become `$STR_CBA_cache_disable_Author`, wouldn't it?

jonpas: Right, that is true. My bad, I am used to having it in one PBO and use `ECSTIRNG(pbo,Author)`.

optionals/jr_disable_long_scopes_on_short_mg_rail/config.cpp
```diff
@@ -9,7 +9,8 @@ class CfgPatches
 		requiredVersion = REQUIRED_VERSION;
 		requiredAddons[] = {"cba_jr"};
 		version = VERSION;
-		author[] = {"Robalo"};
+		author = "$STR_CBA_Author";
+		authors[] = {"Robalo"};
         authorUrl = "https://github.com/CBATeam/CBA_A3";
```
commy2: We should convert all tabs to 4 spaces just to avoid this ^

Killswitch00: Fixed

PR #360, opened 2016-06-04
Update fnc_taskPatrol func header

Wired-Tiger:
I noticed that an error is thrown from fnc_addWaypoint if the position and radius wasn't passed so they would not be optional.

Also, could someone explain to me the point of line 49: private _this += _this;
I've been trying to understand the point of it, but I've removed it and the function works the same with or without it from what I can tell. Is it left over code by chance?

### DIFF
--- a/addons/ai/fnc_taskPatrol.sqf
+++ b/addons/ai/fnc_taskPatrol.sqf
@@ -6,10 +6,10 @@ Description:
 
 Parameters:
     - Group (Group or Object)
+
+Optional:
     - Position (XYZ, Object, Location or Group)
     - Radius (Scalar)
-
-Optional:    
     - Waypoint Count (Scalar)
     - Waypoint Type (String)
     - Behaviour (String)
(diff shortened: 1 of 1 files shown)


esteldunedain:
`_this += _this` creates a copy of the parameters array, so the final element can be deleted without afecting the original data. This is important because arrays are always passed as reference to functions.

Wired-Tiger:
@esteldunedain Thank you for the explanation. I think what is throwing me off is the scope of _this. I was thinking that _this was equal to [1,2,3,1,2,3,4] instead of [1,2,4] for every instance after that particular code block.... Looking over this I realized that sqf doesn't have a "+=" command for adding like other languages. Sorry about that.

Killswitch00:
The position and group parameters _are_ optional and there is no need to change the function header for that reason. Changing the default radius from 100 to 50 in the example call could be useful for the understanding of the function.

Killswitch00:
The function header represents how the function is supposed to (and AFAIK has, at some point in time) work(ed), so the way forward here is to fix the now broken function.

addons/ai/fnc_taskPatrol.sqf
```diff
@@ -47,6 +47,11 @@ _position = _position call CBA_fnc_getPos;
 [_group] call CBA_fnc_clearWaypoints;
 
 private _this =+ _this;
+switch (count _this) do {
+    case 1 : {_this append [_position, _radius]};
+    case 2 : {_this pushback _radius};
```
Killswitch00: `pushBack` (camelCase)

PR #363, opened 2016-06-04
make 'CBA_fnc_players' not return headless clients

### LINKED ISSUE #353: CBA_fnc_players might still be useful
```
allPlayers - The command also returns all connected headless clients.
_justPlayers = allPlayers - entities "HeadlessClient_F"
``` 

This is a discussion if getting back all headless clients is an expected output when using this command. 

I'd imagine that in most scenarios one would **not** want to run some code on HC's. 
My suggestion is to add back CBA_fnc_players and filter out headless clients before returning the array. Perhaps also check if the virtual entity spectators are included here.

If one wants the HC's, he/she can run the command
`entities "HeadlessClient_F"`

There's also another comment that I find weird:
```
NOTE: In player hosted game, the complete array of allPlayers may get delayed at the start
Use BIS_fnc_listPlayers if you need it earlier
```

commy2:
-closes #353

`allPlayers` and `BIS_fnc_listPlayers` report players and headless clients. This function does not.
Can't use `allPlayers`, because apparently it reports not all units in a local hosted MP game.

### DIFF
--- a/addons/common/fnc_players.sqf
+++ b/addons/common/fnc_players.sqf
@@ -21,4 +21,4 @@ Author:
 #include "script_component.hpp"
 SCRIPT(players);
 
-(allUnits + allDead) select {isPlayer _x && {!(_x isKindOf "HeadlessClient_F")}}
+[allUnits + allDead, {isPlayer _x && {!(_x isKindOf "HeadlessClient_F")}}] call BIS_fnc_conditionalSelect
--- a/addons/common/fnc_players.sqf
+++ b/addons/common/fnc_players.sqf
@@ -1,0 +1,24 @@
+/* ----------------------------------------------------------------------------
+Function: CBA_fnc_players
+
+Description:
+    Reports all (human) player objects. Does not include headless client entities.
+
+Parameters:
+    None
+
+Returns:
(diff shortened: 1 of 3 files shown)


addons/common/fnc_players.sqf
```diff
@@ -0,0 +1,24 @@
...
+        [] call CBA_fnc_players
+    (end)
+
+Author:
+    commy2
+---------------------------------------------------------------------------- */
+#include "script_component.hpp"
+SCRIPT(players);
+
+(allUnits + allDead) select {isPlayer _x && {!(_x isKindOf "HeadlessClient_F")}}
```
Killswitch00: `select <CODE>` not usable with the ports, currently at 1.54

addons/common/fnc_players.sqf
```diff
@@ -0,0 +1,24 @@
+/* ----------------------------------------------------------------------------
+Function: CBA_fnc_players
+
+Description:
+    Reports all (human) player objects. Does not include headless client entities.
+
```
Killswitch00: Add a short paragraph after the one-liner description that explains that this is different from`allPlayers`/`BIS_fnc_listPlayers`.

PR #364, opened 2016-06-04
Add 4 New Common Scripting Macros

654wak654:
When merged this PR will add 4 common scripting macros: `MAP()`, `FILTER()`, `UNIQUE()`, and `INTERSECTION()`. The macros will assign the return of the `apply`, `select CODE`, `ARR arrayIntersect ARR`, and `ARR0 arrayIntersect ARR1` commands respectively to the original array.

### DIFF
--- a/addons/main/script_macros_common.hpp
+++ b/addons/main/script_macros_common.hpp
@@ -535,7 +535,7 @@ Description:
     resulting array to the original
 
 Parameters:
-    ARR - Array to be modified
+    ARRAY - Array to be modified
     CODE - Code that'll be applied to each element of the array.
 
 Example:
@@ -556,8 +556,8 @@ Description:
     to the original
 
 Parameters:
-    ARR - Array to be filtered
-    CODE - Condition to pick elements.
+    ARRAY - Array to be filtered
+    CODE - Condition to pick elements
 
 Example:
     (begin example)
@@ -576,7 +576,7 @@ Description:
     Removes duplicate values in given array
 
 Parameters:
-    ARRAY0 - The array to be modified
+    ARRAY - The array to be modified
 
 Example:
     (begin example)
--- a/addons/main/script_macros_common.hpp
+++ b/addons/main/script_macros_common.hpp
@@ -530,6 +530,9 @@ Author:
 
 /* -------------------------------------------
 Macro: MAP()
+Description:
+    Applies given code to each element of the array, then assigns the
+    resulting array to the original
 
 Parameters:
     ARR - Array to be modified
@@ -538,7 +541,7 @@ Parameters:
 Example:
     (begin example)
         _array = [1, 2, 3, 4, 3, 8];
-        MAP(_array,_x+1);
+        MAP(_array,_x + 1);
         // _array is now [2, 3, 4, 5, 4, 9];
     (end)
 Author:
@@ -548,6 +551,9 @@ Author:
 
 /* -------------------------------------------
 Macro: FILTER()
+Description:
+    Filters an array based on given code, then assigns the resulting array 
+    to the original
 
 Parameters:
     ARR - Array to be filtered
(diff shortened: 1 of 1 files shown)


bux578:
I'd prefer keeping the brackets for the code part
`FILTER(_array,{_x % 2 == 0})`
vs.
`FILTER(_array,_x % 2 == 0)`

Imho it's a bit more clear what's going on here

commy2:
Imo, those curly brackets add nothing, might as well leave them out.

654wak654:
They are pretty unnecessary, if anything this makes it more clear. Plus this way it makes sure everyone uses only the `ARRAY select CODE` syntax.

Glowbal:
None of those macros do what they say: you are removing the original array. You just assigned a new array to the same variable.

commy2:
@Glowbal 
the logic behind these is to not have to write down redundant variable names

See this older one:
 #define ADD(var1,var2) var1 = (var1) + (var2)

Just compare:
`UNIQUE(_allThoseUnits);`
vs.
`_allThoseUnits = _allThoseUnits arrayIntersect _allThoseUnits;`

addons/main/script_macros_common.hpp
```diff
@@ -529,6 +529,89 @@ Macro: PUSH()
...
+    (begin example)
+        _someArray = [1, 2, 3, 4, 5, 5];
+        _anotherArray = [4, 5, 6, 7];
+        INTERSECTION(_someArray,_anotherArray);
+        // _someArray is now [4, 5]
+    (end)
+Author:
+    654wak654
+------------------------------------------- */
+#define INTERSECTION(ARG0,ARG1) ARG0 = ARG0 arrayIntersect ARG1
```
commy2: If you add parenthesis around `ARG1`, this would be possible without error:
`INTERSECTION(_respawnedUnits, BluePlayers + RedPlayers);`

(`ARG0` and `ARR` don't need them, because they are always variables)

Glowbal:
I'd prefer 

```sqf
_allThoseUnits = UNIQUE(_allThoseUnits);
```

That makes it a lot easier to understand what is happening compared to just an `UNIQUE` macro somewhere in your code. Plus it adds the flexibility of being able to be used in other context.

Example:
```sqf
{
   /// iterate over unique elements
} forEach UNIQUE(_myList);
```

vs

```sqf
UNIQUE(_myList);
{
   /// iterate over unique elements
} forEach _myList;
```

---
You are also getting rid of the original array - so if you want to use it later on in the code, you need to assign that to something else first. Or you would make a dummy to use later on:

```sqf
private _uniqueElements = _myList;
UNIQUE(_uniqueElements);
```
vs
```sqf
private _uniqueElements = UNIQUE(_myList);
```

---

I think these macros have undesired side effects and should just return the new array. Let the user of those macros determine whatever or not the results should be assigned to the original variable or to a new one.

commy2:
The point of these was to __not__ have to write down the variable name twice. `select`, `apply` and `arrayIntersect` copy the array unfortunately. Just like `+` and `-`.
There were macros for `ADD` and `SUB`/`REM` and these would be the new additons using new command from A3.

Glowbal:
I think it's a bad idea to use it like this. I'd either find a more descriptive name for them, or get rid of the side effects.

esteldunedain:
@Glowbal, I agree with you on principle, but the fact is that this macros are consistent with the existing CBA macros like ADD and SUB.

Glowbal:
I disagree. Those macros are supposed to modify the array by either adding a new element, or removing some element(s). It's clear from the naming what they do. I'd except those to have side effects.

The macros here indicate they give you some sub set of the array. I'd not expect these macros to have side effects.

esteldunedain:
I guess it depends on your POV. If you want to compute the union of two sets (ADD), you can store the result in an existing or new variable. CBA chose to do reuse the existing variable, but that's hardly indicated by the name of the macro.

654wak654:
`MAP()` - I see no problem with this one. The confusion that `apply` actually **applies** given code to the original array still happens, so this is a great little macro.

`FILTER()` - Again, this one is fine for the same reasons with `MAP()`, it does what it says. The name 'filter' (just like map) is existent in other languages and does the same thing as this macro.

`UNIQUE()` - There could be a 5th macro, to solve this (Maybe  `DISTINCT()`?). One whould assign the value to the array and one whould just return the value.

`INTERSECTION()` - Same argument as above. @commy2 saw your note, will add it after we settle the final form of these macros.

SilentSpike:
I'm not for or against the inclusion of these macros, however I personally don't see myself ever using them. Aren't these literally just renamed engine commands?

jonpas:
I am with @SilentSpike here, I'd never use them. It's just a tiny bit better than doing `SEL(array,index)` instead of `_array select _index`.

Killswitch00:
@654wak654 : add the parentheses in INTERSECTION() like commy2 suggested and this could be merged.

654wak654:
Closing because I lost the branch, will make new PR.

PR #365, opened 2016-06-05
Add 'name' value to all CfgPatches

commy2:
See: https://community.bistudio.com/wiki/CfgPatches

- also fixes certain objects placed in 3DEN causing the mission to require CBA due to use touching the class by adding XEH classes.
- also uses the new "name" instead of the configs classname in the help module (if present)

### DIFF
--- a/addons/diagnostic/xeh/config.cpp
+++ b/addons/diagnostic/xeh/config.cpp
@@ -2,6 +2,9 @@
 #ifndef EH_DEBUG_ENABLED
     class CfgPatches {
         class cba_diagnostics_xeh_disabled {
+            author = "$STR_CBA_Author";
+            name = "$STR_CBA_Optional_Component";
+            url = "$STR_CBA_URL";
             units[] = {};
             weapons[] = {};
             worlds[] = {};
@@ -14,6 +17,9 @@
     #include "script_component.hpp"
     class CfgPatches {
         class ADDON {
+            author = "$STR_CBA_Author";
+            name = "$STR_CBA_Optional_Component";
+            url = "$STR_CBA_URL";
             units[] = {};
             weapons[] = {};
             worlds[] = {};
--- a/optionals/cache_disable/config.cpp
+++ b/optionals/cache_disable/config.cpp
@@ -2,14 +2,15 @@
 
 class CfgPatches {
     class ADDON {
+        author = "$STR_CBA_Author";
+        name = ECSTRING(Optional,Component);
+        url = "$STR_CBA_URL";
         units[] = {};
         weapons[] = {};
         requiredVersion = REQUIRED_VERSION;
         requiredAddons[] = {"CBA_Extended_EventHandlers", "CBA_Main"};
         version = VERSION;
-        author = "$STR_CBA_Author";
         authors[] = {"Sickboy"};
-        authorUrl = "http://dev-heaven.net/projects/cca";
     };
 };
 
--- a/optionals/diagnostic_disable_xeh_logging/config.cpp
+++ b/optionals/diagnostic_disable_xeh_logging/config.cpp
@@ -11,6 +11,9 @@
 
 class CfgPatches {
     class Disable_XEH_Logging {
+        author = "$STR_CBA_Author";
+        name = "$STR_CBA_Optional_Component";
+        url = "$STR_CBA_URL";
         units[] = {};
         weapons[] = {};
         requiredVersion = 0.1;
--- a/optionals/diagnostic_enable_logging/config.cpp
+++ b/optionals/diagnostic_enable_logging/config.cpp
@@ -1,27 +1,23 @@
 #include "script_component.hpp"
 
-class CfgPatches
-{
-    class ADDON
-    {
+class CfgPatches {
+    class ADDON {
+        author = "$STR_CBA_Author";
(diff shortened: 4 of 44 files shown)


commy2:
This is used in that required-addons-for-this-mission-feature (CfgPatches rework) from 1.60


Unfortunately the game uses `configSourceMod` to detect the origin of a class instead of `units[]`, which means that every class where we fix for example `class EventHandlers` is detected as CBA-only class.
That is annoying and I'm searching for a work around, but I'm not sure I can fix that.

addons/diagnostic/xeh/config.cpp
```diff
@@ -2,6 +2,9 @@
 #ifndef EH_DEBUG_ENABLED
     class CfgPatches {
         class cba_diagnostics_xeh_disabled {
+            author = "$STR_CBA_Author";
+            name = "$STR_CBA_Optional_Component";
```
jonpas: `ECSTRING(Optional,Component)`

addons/diagnostic/xeh/config.cpp
```diff
@@ -14,6 +17,9 @@
     #include "script_component.hpp"
     class CfgPatches {
         class ADDON {
+            author = "$STR_CBA_Author";
+            name = "$STR_CBA_Optional_Component";
```
jonpas: `ECSTRING(Optional,Component)`

optionals/jr_disable_long_scopes_on_short_mg_rail/config.cpp
```diff
@@ -1,17 +1,16 @@
 #include "script_component.hpp"
 
-class CfgPatches
-{
-    class ADDON
-    {
+class CfgPatches {
+    class ADDON {
+        author = "$STR_CBA_Author";
+        name = "$STR_CBA_Optional_Component";
```
jonpas: `ECSTRING(Optional,Component)`

optionals/diagnostic_enable_perf_loop/config.cpp
```diff
@@ -1,30 +1,20 @@
...
-
-//Class cba_diagnostic_enable_perf_loop : config.bin{
-class CfgPatches
-{
-    class cba_diagnostic_enable_perf_loop
-    {
+class CfgPatches {
+    class ADDON {
+        author = "$STR_CBA_Author";
+        name = "$STR_CBA_Optional_Component";
```
jonpas: `ECSTRING(Optional,Component)`

optionals/diagnostic_enable_logging/config.cpp
```diff
@@ -1,27 +1,23 @@
 #include "script_component.hpp"
 
-class CfgPatches
-{
-    class ADDON
-    {
+class CfgPatches {
+    class ADDON {
+        author = "$STR_CBA_Author";
+        name = "$STR_CBA_Optional_Component";
```
jonpas: `ECSTRING(Optional,Component)`

optionals/diagnostic_disable_xeh_logging/config.cpp
```diff
@@ -11,6 +11,9 @@
 
 class CfgPatches {
     class Disable_XEH_Logging {
+        author = "$STR_CBA_Author";
+        name = "$STR_CBA_Optional_Component";
```
jonpas: `ECSTRING(Optional,Component)`

optionals/cache_disable/config.cpp
```diff
@@ -2,14 +2,15 @@
 
 class CfgPatches {
     class ADDON {
+        author = "$STR_CBA_Author";
+        name = "$STR_CBA_Optional_Component";
```
jonpas: `ECSTRING(Optional,Component)`

jonpas:
Make a ticket on A3 feedback tracker about it?

commy2:
I actually found a funny way to solve this. You can add the `addonRootClass` which is supposed to be used in secondary configs in a component.

By adding:
`addonRootClass = "A3_Characters_F"`

That particular unit no longer requires "CBA - XEH". The problem is, that you can obviously only add one `addonRootClass` per component.

In 3DEN the unit is still shown with the CBA icon, but that has been the case before too.

addons/diagnostic/xeh/config.cpp
commy2: Wouldn't that imply that "Optional" is a component?

jonpas: It would expand to the above. Not sure about semantics, if you are worried people might be looking for "optional" component then it might be best to leave it like that.

commy2:
NVM. Actually `addonRootClass` pushes every redefined addon to the linked component. I'm going with `A3_Characters_F` since logic implies that at least one soldier has to exist in a mission anyway, right?

addons/diagnostic/xeh/config.cpp
commy2: I added it everywhere I can, but there are some PBOs that don't or can't include the script_mod

optionals/diagnostic_disable_xeh_logging/config.cpp
Killswitch00: This addon is very simple and does not `#include...` the usual bits and pieces, hence the use of a literal stringtable name.

PR #367, opened 2016-06-06
Add quotes around setFormation/Behaviour params. Fixes #366

### LINKED ISSUE #366: Script error from fnc_searchNearby
image
https://github.com/CBATeam/CBA_A3/blob/master/addons/ai/fnc_searchNearby.sqf#L37

Looks like they're not strings?
This is from an editor placed waypoint with on act: `0 = [group this] call CBA_fnc_searchNearby`, for some reason it doesn't show up in the .rpt

There's also no locality check on this function, most other group based functions seems to have that.

Killswitch00:


Killswitch00:
Verified working with group in staggered column format (`STAG COLUMN`). Now they actually do the search whereas before it would fail silently and the group did nothing.

Issue #374, opened 2016-06-10
Upgrade CBA_fnc_addEventHandler to support arguments and EH metadata


Tenga:
Currently, it is impossible to pass arguments into the EH code with `CBA_fnc_addEventHandler`.

It would be extremely nice if it was upgraded to support `_thisArgs`, `_thisType`, `_thisId` and` _thisFnc` like the `CBA_fnc_addBISEventHandler` does.

PR #375, opened 2016-06-11
add '_thisArgs', '_thisId', '_thisType' and '_thisFnc'  and  to CBA events

commy2:
Closes: #374

Adds the same meta data `CBA_fnc_addBISEventHandler` has to the CBA events.

- adds optional parameter to `CBA_fnc_addEventHandler` to assign arguments to the event.

`_thisArgs` - Arguments passed when adding the event handler.
`_thisId` - Id that can be used to remove the event via `CBA_fnc_removeEventHandler`
`_thisType` - the name of the event ("testEvent" in `["testEvent", ...] call CBA_fnc_addEventHandler`)
`_thisFnc` - function that is executed when the event happens

Note: If the arguments passed to `CBA_fnc_addEventHandler` are an array, they are stored as array reference and not as copy. That means you can modify these arguments even after the event was added.

~~Needs some more testing, will do later ...~~

### DIFF
--- a/addons/events/script_component.hpp
+++ b/addons/events/script_component.hpp
@@ -33,9 +33,7 @@
 
 #define CALL_EVENT(args,event) {\
     if !(isNil "_x") then {\
-        private _thisType = event;\
-        _x params ["_thisFnc", "_thisArgs", "_thisId"];\
-        args call _thisFnc;\
+        args call _x;\
     };\
 } forEach +(GVAR(eventNamespace) getVariable event) // copy array so events can be removed while iterating safely
 
--- a/addons/events/script_component.hpp
+++ b/addons/events/script_component.hpp
@@ -33,7 +33,9 @@
 
 #define CALL_EVENT(params,event) {\
     if !(isNil "_x") then {\
-        params call _x;\
+        private _thisType = event;\
+        _x params ["_thisFnc", "_thisArgs", "_thisId"];\
+        params call _thisFnc;\
     };\
 } forEach (GVAR(eventNamespace) getVariable event)
 
--- a/addons/events/script_component.hpp
+++ b/addons/events/script_component.hpp
@@ -37,7 +37,7 @@
         _x params ["_thisFnc", "_thisArgs", "_thisId"];\
         args call _thisFnc;\
     };\
-} forEach (GVAR(eventNamespace) getVariable event)
+} forEach +(GVAR(eventNamespace) getVariable event) // copy array so events can be removed while iterating safely
 
 #define GETOBJ(obj) (if (obj isEqualType grpNull) then {leader obj} else {obj})
 
--- a/addons/events/script_component.hpp
+++ b/addons/events/script_component.hpp
@@ -31,11 +31,11 @@
 
 #define SEND_TEVENT_TO_SERVER(params,name,targets) TEVENT_PVAR = [name, params, targets]; publicVariableServer TEVENT_PVAR_STR
 
-#define CALL_EVENT(params,event) {\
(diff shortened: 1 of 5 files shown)


addons/events/script_component.hpp
```diff
@@ -33,7 +33,9 @@
 
 #define CALL_EVENT(params,event) {\
```
Killswitch00: Should the macro argument "params" be renamed so as to not collide with the reserved word (command) `params`?

commy2: Yes. That collided with the params command I added ...

commy2:
One time event handler printing "hello world"

```
["test1", {
    systemChat str _thisArgs;
    [_thisType, _thisId] call CBA_fnc_removeEventHandler
}, "hello world"] call CBA_fnc_addEventHandler;

...

"test1" call CBA_fnc_localEvent
```

Problem is that you are modifing an array GVAR(eventsNamespace) while iterating through it via `forEach` which is undefined ... Searching for a work-around.

commy2:
done

PabstMirror:
I'm really not sure if any of this is necessary, my main concern is that there will be a performance cost to adding these args.

For 10x localEvent calls for something with 3 added ehs:

Old: 0.218436 ms 0.216403 ms  0.217628 ms
New: 0.326691 ms 0.327654 ms 0.324465 ms

commy2:
Good point. It's performance vs. functionality. I have no idea with what we should go here.
+0.10 ms isn't that bad though honestly

PabstMirror:
ACE calls multiple EHs for each bullet fired, so 50% increase isn't good

I think we can do most of what we need by using a function wrapper. Check out my demo code here (not tested, more of an idea): https://github.com/CBATeam/CBA_A3/compare/addEventHandlerWrapper?expand=1

commy2:
Is it a 50% increase even when measuring with the actual event function you are calling?

PabstMirror:
```
Prep:
["test", {x1=_this;}] call CBA_fnc_addEventHandler;
["test", {x2=_this;}] call CBA_fnc_addEventHandler;
["test", {x3=_this;}] call CBA_fnc_addEventHandler;

Test:
["test", []] call CBA_fnc_localEvent;
["test", []] call CBA_fnc_localEvent;
["test", []] call CBA_fnc_localEvent;
["test", []] call CBA_fnc_localEvent;
["test", []] call CBA_fnc_localEvent;
["test", []] call CBA_fnc_localEvent;
["test", []] call CBA_fnc_localEvent;
["test", []] call CBA_fnc_localEvent;
["test", []] call CBA_fnc_localEvent;
["test", []] call CBA_fnc_localEvent;
```

commy2:
My point was that `{x1=_this;}` is a extremly simple function. The bigger the event function is, the smaller the share of the overhead of `CBA_fnc_localEvent` becomes.

commy2:
@PabstMirror I read the link you posted.
I like that idea. It would mean having no additional overhead by default, but an alternative addEventHandler function that would apply all these variables.
Any naming suggestions for `FUNC(addBigEH)`?

I'd still have to copy the array though, but that is only once per event call and not once per event per call.

Killswitch00:
`addEventHandlerArgs`, `addArgsEventHandler`

An alternative/wrapper function sounds like the best way forward, so as to not compromise performance of the original function. Now, to think of a suitable name...

commy2:
I'm going with `addEventHandlerArgs` instead of the usual `addXEventHandler`, because it isn't actually a new thing on it's own. It also uses the same `removeEventHandler` function.

PR #377, opened 2016-06-13
use ASL position instead of AGL to get direction vector

commy2:
Currently it's using the AGL position, which means that there will be an offset in the z-axis by exactly the altitude difference between those two points. It never really comes into play since these points are (typically) __very__ close to each other, but this is the correct way.


esteldunedain:
Exactly

commy2:
Pardon?

esteldunedain:
I meant that your comment is completely accurate.

commy2:
Wait for my next issue report in a few minutes, haha

PR #380, opened 2016-06-14
improve 'CBA_fnc_sortNestedArray'

commy2:
- makes `CBA_fnc_sortNestedArray` use SQFs native `sort` command
- adds optional parameter to reverse sorting order

`BIS_fnc_sortBy` is using the same method now

### DIFF
--- a/addons/arrays/fnc_sortNestedArray.sqf
+++ b/addons/arrays/fnc_sortNestedArray.sqf
@@ -10,7 +10,7 @@ Description:
 Parameters:
     _array - Nested array to be sorted <ARRAY>
     _index - Sub array item index to be sorted on <NUMBER>
-    _order - true: ascending, false: descending (optional: true) <BOOLEAN>
+    _order - true: ascending, false: descending (optional, default: true) <BOOLEAN>
 
 Example:
     (begin example)
--- a/addons/arrays/fnc_sortNestedArray.sqf
+++ b/addons/arrays/fnc_sortNestedArray.sqf
@@ -5,75 +5,37 @@ Description:
     Used to sort a nested array from lowest to highest using quick sort.
 
     Sorting is based on the specified column, which must have numerical data.
+    Original array is modified.
 
 Parameters:
-    _array: array - Nested array to be sorted
-    _index: integer - sub array item index to be sorted on
+    _array - Nested array to be sorted <ARRAY>
+    _index - Sub array item index to be sorted on <NUMBER>
+    _order - true: ascending, false: descending (optional: true) <BOOLEAN>
 
 Example:
     (begin example)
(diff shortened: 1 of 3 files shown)


addons/arrays/fnc_sortNestedArray.sqf
```diff
@@ -5,75 +5,37 @@ Description:
...
+    [_x select _index, _x]
 };
 
-// and start it off
-[_this select 0, _this select 1, 0, 0 max ((count (_this select 0))-1)] call _sort;
+_helperArray sort _order;
+
+{
+    _array set [_forEachIndex, _helperArray select _forEachIndex select 1];
+} forEach _array;
```
PabstMirror: Could do:
```
    {
        _array set [_forEachIndex, _x select 1];
    } forEach _helperArray;
```

Glowbal:
```sqf
_array = _helperArray apply {_x select 1};
```

bux578: `apply` is 1.56, so it won't work with the linux builds that are on 1.54 iirc.

Glowbal: It would have been so beautiful.

commy2: `CBA_fnc_sortNestedArray` originally modified the array by reference, which means `select` and `apply` cannot be used. Even if you reassign the new array to the previous variable, it wouldnt behave the same.

```
blah = [0];
player setVariable ["blah", _blah];
blah set [0, 1];
player getVariable "blah";
-> 1
```
vs.
```
blah = [0];
player setVariable ["blah", _blah];
blah = blah apply {1};
player getVariable "blah";
-> 0
```

bux578: Creating a new array is kind of industry standard though.

commy2: This function defies the standard, but we should go with bwc always.

addons/arrays/fnc_sortNestedArray.sqf
```diff
@@ -5,75 +5,43 @@ Description:
...
 
     Sorting is based on the specified column, which must have numerical data.
+    Original array is modified.
 
 Parameters:
-    _array: array - Nested array to be sorted
-    _index: integer - sub array item index to be sorted on
+    _array - Nested array to be sorted <ARRAY>
+    _index - Sub array item index to be sorted on <NUMBER>
+    _order - true: ascending, false: descending (optional: true) <BOOLEAN>
```
Killswitch00: Should the "optional: true" comment mean that 1) it is optional and 2) the default is "true"? In that case,
`(optional, default: true)` would be easier to understand.

PR #381, opened 2016-06-16
add parameter to 'createNamespace' to create a public namespace

commy2:
- title
- on a global (public) pseudo namespace one can use `setVariable pulic`
- I made a "Static" dummy object, because the game logics (which have to be created with `createUnit`) have AI overhead ...

- I use `isEqualTo true` instead of forcing a \<BOOLEAN\> with `params` because that would break bwc (on screen error) in case someone uses `call CBA_fnc_...`. That syntax carries over the `_this` variable from the parent scope ...

### DIFF
--- a/addons/common/CfgVehicles.hpp
+++ b/addons/common/CfgVehicles.hpp
@@ -1,0 +1,8 @@
+
+class CfgVehicles {
+    class Static;
+    class CBA_NamespaceDummy: Static {
+        scope = 1;
+        displayName = "";
+    };
+};
--- a/addons/common/config.cpp
+++ b/addons/common/config.cpp
@@ -16,6 +16,8 @@ class CfgPatches {
 
 #include "CfgEventHandlers.hpp"
 #include "CfgFunctions.hpp"
+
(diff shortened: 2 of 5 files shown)


PabstMirror:
Tagging @NouberNou - for his experience with hashes.

I thought there might have been a possible issue with the game dropping data on some objects like Houses/`Static`.

commy2:
You shouldn't put variables on streamed objects (map objects). This is a script created "building" and the variables should be persistent even if the building is far away.

commy2:
>Now word of warning before you bash your head against a wall as to “why it ain’t working with this object?!?”. Houses for example like other map objects (and I mean map objects that have been added during map creation, not mission objects) are streamed. Think about it, the maps in ArmA are huge so even if you can load the whole map you will be using too much computer power. This is unnecessary, so map objects are streamed to you as you go. This means that if you attach a variable to an object on your map and try to make it global, on other computers the same object might not exist yet. And when the object appears later on other computers, it might not have variable you attached to it, so remember this.

http://killzonekid.com/arma-scripting-tutorials-variables-part-3/

PR #383, opened 2016-06-17
Add QQ macros for escaped config implementation

jameslkingsley:
* Add the following macros:
    * QQGVAR
    * QQEGVAR
    * QQGVARMAIN
    * QQFUNC
    * QQFUNCMAIN
    * QQFUNC_INNER
    * QQEFUNC

This is useful for when you need to escape these macros in a config.

**Before:**
```c++
expression = QUOTE(\
    GVAR(camSlowDownInBuildings) = _this;\
    [ARR_2(QUOTE(QGVAR(camSlowDownInBuildings)), _this)] call CFUNC(saveSetting);\
);
```

**After:**
```c++
expression = QUOTE(\
    GVAR(camSlowDownInBuildings) = _this;\
    [ARR_2(QQGVAR(camSlowDownInBuildings), _this)] call CFUNC(saveSetting);\
);
```

### DIFF
--- a/addons/main/script_macros_common.hpp
+++ b/addons/main/script_macros_common.hpp
@@ -677,6 +677,8 @@ Author:
 #define EGVAR(var1,var2) TRIPLES(PREFIX,var1,var2)
 #define QGVAR(var1) QUOTE(GVAR(var1))
 #define QEGVAR(var1,var2) QUOTE(EGVAR(var1,var2))
+#define QQGVAR(var1) QUOTE(QGVAR(var1))
+#define QQEGVAR(var1,var2) QUOTE(QEGVAR(var1,var2))
 
 /* -------------------------------------------
 Macro: GVARMAIN()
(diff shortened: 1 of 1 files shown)


commy2:
While I do support this, you could've also used:

```
expression = QUOTE(\
    GVAR(camSlowDownInBuildings) = _this;\
    [ARR_2('GVAR(camSlowDownInBuildings)',_this)] call CFUNC(saveSetting);\
);
```

esteldunedain:
@commy2, do macros work inside single quotes??

commy2:
Yes. See: https://github.com/CBATeam/CBA_A3/pull/317/files#diff-dc5764ac6d31dc8aa39a8b74be90f46fR22

commy2:
Or here in stable: https://github.com/CBATeam/CBA_A3/blob/master/addons/main/script_macros_common.hpp#L713-L719

esteldunedain:
I'm confused. Why don't we always just do this then?

```
expression = 'GVAR(camSlowDownInBuildings) = _this;\
    [ARR_2(QGVAR(camSlowDownInBuildings),_this)] call CFUNC(saveSetting);\
);'
```

jameslkingsley:
I must have got it mixed up. I though that only works in SQF files.

commy2:
>I'm confused. Why don't we always just do this then?
Won't work. I think this is because `'` is the escaping quote mark and not `"`.

PR #386, opened 2016-06-18
fix wrong min/max values in functions header

commy2:
arcus sinus boundaries are plus a half of pi and minus a half of pi, a.k.a. plus 90 degree and minus 90 degree ...


jonpas:
Lovely trigonometry.

PR #387, opened 2016-06-19
JR for APEX

donkeypunchepoch:
A3 Apex gives us Retextured Weapons and Attachments from A3 vanilla. These were added to the CfgWeapons and JRClasses, i left Apex as a not required addon :)
-sTanG

### DIFF
--- a/addons/jr/jr_classes.hpp
+++ b/addons/jr/jr_classes.hpp
@@ -46,6 +46,7 @@ class asdg_UnderSlot: asdg_SlotInfo { // Base under barrel slot
         bipod_02_F_hex = 1;
         bipod_03_F_blk = 1;
         bipod_03_F_oli = 1;
+		bipod_01_F_khk = 1;//dpapexaddon
     };
 };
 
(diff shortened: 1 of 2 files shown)


jonpas:
Please use 4 spaces for indentation and opening braces on the same line as class.

addons/jr/jr_classes.hpp
```diff
@@ -76,6 +77,20 @@ class asdg_OpticRail1913: asdg_OpticRail { // the "medium" rail, long enough to
         optic_KHS_hex = 1;
         optic_KHS_old = 1;
         optic_KHS_tan = 1;
+		//dpapexaddon
```
jonpas: Just `// Apex` will do.

donkeypunchepoch: the 4 spaces for indent does not fit the files current structure as I see it.... and if you guys want to use it please do and make what ever changes you need. If spaces and indents are what you are worried about when a free lance community guys helps update then im helping the wrong people.

jonpas: > the 4 spaces for indent does not fit the files current structure as I see it

Yes it is. You can check the diff here to see the obvious difference.

> If spaces and indents are what you are worried about when a free lance community guys helps update then im helping the wrong people.

So you expect anyone can contribute with their own coding style? That would be an incredible mess. Every project needs to have some standard or you are asking for ugly code and a mess.

commy2:
These configs don't look right. Is it changing class inheritance?
We shouldn't do these configs just yet anyway. My experience says that BI will change details around many times until the release of APEX.

commy2:
close in favour of #388

PR #390, opened 2016-06-22
Update fnc_headDir

654wak654:
When merged this PR will:
1. Update the function using new commands
2. Clean old code

### DIFF
--- a/addons/common/fnc_headDir.sqf
+++ b/addons/common/fnc_headDir.sqf
@@ -54,49 +54,35 @@ Author:
 #include "script_component.hpp"
 SCRIPT(headDir);
 
-private["_azimuth", "_angle", "_dif", "_infov", "_object", "_threed", "_do"];
-
 params ["_unit"];
-_object = param [1,_unit];
-_threed = false;
+private _object = param [1, _unit];
+private _isExternalCam = false;
 
-_do = (typeName _object != typeName _unit);
-if (!_do) then {
-    _do = (_object != _unit);
-};
-
-if (_do) then {
-    _angle = [(_unit call CBA_fnc_getPos),(_object call CBA_fnc_getPos)] call BIS_fnc_dirTo;
+private _azimuth = 0;
+if (_unit != player) then {
+    _azimuth = getDir _unit;
 } else {
-    _angle = 0;
-};
-if (_unit != player) then
-{
-    _azimuth = getdir _unit;
-} else {
-    private ["_position","_viewPos","_vector","_magnitude"];
-    _position = positionCameraToWorld [0, 0, 0];
-    _threed = if ((_position distance _unit)>2) then {true} else {false};
-    _viewPos = positionCameraToWorld [0, 0, 99999999];
-    _vector = [
-        (_viewPos select 0) - (_position select 0),
-        (_viewPos select 1) - (_position select 1),
-        (_viewPos select 2) - (_position select 2)
-    ];
-    _magnitude = [0, 0, 0] distance _vector;
-    _vector = [
-        (_vector select 0) / _magnitude,
-        (_vector select 1) / _magnitude,
-        (_vector select 2) / _magnitude
-    ];
+    private _camPos = positionCameraToWorld [0, 0, 0];
+    _isExternalCam = (_camPos distance _unit) > 2;
+    private _viewPos = positionCameraToWorld [0, 0, 99999999];
+    private _vector = _viewPos vectorDiff _camPos;
+    _vector = _vector vectorMultiply (1 / vectorMagnitude _vector);
     _azimuth = (_vector select 0) atan2 (_vector select 1);
     _azimuth = [_azimuth] call CBA_fnc_simplifyAngle;
 };
 
-_dif = _angle - _azimuth;
+private _diff = -_azimuth;
 
-if (_dif < 0) then { _dif = 360 + _dif;};
-if (_dif > 180) then { _dif = _dif - 360;};
-_infov = if (abs(_dif) < 43) then {true} else {false};
+if (_object != _unit) then {
+    ADD(_diff,_unit getDir _object);
+};
+if (_diff < 0) then {
+    ADD(_diff,360);
+};
+if (_diff > 180) then {
+    SUB(_diff,360);
+};
 
-[_azimuth,_dif,_infov,_threed]
+private _inFOV = abs _diff < 43;
+
+[_azimuth, _diff, _inFOV, _isExternalCam]
--- a/addons/common/fnc_headDir.sqf
+++ b/addons/common/fnc_headDir.sqf
@@ -54,17 +54,19 @@ Author:
 #include "script_component.hpp"
 SCRIPT(headDir);
 
-params ["_unit"];
-private _object = param [1, _unit];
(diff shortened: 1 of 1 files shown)


addons/common/fnc_headDir.sqf
```diff
@@ -54,49 +54,35 @@ Author:
...
+    _isExternalCam = (_camPos distance _unit) > 2;
+    private _viewPos = positionCameraToWorld [0, 0, 99999999];
+    private _vector = _viewPos vectorDiff _camPos;
+    _vector = _vector vectorMultiply (1 / vectorMagnitude _vector);
     _azimuth = (_vector select 0) atan2 (_vector select 1);
     _azimuth = [_azimuth] call CBA_fnc_simplifyAngle;
 };
 
-_dif = _angle - _azimuth;
+private _diff = -_azimuth;
```
commy2: This looks much cleaner, but what happened to `_angle` ?

addons/common/fnc_headDir.sqf
```diff
@@ -54,49 +54,35 @@ Author:
...
-    _do = (_object != _unit);
-};
+private _object = param [1, _unit];
+private _isExternalCam = false;
 
-if (_do) then {
-    _angle = [(_unit call CBA_fnc_getPos),(_object call CBA_fnc_getPos)] call BIS_fnc_dirTo;
-} else {
-    _angle = 0;
-};
```
commy2: cont.
I guess `_angle` is 0 unless `_do` is true. Question is: what is "`_do`". I don't understand the old code.

addons/common/fnc_headDir.sqf
```diff
@@ -54,49 +54,35 @@ Author:
...
-if (_unit != player) then
-{
-    _azimuth = getdir _unit;
+private _azimuth = 0;
+if (_unit != player) then {
```
commy2: maybe replace this `player` with `call CBA_fnc_currentUnit`. That way the function wouldn't bug out when remote controlling as zeus.

addons/common/fnc_headDir.sqf
```diff
@@ -54,49 +54,35 @@ Author:
...
-    ];
-    _magnitude = [0, 0, 0] distance _vector;
-    _vector = [
-        (_vector select 0) / _magnitude,
-        (_vector select 1) / _magnitude,
-        (_vector select 2) / _magnitude
-    ];
+    private _camPos = positionCameraToWorld [0, 0, 0];
+    _isExternalCam = (_camPos distance _unit) > 2;
+    private _viewPos = positionCameraToWorld [0, 0, 99999999];
```
commy2: Converting these to ASL would remove the terrain offset you get with the current function. Simply add `AGLToASL` before every `positionCameraToWorld`
...

addons/common/fnc_headDir.sqf
```diff
@@ -54,49 +54,35 @@ Author:
...
-        (_viewPos select 2) - (_position select 2)
-    ];
-    _magnitude = [0, 0, 0] distance _vector;
-    _vector = [
-        (_vector select 0) / _magnitude,
-        (_vector select 1) / _magnitude,
-        (_vector select 2) / _magnitude
-    ];
+    private _camPos = positionCameraToWorld [0, 0, 0];
+    _isExternalCam = (_camPos distance _unit) > 2;
```
commy2: which also means the vector functions should be used here. apparently `distance` works with AGL and objects are converted to their AGL positions. That is NOT true for the vector commands below though.
Imo we should always do vector math with the fixed ASL coordinate system

addons/common/fnc_headDir.sqf
```diff
@@ -54,49 +54,35 @@ Author:
 #include "script_component.hpp"
 SCRIPT(headDir);
 
-private["_azimuth", "_angle", "_dif", "_infov", "_object", "_threed", "_do"];
-
 params ["_unit"];
```
commy2: while we are at it, we should extend this `params` to do type validation.

addons/common/fnc_headDir.sqf
654wak654: I moved the relative angle code to lines 76-78:
https://github.com/654wak654/CBA_A3/blob/patch-1/addons/common/fnc_headDir.sqf#L76-L78

commy2: roger

addons/common/fnc_headDir.sqf
```diff
@@ -54,49 +54,37 @@ Author:
...
     _azimuth = (_vector select 0) atan2 (_vector select 1);
     _azimuth = [_azimuth] call CBA_fnc_simplifyAngle;
 };
 
-_dif = _angle - _azimuth;
+private _diff = -_azimuth;
+
+if (!isNull _offsetObject) then {
+    ADD(_diff,_unit getDir ([_offsetObject] call CBA_fnc_getPos));
+};
```
commy2: Won't this error if `_offsetobject` is a marker (STRING)?

addons/common/fnc_headDir.sqf
```diff
@@ -54,49 +54,37 @@ Author:
...
-        (_vector select 1) / _magnitude,
-        (_vector select 2) / _magnitude
-    ];
+    private _camPos = AGLToASL positionCameraToWorld [0, 0, 0];
+    _isExternalCam = (_camPos vectorDistance getPosASL _unit) > 2;
+    private _viewPos = AGLToASL positionCameraToWorld [0, 0, 99999999];
+    private _vector = _viewPos vectorDiff _camPos;
+    _vector = _vector vectorMultiply (1 / vectorMagnitude _vector);
     _azimuth = (_vector select 0) atan2 (_vector select 1);
     _azimuth = [_azimuth] call CBA_fnc_simplifyAngle;
```
commy2: A bit strange that we use this here, but then have L81-86

commy2: We probably don't need this. The angle is "simplified" with ^

654wak654: 81-86 are for after the offset is applied. Could move the simplify after the offset and remove 81-86.

addons/common/fnc_headDir.sqf
654wak654: This could work:
```sqf
    ["_offsetObject", ""]

    later on:
if !(_offsetObject isEqualTo "") then {
```

addons/common/fnc_headDir.sqf
```diff
@@ -75,8 +75,8 @@ if (_unit != call CBA_fnc_currentUnit) then {
 
 private _diff = -_azimuth;
 
-if (!isNull _offsetObject) then {
-    ADD(_diff,_unit getDir ([_offsetObject] call CBA_fnc_getPos));
+if !(_offset isEqualTo "") then {
```
commy2: but now it no longer works for markers at all, or am I wrong?

654wak654: If you supply any other string than an empty string, it will return true:
`!("base_marker" isEqualTo "")` will return true.

commy2: Ah. I read it as `isEqualType`

PR #399, opened 2016-06-23
add XEH to new classes

commy2:
probably an ongoing project


commy2:
this works now without APEX. Someone with APEX please confirm that it works there too.

commy2:
This fixes:
```
[XEH]: Plane_Civil_01_base_F does not support Extended Event Handlers! Addon: expansion
[XEH]: B_CTRG_Miller_F does not support Extended Event Handlers! Addon: expansion
```

and probably also those power lines when owning Tanoa

jonpas:
Confirmed, this fixes all XEH support.

jonpas:
Oh I knew I forgot to mention something, I tested with Apex.

PR #401, opened 2016-06-28
Versioning - Handle non-numbers in arrays

PabstMirror:
Using strings inside the version required array causes a error
`ACE[] = {"ace_common", {"3.6.0"}, "(true)"};`
will cause:
`Error >: Type String, expected Number,Not a Number`

Which seems to cascade and break all of postInit.

Simple fix to check that the arrays are all numbers, if not it will fail the check.

### DIFF
--- a/addons/versioning/XEH_preInit.sqf
+++ b/addons/versioning/XEH_preInit.sqf
@@ -46,6 +46,11 @@ PREP(version_check);
 FUNC(version_compare) = {
     private ["_failed", "_c"];
     params ["_value","_localValue"];
+
+    //Handle non-number arrays - eg. {"mod", {"1.0"}, "(true)"}
+    if ((!(_value isEqualTo [])) && {!(_value isEqualTypeAll 0)}) exitWith {true};
+    if ((!(_localValue isEqualTo [])) && {!(_localValue isEqualTypeAll 0)}) exitWith {true};
+
     _failed = false;
     _c = count _localValue;
 


addons/versioning/XEH_preInit.sqf
```diff
@@ -46,6 +46,11 @@ PREP(version_check);
 FUNC(version_compare) = {
     private ["_failed", "_c"];
     params ["_value","_localValue"];
+
+    //Handle non-number arrays - eg. {"mod", {"1.0"}, "(true)"}
+    if ((!(_value isEqualTo [])) && {!(_value isEqualTypeAll 0)}) exitWith {true};
+    if ((!(_localValue isEqualTo [])) && {!(_localValue isEqualTypeAll 0)}) exitWith {true};
```
Glowbal: We should probably log an error here.

addons/versioning/XEH_preInit.sqf
```diff
@@ -46,6 +46,11 @@ PREP(version_check);
 FUNC(version_compare) = {
     private ["_failed", "_c"];
     params ["_value","_localValue"];
+
+    //Handle non-number arrays - eg. {"mod", {"1.0"}, "(true)"}
+    if ((!(_value isEqualTo [])) && {!(_value isEqualTypeAll 0)}) exitWith {true};
```
commy2: inverted logic?

`_value = "1.0.0";
```
(!(_value isEqualTo [])) && {!(_value isEqualTypeAll 0)}
(!("1.0.0" isEqualTo [])) && {!(_value isEqualTypeAll 0)}
(!(false)) && {!(_value isEqualTypeAll 0)}
(true) && {!(_value isEqualTypeAll 0)}
!(_value isEqualTypeAll 0)
!("1.0.0" isEqualTypeAll 0)
-> error:
isEqualTypeAll; type: <STRING>, expected: <ARRAY>
```


`_value = [1,0,0];
```
(!(_value isEqualTo [])) && {!(_value isEqualTypeAll 0)}
(!([1,0,0] isEqualTo [])) && {!(_value isEqualTypeAll 0)}
(!(false)) && {!(_value isEqualTypeAll 0)}
(true) && {!(_value isEqualTypeAll 0)}
!(_value isEqualTypeAll 0)
!([1,0,0] isEqualTypeAll 0)
-> exits prematurely with "true"
```

commy2:
My suggestion:
`params ["_value","_localValue"];`
->
`params [["_value", [], [[]]], ["_localValue", [], [[]]]];`

```
if !(_value isEqualTypeAll 0) exitWith {
    //log error
    false
};

if !(_localValue isEqualTypeAll 0) exitWith {
    //log error
    false
};
```

addons/versioning/XEH_preInit.sqf
commy2: And report an error. true -> false

PabstMirror: It will return true (true = failed) meaning it will report the dependency as being bad, which should be logged well enough.

Glowbal: Except it's not the dependency that is bad, it's the configuration that is incorrect.

PR #402, opened 2016-06-30
Cleanup Version UI for 1.62

PabstMirror:
Fixes:
```
Warning: no type entry inside class RscDisplayMain/controls/VersionNumber 
Warning: no type entry inside class RscDisplayMain/controls/CBA_Credits_Ver 
VAlign attribute invalid value "center"
```

In 1.62 there is no class `VersionNumber`, this redefines it using RscText so it should also be backward compatible.

### DIFF
--- a/addons/help/CfgRscStd.hpp
+++ b/addons/help/CfgRscStd.hpp
@@ -1,5 +1,7 @@
 
 class RscButton;
+class RscText;
+
 class CBA_Credits_Ver_Btn: RscButton {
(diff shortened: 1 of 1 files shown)


jonpas:
This should go into `apex` branch probably.

Killswitch00:
>  so it should also be backward compatible
Is it, and if so, how backwards compatible? (1.60? 1.58? 1.54?)

PR #405, opened 2016-07-02
Make fnc_headDir compatible with the 1.54 ports

Killswitch00:
This makes fnc_headDir load without errors when using the 1.54 MacOS/Linux
ports.

### DIFF
--- a/addons/common/fnc_headDir.sqf
+++ b/addons/common/fnc_headDir.sqf
@@ -76,7 +76,11 @@ if (_unit != call CBA_fnc_currentUnit) then {
 private _diff = -_azimuth;
 
 if !(_offset isEqualTo "") then {
+#ifndef LINUX_BUILD
     ADD(_diff,_unit getDir ([_offset] call CBA_fnc_getPos));
+#else
+    ADD(_diff,[_unit, [_offset] call CBA_fnc_getPos] call BIS_fnc_dirTo);
+#endif
 };
 if (_diff < 0) then {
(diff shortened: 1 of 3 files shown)


addons/common/fnc_headDir.sqf
```diff
@@ -76,7 +76,11 @@ if (_unit != call CBA_fnc_currentUnit) then {
 private _diff = -_azimuth;
 
 if !(_offset isEqualTo "") then {
+#ifndef LINUX_BUILD
     ADD(_diff,_unit getDir ([_offset] call CBA_fnc_getPos));
+#else
+    ADD(_diff,[_unit, [_offset] call CBA_fnc_getPos] call BIS_fnc_dirTo);
```
Killswitch00: Like `ARR_2(_unit,[_offset] call CBA_fnc_getPos)` ?

Killswitch00: Hmm... ARR_x, despite what the name suggests doesn't actually create an _array_ for you.

jokoho48: `ADD(_diff,[ARR_2(_unit, [_offset] call CBA_fnc_getPos)] call BIS_fnc_dirTo);`
this is how it must look like ARR_2 must be use because of some issues with the preprocessor.

Killswitch00: Yeah, I know. I was just thinking out loud... :-)

PR #409, opened 2016-07-06
CBA settings fixes

PabstMirror:
- Use getMissionConfigValue instead of broadcasting variable
- Fix GET_FORCED macro

3den expressions don't happen until after preInit
So `_settingsHash` would always be nil in `CBA_settings_fnc_init` if you add settings at preInit

### DIFF
--- a/addons/settings/Cfg3DEN.hpp
+++ b/addons/settings/Cfg3DEN.hpp
@@ -14,7 +14,7 @@ class Cfg3DEN {
                             displayName = "";
                             tooltip = "";
                             defaultValue = QUOTE(NULL_HASH);
-                            expression = QUOTE(missionNamespace setVariable [ARR_3(QUOTE(QGVAR(hash)),_value,true)]);
+                            expression = ""
                             wikiType = "[[Array]]";
                         };
                     };
--- a/addons/settings/fnc_init.sqf
+++ b/addons/settings/fnc_init.sqf
@@ -141,7 +141,7 @@ if (!isNil "_settingInfo") then {
 };
 
 // --- read previous setting values from mission
-_settingsHash = missionNamespace getVariable [QGVAR(hash), "Scenario" get3DENMissionAttribute QGVAR(hash)];
+_settingsHash = getMissionConfigValue QGVAR(hash);
 
 if (isNil "_settingsHash") then {
     _settingsHash = NULL_HASH;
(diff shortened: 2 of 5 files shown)


addons/settings/fnc_init.sqf
```diff
@@ -141,7 +141,7 @@ if (!isNil "_settingInfo") then {
 };
 
 // --- read previous setting values from mission
-_settingsHash = missionNamespace getVariable [QGVAR(hash), "Scenario" get3DENMissionAttribute QGVAR(hash)];
+_settingsHash = getMissionConfigValue QGVAR(hash);
 
 if (isNil "_settingsHash") then {
     _settingsHash = NULL_HASH;
```
commy2: We could use the alternative syntax of `getMissionConfigValue` to set `NULL_HASH` as default value.

There is one problem with this though. `getMissionConfigValue` only reports something if the mission.sqm exists. This means that all mission settings in 3den use default unless you saved the mission once.

While this makes no difference when previewing the mission (every mission settings use default there currently too), it means that when returning from preview the actual temporary mission settings and what is shown in the settings menu becomes desynched.

test:
`getMissionConfigValue "cba_settings_hash"`
vs.
`"Scenario" get3DENMissionAttribute "cba_settings_hash"`

This could be detected with `missionName` (will report `"tempMissionSP"` and maybe `"tempMissionMP"` in MP 3den).

We could store another copy of these settings in ui namespace if the mission was never saved. idk if that's worth it though.

dixon13:
`getMissionConfigValue` will not work for Linux users. Linux users are still on 1.54 I think. Refer to this for when an experimental port is released https://forums.bistudio.com/topic/188937-experimental-ports-release-announcements/

dixon13:
If this is added, need to change `REQUIRED_VERSION`

commy2:
The functions in the settings component are only compiled on the windows version. There is no script error pop ups or anything. The feature will simply not work on the Linux/MacOS build.

dixon13:
Sucks to be a Linux/MacOS user I guess lol

addons/settings/Cfg3DEN.hpp
```diff
@@ -14,7 +14,7 @@ class Cfg3DEN {
                             displayName = "";
                             tooltip = "";
                             defaultValue = QUOTE(NULL_HASH);
-                            expression = QUOTE(missionNamespace setVariable [ARR_3(QUOTE(QGVAR(hash)),_value,true)]);
+                            expression = ""
```
Killswitch00: Missing semicolon at the end

Killswitch00:
Broken merge

PR #411, opened 2016-07-08
new method for direct call

commy2:
The expression after `isNil` is executed in unscheduled environment. This way we also don't overwrite the `_x` variable.


esteldunedain:
Nice!

PR #415, opened 2016-07-09
Add 4 New Common Scripting Macros

654wak654:
When merged this PR will add 4 common scripting macros: `MAP()`, `FILTER()`, `UNIQUE()`, and `INTERSECTION()`. Their functionality is explained in the docs as well as this old PR -> #364


commy2:
>CBATeam:master

I hope this is the correct branch. The last time bux used that one, it ended up in limbo.

Killswitch00:
It is not the correct branch.

PR #418, opened 2016-07-10
fix: globalEventJIP would be executed twice

commy2:
- `globalEventJIP` will trigger the event twice if the global event arrives on the client before the JIP stack is read.

This PR changes the event to use a wrapper, so that the event can be discarded on each individual machine should the JIP stack be not be processed already.

@jonpas please test if this fixes your issue.


jonpas:
Seems to work @commy2 :+1:

PR #396, opened 2016-06-23
adjust keybinding 'RscDisplayConfigure' grid to 1.62

commy2:
closes  #379

PR #419, opened 2016-07-10
Safe Pre/Post 1.62 changes for keybind GUI

PabstMirror:
version safe changes for keybind gui
uses `productVersion`
should be safe for 1.54/1.60/1.62

Could replace PR #396


commy2:
works. nice :+1:

PR #420, opened 2016-07-11
adjust settings 'RscDisplayConfigure' grid to 1.62, target: master

commy2:
- Same as #397 
- is bwc with pre Apex versions (Linux/MacOS)


Killswitch00:
Am at work, but looks good.

PR #421, opened 2016-07-11
Fixed incorrect substring usage in find

nilsms:
When using substring in fnc_find with initialIndex, the substring should be from the index and to the end of the haystack.
The original code would return -1 for the example:
_result = ["frog-headed fish", "f", 5] call CBA_fnc_find;
And not 12 as shown in the example.

### DIFF
--- a/addons/strings/fnc_find.sqf
+++ b/addons/strings/fnc_find.sqf
@@ -53,7 +53,7 @@ if(_initialIndex < 1) then {
     if(_initialIndex > (count _haystack) ) exitWith {
         -1
     };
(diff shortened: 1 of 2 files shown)


Killswitch00:
For future use, add the example code from the function header to the unit test (`strings\test_strings.sqf`) where `CBA_fnc_find` is tested:
```
_pos = ["frog-headed fish", "f", 5] call CBA_fnc_find;
TEST_OP(_pos,==,12,_fn);
```

Killswitch00:
Thanks. Merging.

PR #424, opened 2016-07-12
Don't run fireteam sync PREH after apex

PabstMirror:
Tag @BaerMitUmlaut 

APEX changelog:
>Added: Multiplayer synchronization for assigned team functionality

We shouldn't need the PFEH running to sync.
It will still run on linux versions before 1.62.

ACE used the `CBA_teamColorChanged` event and I see no reason to remove it as other might have as well.

### DIFF
--- a/addons/common/fnc_onTeamColorChanged.sqf
+++ b/addons/common/fnc_onTeamColorChanged.sqf
@@ -19,6 +19,9 @@ Author:
 params ["_unit", "_team"];
 
 _unit assignTeam _team;
+
+if ((productVersion select 2) > 162) exitWith {};
+
 if (local (leader _unit)) then {
     _unit setVariable [QGVAR(synchedTeam), _team, true];
 };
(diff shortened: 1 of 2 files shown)


BaerMitUmlaut:
:+1:

commy2:
Is it confirmed that the vanilla thing works?

esteldunedain:
> Added: Multiplayer synchronization for assigned team functionality

Yay! And it only took them a decade or so

commy2:
I can confirm that this works since v1.62

See note: https://community.bistudio.com/wiki/assignTeam

addons/common/fnc_onTeamColorChanged.sqf
```diff
@@ -19,6 +19,9 @@ Author:
 params ["_unit", "_team"];
 
 _unit assignTeam _team;
+
+if ((productVersion select 2) > 162) exitWith {};
```
jonpas: Shouldn't this be `>=`?

jonpas: Fixed by #430

PR #430, opened 2016-07-12
Fix fireteam cleanup (don't setVar if apex or above)

PabstMirror:
I messed up inverting the logic, thanks to @jonpas for finding.


jonpas:
:+1:

PR #432, opened 2016-07-13
remove productVersion checks/ IS_LINUX macros

commy2:
removes bwc stuff that no longer applies in for v3.0.0


commy2:
CRLF

Killswitch00:
We may need this mechanism in the future when the ports catch up and then are left behind again.

commy2:
true, but then we can just go back in git history to get the linux.pbo

Killswitch00:
Fair enough.

commy2:
testing ...

commy2:
works

PR #435, opened 2016-07-14
Convert to LF line endings

jonpas:
- Changes default EOL to LF for Git setting (in case where `core.autocrlf` is not enabled)
- Normalizes files that weren't in LF already (most were due to `core.autocrlf` being set on msysgit by default)
- Adds `.editorconfig` file for EditorConfig automatic editor configuration

The advantage of this is that it causes no issues on Linux, where CRLF is not supported unless you are a using a text editor where you can switch it, and Windows has `core.autocrlf` settings for that reason exactly. Additionally it normalizes it across the repo, no one can commit wrong line endings, thus reducing merge conflicts and ensures consistency.

After checking this out (and when merged into master) people might see full files being changed, that's because in their working directories line endings have not changed. Usually, committing those files just makes Git realize it's fine and just does it behind the hood without an actual commit. Anyone who wants to get rid of that annoyance can simply run the following in bash:
```
git ls-files -z | xargs -0 rm
git checkout .
```


Killswitch00:
* Add a LF at the end of .gitignore
* What is a/the "EditorConfig"? What uses that?

jonpas:
> Add a LF at the end of .gitignore

Done.

> This PR appears OK. Before I merge it, how does it work for you people out there using git in Windows?

There is no difference, Git for Windows by default sets `core.autocrlf` to `true`, which changes line endings to CRLF on checkout and back to LF on commit. With this PR even those that don't have that option enabled will commit in LF, because .gitattributes has priority over local user setting. Basically it eliminates the possibility of committing CRLF (unless someone is using an ancient version of Git).

Killswitch00:
Neat.

commy2:
YES :+1:

PR #437, opened 2016-07-16
Rework Settings

PabstMirror:
From discussion https://github.com/acemod/ACE3/pull/4024#issuecomment-230994535

For global settings, if the setting is forced by mission and not by server, then the mission setting will be used..

Adds called events `CBA_beforeSettingsInitialized` and `CBA_settingsInitialized`
Adds listen event `CBA_settings_setSettingMission`

These will allow reading of modules for ace, which should allow us to keep backwards compatibility.
The basic flow would be at `CBA_beforeSettingsInitialized` we could read settings from modules and set them as mission settings.

For global settings priority would be
- Server Forced
- Mission Forced
- Server


commy2:
Is this done?

PabstMirror:
Should be good to go

PR #438, opened 2016-07-16
add scripted garrison waypoint

commy2:
- Adds a scripted waypoint called "GARRISON"
- makes the group enter nearby buildings and static weapons
- based on `CBA_fnc_taskDefend`

### DIFF
--- a/addons/ai/CfgWaypoints.hpp
+++ b/addons/ai/CfgWaypoints.hpp
@@ -1,0 +1,11 @@
+
+class CfgWaypoints {
+    class A3 { // called "Advanced"
+        class CBA_Task_Garrison {
+            displayName = "GARRISON"; // all caps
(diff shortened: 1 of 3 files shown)


addons/ai/CfgWaypoints.hpp
```diff
@@ -0,0 +1,11 @@
+
+class CfgWaypoints {
+    class A3 { // called "Advanced"
+        class CBA_Task_Garrison {
+            displayName = "GARRISON"; // all caps
```
Killswitch00: Put the display string in the stringtable.xml so it can be translated

commy2: was about to write that. unlocalized and all caps. probably to get consistency with the scripting commands. https://community.bistudio.com/wiki/setWaypointType

PR #439, opened 2016-07-16
add 'setUnitPos' control to 3den

commy2:
- Implements the rejected `setUnitPos` control from BI into 3den

### DIFF
--- a/addons/ai/Cfg3DEN.hpp
+++ b/addons/ai/Cfg3DEN.hpp
@@ -1,0 +1,54 @@
+
+class Cfg3DEN {
+    class Object {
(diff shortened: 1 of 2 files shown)


PabstMirror:
Biki says `setUnitPos` needs local unit, but this worked great on dedicated with AI under my control.
Maybe change tittle to "AI Stance" to indicate that it will not have effect on player?

commy2:
My testing showed that the command has to be executed on every machine and only the state of the local machine is respected. Executing it on every machine ensures that it works after changing locality though.

commy2:

PR #441, opened 2016-07-16
add createPerFrameHandlerObject

commy2:
- Reimplements `CBA_fnc_createPerFrameHandlerObject` (formerly known as `cba_common_fnc_addPerFrameHandlerLogic`)

Someone in the ACE public chat asked what happened to this. I guess this was meant to be API after all, so now I feel bad for nuking it. Should've been better documented though tbqh.

old one:
https://github.com/sickboy/cba/blob/542e28c71165864a32211a347a72b206d91ef882/addons/common/init_perFrameHandler.sqf#L58

Since this was removed before 2.4 I also took the opportunity to make some tweaks:
- Use a location instead of an object for minimal overhead.
- Switched around `_args` and `_delay` parameter to make it consistent with `addPerFrameHandler`
- The start code now also gets the logic as argument instead of the passed arguments directly. This makes it consistent with all the other called functions.
- And obviously removed `_common_` from the function name.
- `private keyword`, `pushBack` etc. etc.

### DIFF
--- a/addons/common/fnc_createPerFrameHandlerObject.sqf
+++ b/addons/common/fnc_createPerFrameHandlerObject.sqf
@@ -109,7 +109,7 @@ private _handle = [{
     private (_logic getVariable "private");
     { call _x } forEach (_logic getVariable "deserialize");
 
-    // check exit condition - exit if false
+    // check exit condition - exit if true
     if (_logic call (_logic getVariable "exit_condition")) exitWith {
         // execute end code
         _logic call (_logic getVariable "end");
--- a/addons/common/fnc_createPerFrameHandlerObject.sqf
+++ b/addons/common/fnc_createPerFrameHandlerObject.sqf
@@ -1,5 +1,5 @@
 /* ----------------------------------------------------------------------------
-Function: CBA_fnc_createPerFrameHandlerLogic
+Function: CBA_fnc_createPerFrameHandlerObject
 
 Description:
     Creates a PFH object, that will execute code every frame, or every x number of seconds.
@@ -47,7 +47,7 @@ Examples:
             { random 1 > 0.5 },
             { random 1 > 0.8 },
             "_test"
-        ] call CBA_fnc_createPerFrameHandlerLogic;
+        ] call CBA_fnc_createPerFrameHandlerObject;
     (end)
 
 Author:
--- a/addons/common/fnc_deletePerFrameHandlerObject.sqf
+++ b/addons/common/fnc_deletePerFrameHandlerObject.sqf
@@ -1,0 +1,30 @@
+/* ----------------------------------------------------------------------------
+Function: CBA_fnc_deletePerFrameHandlerObject
+
+Description:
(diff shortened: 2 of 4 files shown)


Glowbal:
I think the name `addPerFrameHandlerLogic` is strange and doesn't properly say what the function does. I'm not sure exactly what a better name should be yet.

commy2:
`createPerFrameHandlerLogic` ?

Glowbal:
It's the logic bit that makes it not fit, I think. Though create is indeed better.

commy2:
Hmm. Other suggestion: `createPerFrameHandlerObject`. Although `Logic` is the term for dummy objects that do magical script things in Arma. (e.g. `class Logic;`)

commy2:
I'm thinking about using these for my port of ACE2 fire propagation.

Glowbal:
It is, but I see no reason to indicate we use logic objects in the function name. I will think of some suggestions for a better name.

commy2:
The logic is a fundamental part of this though. It is the return value and the only passed argument in the called functions (aside from the `_private` functions).

Killswitch00:
Object? Entity? Thing? Manager? Agent? (perhaps not, since the concept of "agents" exist in the game).

(Logic seems fine though)

commy2:
`Object` would perfectly describe the character of this being. But the problem is that the game already has \<OBJECT\> as data type and this is actually a \<LOCATION\>...

Killswitch00:
Runner? `CBA_fnc_createPerFrameHandlerRunner`

Glowbal:
Object is the best of those so far. But you could also just leave it out and use `createPerFrameHandler`. 

Logic might be confused with some kind of internal function that handles the per frame handler logic.

Killswitch00:
How about a remote/destroy/delete function as well?

addons/common/fnc_createPerFrameHandlerObject.sqf
```diff
@@ -0,0 +1,140 @@
...
+
+    if (isNil "_logic" || {isNull _logic}) exitWith {
+        (_logic getVariable "handle") call CBA_fnc_removePerFrameHandler;
+    };
+
+    // deserialize
+    private (_logic getVariable "private");
+    { call _x } forEach (_logic getVariable "deserialize");
+
+    // check exit condition - exit if false
```
Killswitch00: exit if true

addons/common/fnc_deletePerFrameHandlerObject.sqf
```diff
@@ -0,0 +1,30 @@
...
+
+Author:
+    commy2
+---------------------------------------------------------------------------- */
+#include "script_component.hpp"
+
+params [["_logic", locationNull, [locationNull]]];
+
+if (!isNull _logic) then {
+    _logic call (_logic getVariable "end");
```
Killswitch00: How about changing the handler's run and exit conditions to make it "end itself"?

Killswitch00:
Looks about done now.

PR #442, opened 2016-07-16
Update geometry related functions

SilentSpike:
- Add a new getArea function to return the array form of any area construct
- Function randPosArea now accepts areas in the form of an array or location (same as the `inArea` command)
- Function taskSearchArea now accepts areas in the form of an array or location
- Function inArea deprecated in favour of `inArea` command

I'm not sure if CBA has a deprecation system we might want to use for the inArea function now that there's an equivalent engine command


jonpas:
Yes there is a "system", reference [this](https://github.com/CBATeam/CBA_A3/blob/15d2518ae2a041f75a55cf480fe36688b0e8be30/addons/common/backwards_comp.sqf). Basically move it into a backwards compat file so it doesn't get included in documentation and prints warnings. CBA does not remove deprecated functions like we do in ACE3.

commy2:
Suggestion for `fnc_randPosArea`.
Can't we simply make it create a random position with `fnc_randPos` and if the result is not `inArea`, then we discard it and try again?

commy2:
`[[10,10], 0] call cba_fnc_randPos`

Before:
`[10,10]`

After:
`[10,10,0]`

While `getPos` supports 2d positions, the return value is always a 3d position. That is not true for our function.

I have no idea why they chose to make this the alternative syntax of `getPos`. It is completely different from the original purpose.

addons/common/fnc_getArea.sqf
```diff
@@ -0,0 +1,62 @@
+/* ----------------------------------------------------------------------------
+Function: CBA_fnc_getArea
+
+Description:
+    Returns the array form of any area construct:
+    [center, a, b, angle, isRectangle]
```
commy2: Has to be a single line. The format should probably be in L:09 f anyway though

SilentSpike:
> Can't we simply make it create a random position with fnc_randPos and if the result is not inArea, then we discard it and try again?

While rejection sampling would work, the reason I originally wrote this function was to avoid that since it requires looping an indeterminate amount of times. This way you're guaranteed to get a position in a single shot.

commy2:
But that "single shot" could take way longer on average. I guess it also depends on the marker shape. For circles and hexagons it will on average be faster, but it could take a long time on narrow rectangles and ellipses.

SilentSpike:
My thinking is that you'd use the function for a guaranteed result, whereas if you wanted to use rejection sampling for possibly quicker execution you'd just write it manually

SilentSpike:
Trying to figure out if there's a way to optimise this part of randPosArea:

```
        private _rho = random (4*(_a + _b));

        private _x1 = (_rho min _2a);
        private _y1 = ((_rho - _x1) min _2b) max 0;
        private _x2 = ((_rho - _x1 - _y1) min _2a) max 0;
        private _y2 = ((_rho - _x1 - _y1 - _x2) min _2b) max 0;
        _posVector = [(_x1 - _x2) - _a, (_y1 - _y2) - _b, 0];
```

It's used to uniformly find a position along the perimeter of a rectangle, but I feel like there must some way to condense it down.

addons/common/backwards_comp.sqf
```diff
@@ -71,3 +71,11 @@ CBA_fnc_intToString = {
     if (isNil "_int") exitWith {""};
     str _int
 };
+
+CBA_fnc_inArea = {
```
Killswitch00: This won't be compileFinal:ed, which, if I'm not mistaken, CfgFunctions functions are.

commy2: That's a general problem with this file and should probably be fixed in another PR

Killswitch00: Indeed, so we should not add to the problem.

Instead, change the old function file and keep its CfgFunctions entry.

commy2: The problem with that is, that the function will appear in the config viewer. Deprecated functions shouldn't appear there in my opinion.

PR #444, opened 2016-07-24
Disable default settings button if setting is already default

PabstMirror:
For CBA Settings, this will grey-out (`ctrlEnabled false`) the button for default settings if the setting is already set to the default value

I think this will make it easier to find settings that have been changed and easier to get all settings back to default.

Quick demo: https://www.youtube.com/watch?v=A90L1dy9dXE

Also disabled the Import button when in multiplayer as it did not work.

### DIFF
--- a/addons/settings/gui_createMenu_color.sqf
+++ b/addons/settings/gui_createMenu_color.sqf
@@ -1,7 +1,7 @@
 // inline function, don't include script_component.hpp
 
-private _ctrlSettingPreview = _display ctrlCreate ["RscText", count _contols + IDC_OFFSET_SETTING, _ctrlOptionsGroup];
-_contols pushBack _ctrlSettingPreview;
+private _ctrlSettingPreview = _display ctrlCreate ["RscText", count _controls + IDC_OFFSET_SETTING, _ctrlOptionsGroup];
+_controls pushBack _ctrlSettingPreview;
 
 _ctrlSettingPreview ctrlSetPosition [
     POS_W(9.5),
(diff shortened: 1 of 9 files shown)


Killswitch00:
Looks nice.

addons/settings/gui_createMenu_color.sqf
```diff
@@ -98,6 +103,11 @@ for "_index" from 0 to (((count _defaultValue max 3) min 4) - 1) do {
         _color set [_index, _value];
         _ctrlSettingPreview ctrlSetBackgroundColor _color;
         SET_TEMP_NAMESPACE_VALUE(_setting,_currentValue,_source);
+
+        //If new setting is same as default value, grey out the default button
+        (_control getVariable QGVAR(linkedControls)) params ["", "", "_defaultControl"];
```
commy2: should be added to L92/97

addons/settings/gui_createMenu_slider.sqf
```diff
@@ -59,6 +64,11 @@ _ctrlSettingEdit ctrlAddEventHandler ["KeyUp", {
 
     _value = sliderPosition _linkedControl;
     SET_TEMP_NAMESPACE_VALUE(_setting,_value,_source);
+
+    //If new setting is same as default value, grey out the default button
+    (_control getVariable QGVAR(linkedControls)) params ["", "", "_defaultControl"];
```
commy2: Should be moved to L57/62 and be used to define `_linkedControl` directly.

addons/settings/gui_createMenu_slider.sqf
```diff
@@ -30,10 +30,15 @@ _ctrlSetting ctrlAddEventHandler ["SliderPosChanged", {
     _linkedControl ctrlSetText ([_value, 1, _trailingDecimals] call CBA_fnc_formatNumber);
 
     SET_TEMP_NAMESPACE_VALUE(_setting,_value,_source);
+    
+    //If new setting is same as default value, grey out the default button
+    (_control getVariable QGVAR(linkedControls)) params ["", "", "_defaultControl"];
```
commy2: Should be moved to L29 and be used to define `_linkedControl` directly.

addons/settings/gui_createMenu_default.sqf
```diff
@@ -23,6 +23,8 @@ _ctrlSettingDefault ctrlAddEventHandler ["ButtonClick", {
     (_control getVariable QGVAR(data)) params ["_setting", "_source", "_defaultValue", "_settingType", "_settingData"];
     SET_TEMP_NAMESPACE_VALUE(_setting,_defaultValue,_source);
 
+    _control ctrlEnable false;
```
commy2: why?

commy2:
Looks good. Suggested minor optimizations. After that can be merged.

addons/settings/gui_createMenu_color.sqf
commy2: should be added to L92/97

commy2: OK. IDK why, but it keeps messing up the line of this ^  comment. Meant for L108.

PabstMirror:
Fixed

PR #447, opened 2016-07-28
Add function to filter hashes

PabstMirror:
CBA_fnc_hashEachPair fails when removing keys from within the loop.

```
theHash = [[["aa", 11], ["bb", 22], ["cc", 33], ["dd", 44], ["ee", 55], ["ff", 66]]] call CBA_fnc_hashCreate;  
_dumpHash = { 
    diag_log format ["Key: %1, Value: %2", _key, _value]; 
    [theHash, _key] call CBA_fnc_hashRem; 
};
[theHash, _dumpHash] call CBA_fnc_hashEachPair; 
```
"Key: aa, Value: 11"
"Key: cc, Value: 33"
"Key: ee, Value: 55"
theHash = ["#CBA_HASH#",["bb","dd","ff"],[22,44,66],any]

We could replace instead of creating a new function, but the new version is a bit slower than the old one.

Ref: https://github.com/acemod/ACE3/pull/4171

### DIFF
--- a/addons/hashes/fnc_hashFilter.sqf
+++ b/addons/hashes/fnc_hashFilter.sqf
@@ -3,7 +3,7 @@ Function: CBA_fnc_hashFilter
 
 Description:
     Iterate through all keys and values in a Hash.
-    If function returns true, the key is removed from the hash.
+    If function returns false, the key is removed from the hash (just like `[] select {}`)
 
     Data passed to the function on each iteration,
     * _key - Key from the Hash.
@@ -13,7 +13,7 @@ Description:
 
 Parameters:
     _hash - Hash to iterate [Array which is a Hash structure]
-    _code - Function to call with each pair which returns a bool (true will remove key from hash) [Code]
+    _code - Function to call with each pair which returns a bool (false will remove key from hash) [Code]
 
 Returns:
     Number of removed entrys [Number]
@@ -23,7 +23,7 @@ Example:
     _hash = [[["A1", 1], ["A2", 1], ["B", 2], ["C", 3], ["D1", 4], ["D2", 4], ["E1", 5], ["E2", 5]]] call CBA_fnc_hashCreate;
     _removeOddValues = {
         diag_log format ["Key: %1, Value: %2", _key, _value];
-        ((_value % 2) == 1)
+        ((_value % 2) == 0)
     };
     _removedCount = [_hash, _removeOddValues] call CBA_fnc_hashFilter;
     (end)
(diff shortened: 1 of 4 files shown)


Killswitch00:
"Safe" suggests the other one is "unsafe" and should no longer be used. I wouldn't call `hashEachPair` "unsafe for use", it's just that one must not modify the hash when using `hashEachPair`.

How about calling the new function `hashEachPairRemove`, where the name hints at there being a possibility the hash is modified? It could return the number of elements removed, aswell.

(Inspiration taken from [g_hash_table_foreach_remove()](https://developer.gnome.org/glib/stable/glib-Hash-Tables.html#g-hash-table-foreach-remove)

commy2:
Unnecessary imo.

```
theHash = [[["aa", 11], ["bb", 22], ["cc", 33], ["dd", 44], ["ee", 55], ["ff", 66]]] call CBA_fnc_hashCreate;   
_dumpHash = {  
    diag_log format ["Key: %1, Value: %2", _key, _value];  
    [theHash, _key] call CBA_fnc_hashRem;  
}; 
[+theHash, _dumpHash] call CBA_fnc_hashEachPair; 
```
Just copy the array before iterating through it.

PabstMirror:
@commy2 that should work great for what I need.

In testing, this function is faster when removing items, but slower when not.
For what I need, the common case is not going to remove anything on most runs.

```
[+theHash, _dumpHash] call CBA_fnc_hashEachPair; 
remove 6/6: 0.268673 ms
remove 3/6: 0.172265 ms
remove 1/6: 0.101554 ms
remove 0/6: 0.0692 ms

[theHash, _dumpHash] call CBA_fnc_hashEachPairRemove
remove 6/6: 0.0932 ms
remove 3/6: 0.092 ms
remove 1/6: 0.0869 ms
remove 0/6: 0.0797 ms
```

commy2:
This function is similar to `select <CODE>`, so maybe it should be named `CBA_fnc_hashSelect`.

commy2:
The name of the function is obvious! Name it `hashFilter`

Killswitch00:
`hashFilter` is nice and short

PabstMirror:
Updated func name

addons/hashes/fnc_hashFilter.sqf
```diff
@@ -0,0 +1,59 @@
+/* ----------------------------------------------------------------------------
+Function: CBA_fnc_hashFilter
+
+Description:
+    Iterate through all keys and values in a Hash.
+    If function returns true, the key is removed from the hash.
```
commy2: Maybe it should remove it when the function returns false instead. Like `select <CODE>` does.

Killswitch00: I don't know - it seems fine as it is (remove if true)

- `count` will count an array element if the expression is `true`
- `select` will use an array element if the expression is `true`
- imagine two CBA hash functions `hashCount` and `hashSelect` that would do for the same  for hashes

--> `hashFilter` modifying the hash when the function provided returns `true` sounds consistent

commy2:
:+1:

PR #449, opened 2016-07-31
parse settings array instead of compiling them

commy2:

### DIFF
--- a/addons/settings/fnc_parse.sqf
+++ b/addons/settings/fnc_parse.sqf
@@ -28,6 +28,30 @@ Author:
 #define ASCII_ARRAY_OPEN (toArray "[" select 0)
 #define ASCII_ARRAY_CLOSE (toArray "]" select 0)
 
+private _fnc_parseArray = {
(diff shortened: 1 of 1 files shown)


addons/settings/fnc_parse.sqf
```diff
@@ -28,6 +28,30 @@ Author:
 #define ASCII_ARRAY_OPEN (toArray "[" select 0)
 #define ASCII_ARRAY_CLOSE (toArray "]" select 0)
 
+private _fnc_parseArray = {
```
Killswitch00: The function takes a string as input, so `_fnc_parseString`?

commy2: It's a stringified array. `"[<...>]"`. `parseNumber` for example also takes \<STRING\> and not \<NUMBER\>

```
parseNumber "1"
-> 1
```

Killswitch00: Fair enough.

PR #450, opened 2016-07-31
add optional pbo to load settings from userconfig file

commy2:
- Adds optional PBO that auto loads settings from a settings file
- Can be used on dedicated servers to have settings that not have to be set via gui (e.g. when the mod is only server side and the settings wouldn't show up for the clients)

`userconfig\cba\settings.sqf`

### DIFF
--- a/optionals/auto_load_settings_file/XEH_preInit.sqf
+++ b/optionals/auto_load_settings_file/XEH_preInit.sqf
@@ -1,0 +1,12 @@
+#include "script_component.hpp"
+
+ADDON = false;
+
+// frame after preInit, but before postInit
+0 spawn {
+    {
+        #include "loadSettingsFile.sqf"
+    } call CBA_fnc_directCall;
+};
+
+ADDON = true;
--- a/optionals/auto_load_settings_file/loadSettingsFile.sqf
+++ b/optionals/auto_load_settings_file/loadSettingsFile.sqf
@@ -1,0 +1,9 @@
+
+if (isFilePatchingEnabled) then {
+	private _source = ["client", "server"] select (isMultiplayer && isServer);
+
+    [loadFile PATH_SETTINGS_FILE, _source] call FUNC(import);
+    diag_log text "[CBA] (settings): Settings file loaded.";
+} else {
+    diag_log text "[CBA] (settings): Cannot load settings file. File patching disabled. Use -filePatching flag.";
+};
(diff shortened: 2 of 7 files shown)


optionals/auto_load_settings_file/XEH_preInit.sqf
```diff
@@ -0,0 +1,12 @@
+#include "script_component.hpp"
+
+ADDON = false;
+
+// frame after preInit, but before postInit
+0 spawn {
```
esteldunedain: Maybe it would be safer and cleaner to have cba launch an event when preInit finishes

optionals/auto_load_settings_file/loadSettingsFile.sqf
```diff
@@ -0,0 +1,14 @@
+
+if (isFilePatchingEnabled) then {
+    private _source = ["client", "server"] select (isMultiplayer && isServer);
+    private _file = loadFile PATH_SETTINGS_FILE;
+
+    if (_file != "") then {
+        [_file, _source] call FUNC(import);
```
PabstMirror: `EFUNC(settings,import)`

and import calls `FUNC(gui_refresh)` which won't work on a dedicated or a client that hasn't opened the menu
maybe a call to `CBA_settings_fnc_set` with an extra flag that doesn't change the profileNamespace setting?

optionals/auto_load_settings_file/XEH_preInit.sqf
Killswitch00: This looks about as safe and clean as it gets - only depends on `directCall`.

commy2: I would've used `CBA_fnc_execNextFrame`, but the `EachFrame` handle is installed at postInit - many frames after preInit.

optionals/auto_load_settings_file/loadSettingsFile.sqf
commy2: will fix

optionals/auto_load_settings_file/XEH_preInit.sqf
esteldunedain: It's a long shot, but maybe some other mod could conceivably clog the scheduler during preinit, preventing this from running before postInit.

PabstMirror:
@commy2 - this still has no effect on dedicated

`CBA_settings_fnc_import` just saves the to the temp namespace and then exits

`isNull (uiNamespace getVariable ["cba_settings_display", displayNull])` 
is always going to be true on dedicated and will be true on clients if that menu is closed

commy2:
It's 
`!isNull (uiNamespace getVariable [QGVAR(display), displayNull])`
and that is not always true.

I'll see about that temp namespace issue.

PR #451, opened 2016-07-31
add template for a settings file inside a pbo

commy2:
- Adds a template that can be binarized to have a signed settings file inside a PBO
- Can be used to enforce settings for a whole repository

@PabstMirror Would that work for you?


jonpas:
That looks handy!

Killswitch00:
> ...enforce settings for a whole repository

The term "repository" is confusing. (It is not a term usually associated with playing the game. A user of CBA may wonder "what/where is a/the 'repository'?)

commy2:
Other suggestion? "modpack" sounds weird.

Killswitch00:
How about "static settings addon" instead of "repository settings file".

The template shows an example of an addon that can be used to enforce settings for a mod/addon collection or a mission/mission type.

Killswitch00:
Looks good. Needs documentation that users can read (what it is, what it can be used for and so on). This could be in the form of a separate text file in the templates folder or placed in a comment header in the `config.cpp`

Killswitch00:
Ok, so no docs then.

PR #454, opened 2016-08-06
Make validation scripts work with Python 2

Killswitch00:

- Make it possible to run the config and sqf validation scripts with Python 2

With Python 2, use the codecs module to read UTF-8 files.

### DIFF
--- a/tools/config_style_checker.py
+++ b/tools/config_style_checker.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python3
+#!/usr/bin/env python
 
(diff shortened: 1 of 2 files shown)


Glowbal:
Might want to consider adding travis support for python2.

in `.travis.yml`:

```yml
python:
  - '2.7'
  - '3.5'
```

jonpas:
No need, as the script is compatible with both Python 2 and 3, codecs is only used if Python 2 is used.

PR #455, opened 2016-08-07
Fix default value

AACO:
Fix default value, so we don't get an extra custom attributes per unit, even when the unit's stance hasn't changed

Helps keep the files size down on large missions/frameworks.


PabstMirror:
BIS is (re-)adding the stance setting to 3den in 1.64
`Added: An engine implementation of the Stance attribute`
https://twitter.com/KarelMoricky/status/760819598031724545

So I think we should just remove CBA's version.

commy2:

PR #457, opened 2016-08-08
add 'CBA_fnc_removeWhitespace'

commy2:
Ports and improves `ace_common_fnc_stringRemoveWhiteSpace` (https://github.com/acemod/ACE3/blob/fd26e2e44ec5dadcb08f856b9769173bb33d9158/addons/common/functions/fnc_stringRemoveWhiteSpace.sqf) as suggested

### DIFF
--- a/addons/strings/test_strings.sqf
+++ b/addons/strings/test_strings.sqf
@@ -299,4 +299,18 @@ _result = ["aardwolf", "aardvark"] call CBA_fnc_compare;
 TEST_OP(_result, ==, +1, _fn);
 */
 
-nil;
+// ----------------------------------------------------------------------------
+// UNIT TESTS (removeWhitespace)
+_fn = "CBA_fnc_removeWhitespace";
+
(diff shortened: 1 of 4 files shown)


addons/strings/test_strings.sqf
```diff
@@ -299,4 +299,18 @@ _result = ["aardwolf", "aardvark"] call CBA_fnc_compare;
...
+TEST_DEFINED("CBA_fnc_removeWhitespace","");
+
+_str = " foo  bar   " call CBA_fnc_removeWhitespace;
+_expected = "foobar";
+TEST_OP(_str,isEqualTo,_expected,_fn);
+
+_str = [" foo  bar   ", true] call CBA_fnc_removeWhitespace;
+_expected = "foo bar";
+TEST_OP(_str,isEqualTo,_expected,_fn);
+
```
Killswitch00: Test removal of tabs and newlines as well.

commy2: done

PR #458, opened 2016-08-09
remove obsolete stance setting from 3den

commy2:
Is in vanilla since 1.64, removes duplicate


jonpas:
It served it's purpose... not. :laughing:

commy2:

PR #465, opened 2016-08-14
simplify 'CBA_fnc_filter'

commy2:
- Simplifies 'CBA_fnc_filter' as suggested by @pennyworth


Glowbal:
:+1:

PR #469, opened 2016-08-18
compileFinal all missionNamespace functions

commy2:
- adds: `CBA_fnc_compileFinal`
- See: https://github.com/CBATeam/CBA_A3/pull/442#discussion_r71128418


commy2:
Could be merged now. Other components should get their own (general cleanup) PR's.

PR #473, opened 2016-08-20
remove non functional marker persistency from network component

commy2:
- `CBA_fnc_set/add/remove/getMarkerPersistent` do nothing in Arma 3. All global markers are always persistent
- No one uses these functions, so we should just nuke them.


jonpas:
:boom:

Killswitch00:
Ideally, we'd have a "stop using this function, it will be removed soon" helper macro similar to `DEPRECATE`. How did you determine that "no one uses these functions"?

commy2:
>How did you determine that "no one uses these functions"?
All I did was a google search. At least no open source project is using it. All I could find is A2 scripts from 2010.
I really doubt though any closed project uses it either. The function never did anything in Arma 3. There was a need for this back in A2, but not anymore.

I could make DEPRECATED functions for these (at least for `set` and `get`, `add` and `remove` where already deprectaed years ago.) But all that would do is write an error message into the RPT, nothing else.
If someone actually uses these functions and we remove them, all that would happen is that a "undefined variable used" error is logged. So essentially the same thing as if we properly deprecate them.
I somehow feel like making a function just to exchange one error for another is not worth it.

Killswitch00:
Ok, let's ditch them.

dixon13:
https://github.com/CBATeam/CBA_A3/blob/64034f59109867adb92e830035b9159a7282db07/addons/common/fnc_createMarker.sqf#L62

Above is still using `CBA_fnc_setMarkerPersistent `

PR #475, opened 2016-08-21
Add macros 'PATHTO_FNC ' and 'RECOMPILE'

dixon13:
- Add the macros `PATHTO_FNC ` and `RECOMPILE`
- Argument for `PATHTO_FNC ` is the name of a function that is located at `\MAINPREFIX\PREFIX\SUBPREFIX\COMPONENT\` and the name of the file is `fnc_func.sqf` where 'func' is the name of you're function. `func` will be the argument for `PATHTO_FNC `
- `RECOMPILE` does not accept an argument. If `RECOMPILE` is defined before including `script_macros_common.hpp` in you're component, then the functions defined with `PATHTO_FNC ` will be recompiled at mission start. (For debug purposes)

These macros emulate the below...
```
class CfgFunctions {
    class func {
        file = "\MAINPREFIX\PREFIX\SUBPREFIX\COMPONENT\fnc_func.sqf";
        recompile = 0; //if RECOMPILE defined before including script_macros_common.hpp then recompile = 1
    };
};
```

The macro 'PATHTO_FNC ' can be found already in the components `common` and `settings`. Just making it official :)

#### Ignore the edits to `build.py`. I guess I removed some spaces or something but no edits to any of the code.

### DIFF
--- a/addons/main/script_macros_common.hpp
+++ b/addons/main/script_macros_common.hpp
@@ -834,7 +834,7 @@ Author:
     dixon13, commy2
  ------------------------------------------- */
 #define PATHTO_FNC(func) class func {\
-    file = QUOTE(PATHTOF(DOUBLES(fnc,func).sqf));\
+    file = QPATHTOF(DOUBLES(fnc,func).sqf);\
     RECOMPILE;\
 }
 
--- a/addons/main/script_macros_common.hpp
+++ b/addons/main/script_macros_common.hpp
@@ -804,8 +804,37 @@ Author:
     #define RECOMPILE recompile = 0
 #endif
 
+/* -------------------------------------------
+Macro: PATHTO_FNC()
+
+Description:
+    Defines a function inside CfgFunctions.
+
+    Full file path in addons:
+        '\MAINPREFIX\PREFIX\SUBPREFIX\COMPONENT\fnc_<FNC>.sqf'
+    Define 'RECOMPILE' to enable recompiling.
+
+Parameters:
+    FUNCTION NAME - Name of the function, unquoted <STRING>
+
+Examples:
+    (begin example)
+        // file name: fnc_addPerFrameHandler.sqf
+        class CfgFunctions {
+            class CBA {
+                class Misc {
+                    PATHTO_FNC(addPerFrameHandler);
+                };
+            };
+        };
+        // -> CBA_fnc_addPerFrameHandler
+    (end)
+
+Author:
+    dixon13, commy2
+ ------------------------------------------- */
 #define PATHTO_FNC(func) class func {\
-    file = QUOTE(PATHTOF(DOUBLES(fnc,func).sqf));\ // \MAINPREFIX\PREFIX\SUBPREFIX\COMPONENT\fnc_func.sqf
+    file = QUOTE(PATHTOF(DOUBLES(fnc,func).sqf));\
     RECOMPILE;\
 }
 
--- a/addons/main/script_macros_common.hpp
+++ b/addons/main/script_macros_common.hpp
@@ -550,7 +550,7 @@ Author:
 /* -------------------------------------------
 Macro: FILTER()
 Description:
-    Filters an array based on given code, then assigns the resulting array 
+    Filters an array based on given code, then assigns the resulting array
     to the original
 Parameters:
     ARRAY - Array to be filtered
@@ -795,6 +795,17 @@ Author:
     #define PREP(var1) ['PATHTO_SYS(PREFIX,COMPONENT_F,DOUBLES(fnc,var1))', 'TRIPLES(ADDON,fnc,var1)'] call SLX_XEH_COMPILE_NEW
     #define PREPMAIN(var1) ['PATHTO_SYS(PREFIX,COMPONENT_F,DOUBLES(fnc,var1))', 'TRIPLES(PREFIX,fnc,var1)'] call SLX_XEH_COMPILE_NEW
 #endif
+
+#ifdef RECOMPILE
+    #define RECOMPILE recompile = 1
+#else
+    #define RECOMPILE recompile = 0
+#endif
+
+#define F_FILEPATH(func) class func {\
+    file = QUOTE(PATHTOF(DOUBLES(fnc,func).sqf));\
+    RECOMPILE;\
+}
 
 #define FUNC(var1) TRIPLES(ADDON,fnc,var1)
 #define FUNCMAIN(var1) TRIPLES(PREFIX,fnc,var1)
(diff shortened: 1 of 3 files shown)


addons/main/script_macros_common.hpp
```diff
@@ -796,6 +796,17 @@ Macro: GVARMAIN()
     #define PREPMAIN(var1) ['PATHTO_SYS(PREFIX,COMPONENT_F,DOUBLES(fnc,var1))', 'TRIPLES(PREFIX,fnc,var1)'] call SLX_XEH_COMPILE_NEW
 #endif
 
+#ifdef RECOMPILE
+    #define RECOMPILE recompile = 1
```
commy2: does this work with Mikero's Tools? I think they error out now when trying to redefine a macro in config space.

commy2:
Would be cool if we could redefine it for `script_macros_mission.hpp`, so it doesn't use the addon space paths `PATHTOF`, but accesses the current mission folder.
Also we might want to think about a potentially better name than `F_FILEPATH` if we make it API.

addons/main/script_macros_common.hpp
Killswitch00: Add `#undef` to keep the macro parsers happy

jonpas:
> Ignore the edits to build.py. I guess I removed some spaces or something but no edits to any of the code.

Those edits are actually good, removing redundant whitespace. :+1: :laughing:

dixon13:
How about `FUNC_PATH` or `PATHTO_FUNC`?

commy2:
`PATHTO_FNC` I'd say. It's similar enough to `PATHTOF` to remember and `FNC` matches the `fnc_` tag of the files.

dixon13:
Done. Added variant of `PATHTO_FNC` to `script_macros_mission.hpp` to be used by mission makers.

jonpas:
If you want to emulate `PATHTOF` macro then shouldn't it be `PATHTOFNC`?

dixon13:
Personally I like `PATHTO_FNC`. It just seems more readable.

addons/main/script_macros_common.hpp
```diff
@@ -796,6 +796,19 @@ Macro: GVARMAIN()
     #define PREPMAIN(var1) ['PATHTO_SYS(PREFIX,COMPONENT_F,DOUBLES(fnc,var1))', 'TRIPLES(PREFIX,fnc,var1)'] call SLX_XEH_COMPILE_NEW
 #endif
 
+#ifdef RECOMPILE
+    #undef RECOMPILE
+    #define RECOMPILE recompile = 1
+#else
+    #undef RECOMPILE
```
commy2: If RECOMPILE already is undefined, you cannot undefine RECOMPILE. This might or might not error out with Mikeros Tools too.

addons/main/script_macros_common.hpp
```diff
@@ -796,6 +796,19 @@ Macro: GVARMAIN()
...
+    #define RECOMPILE recompile = 0
+#endif
+
+#define PATHTO_FNC(func) class func {\
+    file = QUOTE(PATHTOF(DOUBLES(fnc,func).sqf));\ // \MAINPREFIX\PREFIX\SUBPREFIX\COMPONENT\fnc_func.sqf
```
commy2: Should use a function (macro) header for explaining the usage instead of having in code comments:
```
/* -------------------------------------------
Macro: PATHTO_FNC()
 
Description:
    Defines a function inside CfgFunctions.
 
    Full file path in addons: '\MAINPREFIX\PREFIX\SUBPREFIX\COMPONENT\fnc_<FNC>.sqf'
    Define 'RECOMPILE' to enable recompiling for that function.

Parameters:
    FUNCTION NAME - Name of the function, unquoted <STRING>
 
Examples:
    (begin example)
        // file name: fnc_addPerFrameHandler.sqf
        class CfgFunctions {
            class CBA {
                class Misc {
                    PATHTO_FNC(addPerFrameHandler);
                };
            };
        };
        // -> CBA_fnc_addPerFrameHandler
    (end)
 
Author:
    dixon13
```

Or something like that.

commy2:
@jonpas making it `PATHTOFNC` would make it sound too similar to `PATHTOF`. Keep in mind that the macro does not expand to the file path alone, but also creates the encasing class responsible for defining the function. It only works inside `CfgFunctions`, while `PATHTOF` is more of a general utility thing. (for example used here in `PATHTOF`.

This reminds me. We should simplify `QUOTE(PATHTOF( ... ))` to `QPATHTOF( ... )` inside this macro.

dixon13:
Macro header added for explanation.

addons/main/script_macros_common.hpp
Killswitch00: There's a check for `#ifdef RECOMPILE` just before the `#undef`. That should do it, right?

Killswitch00: Yeah, undef is only needed in the first case, where `UNDEF` is defined.

dixon13:
This is done unless `QUOTE(PATHTOF( ... ))` is simplified to `QPATHTOF( ... )`

jonpas:
Yes it is, `QPATHTOF()` is the simplified version.

dixon13:
Ah ok. Didn't realize it was a macro already. Utilizing `QPATHOF()` now instead of `QUOTE()` and `PATHOF()`. Done..

addons/main/script_macros_common.hpp
Killswitch00: This still needs action.

marceldev89:
FYI, the RECOMPILE changes broke Mikero tools for ALiVE and probably others. We had to update all the CfgFunctions.hpp files. 😢

commy2:
Mikero's Tools can't handle macro redefinitions anymore. I guess you had a macro named `RECOMPILE` ?

marceldev89:
Yes we have a global macro `RECOMPILE` set in `script_mod.hpp`. Changing all the `recompile = RECOMPILE;` occurrences to `RECOMPILE;` fixes the errors in the `CfgFunctions.hpp` files but then it errors out on the `#define RECOMPILE 1` in `script_mod.hpp` with `In File P:\x\alive\addons\main\script_mod.hpp: Line 61 #define altered without a #undef`. Not sure how to go from there.

marceldev89:
```
#ifdef RECOMPILE
    #undef RECOMPILE
#endif
#define RECOMPILE recompile = 1
```
In `script_mod.hpp` seems to "solve" it. Good enough I guess.

commy2:
Yeah. This is the only solution. No idea why this change was made. It's really stupid.

PR #476, opened 2016-08-21
fix 'CBA_fnc_peak'

commy2:
- `call compile _varName` replaced with `missionNamespace getVariable _varName`
- reduce network traffic by using the new `CBA_fnc_serverEvent` and `CBA_fnc_ownerEvent` instead of `CBA_fnc_globalEvent` twice
- fix example was wrong (`cba_diagnostics_receive_peak` has a S too much)
- fix `CBA_fnc_peek` is broken (undefined `GVAR` instead of `QGVAR`):
```
18:52:49 Error in expression <nostic\peek';
params ["_variable"];



[cba_diagnostic_peek, _variable] call CBA>
18:52:49   Error position: <cba_diagnostic_peek, _variable] call CBA>
18:52:49   Error Undefined variable in expression: cba_diagnostic_peek
18:52:49 File x\cba\addons\diagnostic\fnc_peek.sqf, line 9
```
- formating

Since the function was broken anyway and actually never worked, we might as well fix the spelling and rename it to `CBA_fnc_peak` instead of `CBA_fnc_peek`


commy2:
Alternatively, we just remove this. It doesn't actually seem all that usefull.

PabstMirror:
peek was the right spelling

commy2:
Ok. But then it was misspelled here: https://github.com/CBATeam/CBA_A3/blob/master/addons/diagnostic/XEH_preInit.sqf#L16

addons/diagnostic/CfgFunctions.hpp
```diff
@@ -34,10 +34,10 @@ class CfgFunctions
                 file = "\x\cba\addons\diagnostic\fnc_log.sqf";
             };
             // CBA_fnc_peek
-            class peek
+            class peak
             {
-                description = "Peek at variable on the server To receive the variable content back, you will have to [""cba_diagnostics_receive_peak"", {_this call myFunction}] call CBA_fnc_addEventHandler;";
-                file = "\x\cba\addons\diagnostic\fnc_peek.sqf";
+                description = "Peak at variable on the server To receive the variable content back, you will have to [""cba_diagnostic_receivePeak"", {_this call myFunction}] call CBA_fnc_addEventHandler;";
```
jonpas: server **To** receive

PabstMirror:
If it's been broken for a long time that probably means no one is using it, could just remove.

commy2:
According to github, it never worked. At least not since 2009.

Killswitch00:
Peek, not peak, as mentioned.

Killswitch00:
Function removed, since it's been broken for so long.

PR #484, opened 2016-08-22
check data type of the arguments of 'CBA_fnc_addKeybind'

commy2:
- @Nathan423 was using the function like this:
`["United Operations", "Admin Console", ["player", [], -10, "_this call uo_fnc_selfInteractMenu", "main"], [DIK_APPS, [false, false, false]], false, "keydown"] call cba_fnc_addKeybind;`,
but the third argument is meant to be a string.

- This is a really nasty error, because that array is stored in the profile and will lead to the following persistent error message:
```
17:05:28 Error in expression <aram [_forEachIndex, []];
 
_actionEntry params [["_displayName", "", ["", []]], >
17:05:28   Error position: <params [["_displayName", "", ["", []]], >
17:05:28   Error Params: Type String, expected Array
17:05:28 File x\cba\addons\help\XEH_postClientInit.sqf, line 79
```

- The only way to remove this error is to clear the keybinding variable from the profile.
- This PR makes it so the function checks the input to prevent errors like this in future.

### DIFF
--- a/addons/keybinding/fnc_addKeybind.sqf
+++ b/addons/keybinding/fnc_addKeybind.sqf
@@ -48,7 +48,7 @@ _nullKeybind = [-1, [false,false,false]];
 params [
     ["_modName", "", [""]],
     ["_actionId", "", [""]],
-    ["_displayName", "", [""]],
+    ["_displayName", "", ["", []]],
     "_downCode",
     "_upCode",
     ["_defaultKeybind", _nullKeybind],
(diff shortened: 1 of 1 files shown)


PabstMirror:
3rd arg can be array 
Throws: `Error Params: Type Array, expected String`
But just letting it take `[]` for 3rd arg might not fix the problem with the bad code.

commy2:
>_displayName       - Pretty name, or an array of strings for the pretty name and a tool tip [String]

Oh, yeah. Missed that. Will have to do further stuff with this...

PR #490, opened 2016-08-30
playerEH - Optional call function if player defined

PabstMirror:
- Adds optional param to call event handler function if player is already defined.

This makes it simpler to use `addPlayerEventHandler` by adding an option to guarantee the function will be run even if the changed xxxEvent has already happened. 

Example use:
```
[] spawn {
    // Dummy playerEH - represents some other mod has already installed the player EH system
    ["unit", {}] call CBA_fnc_addPlayerEventHandler;
    sleep 0.1;

    // This will not print out anything as the unitEvent has already happened:
    ["unit", {systemChat format ["unit eh (False) - %1",_this]}] call CBA_fnc_addPlayerEventHandler;

    // This will print right away
    ["unit", {systemChat format ["unit eh (True) - %1",_this]}, true] call CBA_fnc_addPlayerEventHandler;
};
```

Ref https://github.com/acemod/ACE3/pull/4289

### DIFF
--- a/addons/events/fnc_addPlayerEventHandler.sqf
+++ b/addons/events/fnc_addPlayerEventHandler.sqf
@@ -17,7 +17,7 @@ Description:
 Parameters:
     _type      - Event handler type. <STRING>
     _function  - Function to add to event. <CODE>
-    _autoCall  - Call function immediately if player is defined already <BOOL>
+    _applyRetroactively - Call function immediately if player is defined already (optional, default: false) <BOOL>
 
 Returns:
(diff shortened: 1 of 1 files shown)


commy2:
Rename `_autoCall` to `_applyRetroactively` for consistency with [fnc_addClassEventHandler](https://github.com/CBATeam/CBA_A3/blob/64034f59109867adb92e830035b9159a7282db07/addons/xeh/fnc_addClassEventHandler.sqf#L13) ?

PabstMirror:
I like that var name better, switched.

PR #503, opened 2016-09-17
fix default value of 'CBA_fnc_hashHasKey', close #502

### LINKED ISSUE #502: Proper default value for hashHasKey
@commy2 

https://github.com/CBATeam/CBA_A3/blob/master/addons/hashes/fnc_hashHasKey.sqf#L26-L28

a

_sorry for this few detail, I'm in a hurry, but don't want to forget_

commy2:
- Default value should be an empty hash and not an empty array. Otherwise `(_hash select 1)` will error out.
- closes #502

### DIFF
--- a/addons/hashes/fnc_hashHasKey.sqf
+++ b/addons/hashes/fnc_hashHasKey.sqf
@@ -23,6 +23,6 @@ Author:
 SCRIPT(hashHasKey);
 
 // -----------------------------------------------------------------------------
-params [["_hash", [], [[]]], "_key"];
+params [["_hash", [] call CBA_fnc_hashCreate, [[]]], "_key"];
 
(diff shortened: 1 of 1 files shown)


addons/hashes/fnc_hashHasKey.sqf
```diff
@@ -23,6 +23,6 @@ Author:
 SCRIPT(hashHasKey);
 
 // -----------------------------------------------------------------------------
-params [["_hash", [], [[]]], "_key"];
+params [["_hash", [] call CBA_fnc_hashCreate, [[]]], "_key"];
```
PabstMirror: The array is fully evaluated first, so `CBA_fnc_hashCreate` function would be called every time,
even on good input.

Could just use `[[], []]` which will be good enough for this function (`_hash select 1`)

PR #505, opened 2016-09-20
add event handler for creating and deleting markers

commy2:
- Adds `CBA_fnc_addMarkerEventHandler` and `CBA_fnc_removeMarkerEventHandler`
- Supports `created` and `deleted` events
- Passes the created or deleted marker name as `_this`

@bux578 is that what you needed?

### DIFF
--- a/addons/events/CfgFunctions.hpp
+++ b/addons/events/CfgFunctions.hpp
@@ -82,6 +82,8 @@ class CfgFunctions {
                 description = "Raises a CBA event on the target client ID's machine.";
                 file = "\x\cba\addons\events\fnc_ownerEvent.sqf";
             };
+            PATHTO_FNC(addMarkerEventHandler);
+            PATHTO_FNC(removeMarkerEventHandler);
         };
     };
 };
--- a/addons/events/fnc_addMarkerEventHandler.sqf
+++ b/addons/events/fnc_addMarkerEventHandler.sqf
@@ -42,7 +42,7 @@ if (isNil QGVAR(oldMarkers)) then {
 
             GVAR(oldMarkers) = _newAllMapMarkers;
         };
-    }, 0, GVAR(markerEvents)] call CBA_fnc_addPerFrameHandler;
+    }, 0] call CBA_fnc_addPerFrameHandler;
 };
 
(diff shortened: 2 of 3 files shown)


bux578:
@commy2 Looks perfect. How bad could performance suffer if `allMapMarkers` is checked each frame?
Maybe a slight delay could make sense here?

commy2:
>How bad could performance suffer if allMapMarkers is checked each frame?
I don't think much. `allMapMarkers` is not like `nearestObjects`. It just reads an already existing array out of C++ land afaik.

>Maybe a slight delay could make sense here?
maybe. although if you make the delay too big, you could miss some obviously. I guess 0.1 is fine?

>And what about automatic generated markers like from Blue Force Tracking etc...
it would execute for every marker that appears in `allMapMarkers`

jokoho48:
why not using a Statemachine?

commy2:
There are no states here.

esteldunedain:
I'm not sure about the delay; all CBA event handlers are instantaneous as far as I know, and having these ones behaving differently might be confusing. Also, it might break some potential usages that require low latency.

It all comes down to the cost of:

```sqf
private _newAllMapMarkers = allMapMarkers;
if !(_newAllMapMarkers isEqualTo GVAR(oldMarkers)) then {
```

have you measured the performance of that?

commy2:
`oldMarkers = allMapMarkers;`

```
private _newAllMapMarkers = allMapMarkers; 
if !(_newAllMapMarkers isEqualTo oldMarkers) then {};
```
0 markers: 0.0036 ms
100 markers: 0.0148 ms

100 markers using `private _newAllMapMarkers = allMapMarkers`
0.0137 ms

So `allMapMarkers` does get slower with many markers placed, but it isn't that bad. Not sure if that should run every frame though.

esteldunedain:
This are the theoretical frame decreases of running this every frame with 200 markers (assuming 0.015 x 2 = 0.03 ms):

60.0000 fps ->59.9999 fps
90.0000 fps -> 89.9997 fps
120.0000 fps -> 119.9995 fps

All in all completely unnoticeable.

commy2:
reverted it :)

PR #510, opened 2016-09-23
Add param for slider setting to only be whole numbers

PabstMirror:
Setting `_trailingDecimals` to be less than 0 will only allow whole numbers to be allowed.

E.G.

```
["Test_Setting_3", "SLIDER",   ["-test slider-",   "-tooltip-"], "My Category", [0, 10, 5, -1]] call cba_settings_fnc_init;
```

Useful for settings where the value needs to be an integer.
Like a max number of objects or a `for "_i" from 0 to Test_Setting_3 {`

### DIFF
--- a/addons/settings/fnc_check.sqf
+++ b/addons/settings/fnc_check.sqf
@@ -31,8 +31,8 @@ switch (toUpper _settingType) do {
         _value in _values
     };
     case ("SLIDER"): {
-        _settingData params ["_min", "_max"];
-        _value isEqualType 0 && {_value >= _min} && {_value <= _max}
+        _settingData params ["_min", "_max", "_trailingDecimals"];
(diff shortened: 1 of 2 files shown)


jonpas:
How is that different from setting it to 0?

PabstMirror:
With 0, the edit text will only say whole numbers, but the actual setting value will be a float.
So setting the slider about halfway between 0 and 10 and it would be 5.15010334

jonpas:
Ah, but is that intended behaviour? IMO it would be better to change the value it saves to the same amount of decimals as is displayed.

commy2:
Optional boolean parameter to make the value snap to however many decimals are set to be displayed?

BaerMitUmlaut:
Changelog of 1.66 had something about a stepped slider, does that deprecate this PR?

commy2:
No, this PR is still a thing. I'll merge it when I get back to https://github.com/commy2/CBA_A3/pull/4

commy2:
merged in #606

PR #511, opened 2016-09-23
remove obsolete team synch code

commy2:
- removes team synch code
- no longer needed as teams are synched since 1.62
- old versions are no longer supported by 3.1.0, because of changes in class inheritance for the JR component


PabstMirror:
ace still uses the event 
https://github.com/acemod/ACE3/blob/master/addons/interaction/functions/fnc_joinTeam.sqf#L21

commy2:
But does it do anything?

PabstMirror:
We call the cba event to change fireteams, should probably just make our own

jonpas:
Other people might be using the event as well.

commy2:
^ this should solve it

PR #534, opened 2016-10-25
Add projectile tracking

bux:
- add a function to track projectiles as colored lines
  - uses ~hashes and~ PFHs and should be performing better than `BIS_fnc_traceBullets`
  - works with vehicles
  - color gradients for projectile speed (normalized for one projectile; green: fast, red: slow)

Why CBA?:
- Useful for testing/diagnostics for a wider audience (mod makers, mission makers, communities)

---

**TODO:**
- [x]  make line counts variable


addons/diagnostic/fnc_projectileTracking_drawProjectilePaths.sqf
```diff
@@ -0,0 +1,61 @@
...
+
+            private _currentSpeed = _currentProjectileData select 1;
+
+            private _green = _currentSpeed / _startSpeed;
+            private _red = 1 - _green;
+
+            drawLine3D [_prevProjectileData select 0, _currentProjectileData select 0, [_red, _green, 0, 1]];
+
+            _i = _i + 1;
+        };
```
commy2: `while` is pretty slow in SQF due to having to execute two code blocks each cycle. Maybe use `forEach _projectileData` with `_currentProjectileData` == `_x`.

addons/diagnostic/fnc_projectileTracking_handleFired.sqf
```diff
@@ -0,0 +1,42 @@
...
+if ( [GVAR(projectileData), _index] call EFUNC(hash,hashHasKey) ) then {
+    private _data = [GVAR(projectileData), _index] call EFUNC(hash,hashGet);
+    private _handle = _data select 0;
+
+    [_handle] call EFUNC(common,removePerFrameHandler);
+    [GVAR(projectileData), _index] call EFUNC(hash,hashRem);
+};
+
+// using 0.1 to improve performance, we don't need that many bullet position to draw a line
+[FUNC(projectileTracking_trackProjectile), 0.1, [_projectile, _index]] call EFUNC(common,addPerFrameHandler);
```
commy2: `CBA_fnc_addPerFrameHandler`

commy2:
I think the slowest part about the BI functions is that it uses a too small resolution / too many lines per traced bullet.

bux578:
I'm done. Testing local MP with two PCs was working very well.

commy2:
FPS VA, no tracked bullets: 99
FPS VA, 30 tracked bullets via CBA: 24
FPS VA, 30 tracked bullets via BIS fnc: 11

used:
`player call cba_fnc_projectileTracking_addUnit`
and
`[player, 30] call BIS_fnc_traceBullets`
and emptied one MX mag

So this is a >100% improvement. Very good.
But 99>24 is still bad. I'm wondering if we can't push this further?

commy2:
`BIS_fnc_traceBullets` has an optional parameter to limit the number of drawn projectiles. It defaults to 10.
Even when only 10 bullets are drawn, the FPS dips to 27 (so your function can basically draw thrice as many lines with the same cost)

- `CBA_fnc_projectileTracking_addUnit` should be able to use this optional parameter too imo
- We should somehow integrate this into this standard VA mission you can access from the main menu (we should also add a debug console to it)
- I think the name is clumsy compared to the BI name "traceBullets". There is a rather nice naming convention with some new BI SQF commands that use `setUnitX`. `CBA_fnc_setUnitTraceBullets` with boolean parameter to turn it on and off? Or maybe `CBA_fnc_addUnitTraceBullets`,  `CBA_fnc_removeUnitTraceBullets`?

addons/diagnostic/fnc_projectileTracking_drawProjectilePaths.sqf
```diff
@@ -0,0 +1,62 @@
...
+Author:
+    bux578
+---------------------------------------------------------------------------- */
+#include "script_component.hpp"
+
+private _maxLines = GVAR(projectileMaxLines) min 20;
+
+for "_index" from 0 to _maxLines do {
+
+    if ( ([GVAR(projectileData), _index] call CBA_fnc_hashHasKey) ) then {
```
Glowbal: if we are going for performance, we are doing an `in` and a `find` with `hashHasKey` and `hashGet`. We can probably skip one of them and just do a `hashGet` followed by a `!isNil` check.

bux578:
> CBA_fnc_projectileTracking_addUnit should be able to use this optional parameter too imo

I've decided against using an optional parameter:

1. Too much hassle
2. It doesn't make sense for the "addUnit" function because the amount of paths drawn is not per unit but globally. Imo it's easier to just set a global variable to a different number.
3. I've set the function to max draw 20 paths even if a user enters e.g. `150`

> We should somehow integrate this into this standard VA mission you can access from the main menu (we should also add a debug console to it)

?

> I think the name is clumsy compared to the BI name "traceBullets". There is a rather nice naming convention with some new BI SQF commands that use setUnitX. CBA_fnc_setUnitTraceBullets with boolean parameter to turn it on and off? Or maybe CBA_fnc_addUnitTraceBullets, CBA_fnc_removeUnitTraceBullets?

The thing is, you don't only track bullets. Everything that triggers the "Fired" EH. That's why I went with "projectile".
I could change the public function names to `cba_fnc_addUnitTraceProjectiles`

> Not convinced that we need CBA hashes for this in the first place. I mean, this should be possible simply by iterating through an array, shouldn't it be?

I've started this thing as an exercise for me to get to know PFHs and hashes more. That's why I used hashes in the first place. I'll look into changing it into a standard array.

bux578:
@commy2 I've committed some of your requested changes.

PR #537, opened 2016-10-26
Add RETDEF() Macro, Improve RETNIL(), fix IS_INTEGER()

654wak654:
- Add a logical RETDEF() macro
- Check the function header for more

### DIFF
--- a/addons/main/script_macros_common.hpp
+++ b/addons/main/script_macros_common.hpp
@@ -539,27 +539,6 @@ Author:
 #define DEC(var) var = (var) - 1
 
 /* -------------------------------------------
-Macro: INV()
-
-Description:
-    Invert a number's sign. Opposite of the abs command with positive numbers.
-
-Parameters:
-    VAR - Variable to invert [Number]
-
-Example:
-    (begin example)
-    _degrees = 45;
-    INV(_degrees);
-    // _degrees => -45
-    (end)
-
-Author:
-    654wak654
-------------------------------------------- */
-#define INV(var) var = -var
-
-/* -------------------------------------------
 Macro: ADD()
 
 Description:
--- a/addons/main/script_macros_common.hpp
+++ b/addons/main/script_macros_common.hpp
@@ -371,7 +371,7 @@ Example:
 Author:
     Alef (see CBA issue #8514)
 ------------------------------------------- */
-#define RETNIL(VARIABLE) ([VARIABLE] param [0,nil])
+#define RETNIL(VARIABLE) ([VARIABLE] param [0])
 
 /* -------------------------------------------
 Macro: RETDEF()
--- a/addons/main/script_macros_common.hpp
+++ b/addons/main/script_macros_common.hpp
@@ -365,13 +365,34 @@ Parameters:
 Example:
     (begin example)
         // _var is undefined
-        hintSilent format ["_var=%1", RETNIL(_var) ]; // "_var=any"
+        hintSilent format ["_var=%1", RETNIL(_var)]; // "_var=any"
     (end example)
 
 Author:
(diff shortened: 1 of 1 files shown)


addons/main/script_macros_common.hpp
```diff
@@ -518,6 +539,27 @@ Macro: DEC()
...
+    (begin example)
+    _degrees = 45;
+    INV(_degrees);
+    // _degrees => -45
+    (end)
+
+Author:
+    654wak654
+------------------------------------------- */
+#define INV(var) var = (var) * -1
```
commy2: `var = -var`
?

addons/main/script_macros_common.hpp
```diff
@@ -365,13 +365,34 @@ Macro: RETNIL()
...
+    (begin example)
+        // _var is undefined
+        hintSilent format ["_var=%1", RETDEF(_var,5)]; // "_var=5"
+        _var = 7;
+        hintSilent format ["_var=%1", RETDEF(_var,5)]; // "_var=7"
+    (end example)
+Author:
+    654wak654
+------------------------------------------- */
+#define RETDEF(VARIABLE,DEFAULT_VALUE) if (isNil {VARIABLE}) then {DEFAULT_VALUE} else {VARIABLE}
```
commy2: Should be encapsulated in parenthesis. Same for `RETNIL()`. Would make them safe to be used in longer expressions.

Dorbedo: The following is faster:
``` #define RETDEF(VARIABLE,DEFAULT_VALUE) ([VARIABLE] param [0,DEFAULT_VALUE]) ```

File: addons/main/script_macros_common.hpp (hunk 1 above)
Comment (654wak654) [inline addons/main/script_macros_common.hpp]: That... Makes a lot of sense actually. Do you know what happens if `_var` is already negative?

Comment (jonpas) [inline addons/main/script_macros_common.hpp]: Mathemetically `--` == `+`.

File: addons/main/script_macros_common.hpp (hunk 2 above)
Comment (654wak654) [inline addons/main/script_macros_common.hpp]: It's shorter too, I like it

File: addons/main/script_macros_common.hpp [hunk 3]
```diff
@@ -365,13 +365,34 @@ Macro: RETNIL()
...
         // _var is undefined
-        hintSilent format ["_var=%1", RETNIL(_var) ]; // "_var=any"
+        hintSilent format ["_var=%1", RETNIL(_var)]; // "_var=any"
     (end example)
 
 Author:
     Alef (see CBA issue #8514)
 ------------------------------------------- */
-#define RETNIL(VARIABLE) if (isNil{VARIABLE}) then {nil} else {VARIABLE}
+#define RETNIL(VARIABLE) ([VARIABLE] param [0,nil])
```
commy2: Otherwise, maybe this is the fastest:

`(if !(isNil {VARIABLE}) then {VARIABLE})`

But I guess that depends on the expression of `VARIABLE`

commy2:
lgtm, untested

commy2:
Q:
Do we need `INV(var)` when it's just `- var`? I don't see the benefit tbf.

654wak654:
It makes it obvious what operation is being done, but yes having a macro for a single char does sound excessive.

commy2:
lgtm

commy2:
tested:
```
private _banana = 127;
diag_log RETNIL(_banana);
diag_log RETDEF(_banana,-1);
diag_log RETDEF(_potato,-1);
```

result:
```
13:47:31 127
13:47:31 127
13:47:31 -1
```

works, merged 👍 
thanks for the contribution : )

PR #538, opened 2016-10-26
Function headers with debugging callstack

BaerMitUmlaut:
- Add a small header to each compiled function, very similar to what BI does in `initFunctions.sqf`, but with an added root scope
- Make it possible to use a `RETURN`/`RETURN(value)` macro
- Add callstack logging and dumping when `DEBUG_MODE_FULL` is defined

The callstack logging and dumping happens automatically. It's a bit of overhead, but since it's only used for debugging I think that's fine. It's also not fully compatible with functions running in the scheduled environment, but there's nothing you can really do about that.

I've included a [test function](https://github.com/BaerMitUmlaut/CBA_A3/blob/compile-headers/addons/xeh/fnc_test_prep.sqf). When it was compiled with the callstack header and you run this code:

```
diag_log (5 call cba_xeh_fnc_test_prep);
call CBA_fnc_dumpCallstack;
```

The follwing will get dumped into your RPT file:

```
20:17:21 "test"
20:17:21 Dumping callstack:
-> cba_xeh_fnc_test_prep (via unknown function)
    -> cba_xeh_fnc_test_prep
        -> cba_xeh_fnc_test_prep
            -> cba_xeh_fnc_test_prep (via CBA_fnc_directCall)
                -> cba_xeh_fnc_test_prep
                    -> cba_xeh_fnc_test_prep
            -> cba_xeh_fnc_test_prep
                -> cba_xeh_fnc_test_prep
                    -> cba_xeh_fnc_test_prep
```

As you can see the formatting enables you to see which function called which and it's not just a simple list (which is especially useful with recursive functions like this one).

### DIFF
--- a/addons/main/script_macros_common.hpp
+++ b/addons/main/script_macros_common.hpp
@@ -871,13 +871,28 @@ Author:
 
 //#define PREP(var1) PREP_SYS(PREFIX,COMPONENT_F,var1)
 
-#ifdef DISABLE_COMPILE_CACHE
-    #define PREP(var1) TRIPLES(ADDON,fnc,var1) = compile preProcessFileLineNumbers 'PATHTO_SYS(PREFIX,COMPONENT_F,DOUBLES(fnc,var1))'
-    #define PREPMAIN(var1) TRIPLES(PREFIX,fnc,var1) = compile preProcessFileLineNumbers 'PATHTO_SYS(PREFIX,COMPONENT_F,DOUBLES(fnc,var1))'
+#ifndef DISABLE_COMPILE_CACHE
+    #define ENABLE_CACHING true
 #else
-    #define PREP(var1) ['PATHTO_SYS(PREFIX,COMPONENT_F,DOUBLES(fnc,var1))', 'TRIPLES(ADDON,fnc,var1)'] call SLX_XEH_COMPILE_NEW
-    #define PREPMAIN(var1) ['PATHTO_SYS(PREFIX,COMPONENT_F,DOUBLES(fnc,var1))', 'TRIPLES(PREFIX,fnc,var1)'] call SLX_XEH_COMPILE_NEW
+    #define ENABLE_CACHING false
 #endif
+
+#ifdef DEBUG_MODE_FULL
+    #define ENABLE_CALLSTACK true
+#else
+    #define ENABLE_CALLSTACK false
+#endif
+
+#define PREP(var1) ['PATHTO_SYS(PREFIX,COMPONENT_F,DOUBLES(fnc,var1))', 'TRIPLES(ADDON,fnc,var1)', ENABLE_CACHING, ENABLE_CALLSTACK] call CBA_fnc_compileFunction
+#define PREPMAIN(var1) ['PATHTO_SYS(PREFIX,COMPONENT_F,DOUBLES(fnc,var1))', 'TRIPLES(PREFIX,fnc,var1)', ENABLE_CACHING, ENABLE_CALLSTACK] call CBA_fnc_compileFunction
+
+// Be aware that the RETURN macro only works with functions compiled with CBA_fnc_compileFunction!
+#define RETURN breakOut "CBA_ROOT_SCOPE"
+#define RETURN(var1) if (isNil {var1}) then { \
+        breakOut "CBA_ROOT_SCOPE" \
+    } else { \
+        (var1) breakOut "CBA_ROOT_SCOPE" \
+    }
 
 #ifdef RECOMPILE
     #undef RECOMPILE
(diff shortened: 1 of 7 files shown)


addons/main/script_macros_common.hpp
```diff
@@ -871,14 +871,29 @@ Macro: GVARMAIN()
...
+#else
+    #define ENABLE_CALLSTACK false
+#endif
+
+#define PREP(var1) ['PATHTO_SYS(PREFIX,COMPONENT_F,DOUBLES(fnc,var1))', 'TRIPLES(ADDON,fnc,var1)', ENABLE_CACHING, ENABLE_CALLSTACK] call CBA_fnc_compileFunction
+#define PREPMAIN(var1) ['PATHTO_SYS(PREFIX,COMPONENT_F,DOUBLES(fnc,var1))', 'TRIPLES(PREFIX,fnc,var1)', ENABLE_CACHING, ENABLE_CALLSTACK] call CBA_fnc_compileFunction
+
+// Be aware that the RETURN macro only works with functions compiled with CBA_fnc_compileFunction!
+#define RETURN breakOut "CBA_ROOT_SCOPE"
+#define RETURN(var1) if (isNil {var1}) then { \
```
PabstMirror: #define altered without a #undef

addons/main/script_macros_common.hpp
```diff
@@ -871,14 +871,29 @@ Macro: GVARMAIN()
...
-    #define PREPMAIN(var1) TRIPLES(PREFIX,fnc,var1) = compile preProcessFileLineNumbers 'PATHTO_SYS(PREFIX,COMPONENT_F,DOUBLES(fnc,var1))'
+#ifndef DISABLE_COMPILE_CACHE
+    #define ENABLE_CACHING true
 #else
-    #define PREP(var1) ['PATHTO_SYS(PREFIX,COMPONENT_F,DOUBLES(fnc,var1))', 'TRIPLES(ADDON,fnc,var1)'] call SLX_XEH_COMPILE_NEW
-    #define PREPMAIN(var1) ['PATHTO_SYS(PREFIX,COMPONENT_F,DOUBLES(fnc,var1))', 'TRIPLES(PREFIX,fnc,var1)'] call SLX_XEH_COMPILE_NEW
+    #define ENABLE_CACHING false
 #endif
 
+#ifdef DEBUG_MODE_FULL
```
PabstMirror: What about allowing callstack without debug mode full?
```
#ifndef ENABLE_CALLSTACK
    #ifdef DEBUG_MODE_FULL
        #define ENABLE_CALLSTACK true
    #else
        #define ENABLE_CALLSTACK false
    #endif
#endif
```

addons/main/script_macros_common.hpp
BaerMitUmlaut: The preprocessor allows overloading macros.

commy2: But certains tools do no longer. And these macros do apply to config.

PabstMirror:
I don't really like adding the header to everything by default
This adds overhead even when not debugging
e.g. ace_frag_fnc_addBlackList goes from 2 lines to 
```
{
    scopeName "CBA_ROOT_SCOPE";
    private _fnc_scriptNameParent = if (isNil "_fnc_scriptName") then {"ace_frag_fnc_addBlackList"} else {_fnc_scriptName};
    private _fnc_scriptName = "ace_frag_fnc_addBlackList";
    scriptName _fnc_scriptName;

    params ["_round"];
    ace_frag_blackList set [(count ace_frag_blackList), _round];
}
```

addons/main/script_macros_common.hpp
BaerMitUmlaut: Certain tools need to die...

addons/xeh/script_header.hpp
```diff
@@ -0,0 +1,4 @@
+scopeName "CBA_ROOT_SCOPE";
```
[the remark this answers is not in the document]
dedmen: As said on Slack.
turn
`scopeName "CBA_ROOT_SCOPE";` into
```
scopeName "CBA_ROOT_SCOPE";
if (true) then {
```
and add a `};` to the end of the function inside `compileFunction`
This forces the function into a new scope thus getting rid of the scopeName error.

addons/main/script_macros_common.hpp
```diff
@@ -871,14 +871,29 @@ Macro: GVARMAIN()
...
+    #define ENABLE_CALLSTACK true
```
dedmen: Conflicts with https://github.com/acemod/ACE3/blob/master/addons/main/script_debug.hpp#L25

addons/xeh/script_header.hpp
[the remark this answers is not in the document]
commy2: `if (true) then {`
->
`call {`

PR #540, opened 2016-11-02
Move CBA settings into hidden category

BaerMitUmlaut:
- Move CBA settings into their own hidden category

A bit cleaner imo. Doesnt cause any problems with BWC either.


commy2:
lgtm. will test later.

Killswitch00:
Waiting for commy2 to test and OK this.

PR #542, opened 2016-11-06
Add fnc_insert

654wak654:
- Title

### DIFF
--- a/addons/arrays/fnc_insert.sqf
+++ b/addons/arrays/fnc_insert.sqf
@@ -16,8 +16,8 @@ Returns:
 
 Examples:
     (begin example)
-        _arr = ["a", "b", "d", "e"];
-        [_arr, 2, "c"] call CBA_fnc_insert;
+        _arr = ["a", "b", "e"];
+        [_arr, 2, ["c", "d"]] call CBA_fnc_insert;
         // _arr is now ["a", "b", "c", "d", "e"]
     (end)
 
--- a/addons/arrays/fnc_insert.sqf
+++ b/addons/arrays/fnc_insert.sqf
@@ -12,7 +12,7 @@ Parameters:
     _element - Element to be inserted <ANY>
 
 Returns:
-    The modified array. Empty array in case of errors <ARRAY>
+    The modified array <ARRAY>
 
 Examples:
     (begin example)
@@ -29,11 +29,8 @@ SCRIPT(insert);
 
 params ["_array", "_index", "_element"];
 
-private _size = count _array;
-if (_index >= _size || {_index < 0}) exitWith {[]};
-
-_right = _array select [_index, _size];
-_array deleteRange [_index, _size];
+private _right = _array select [_index, count _array];
+_array resize _index;
 _array pushBack _element;
 _array append _right;
 _array
--- a/addons/arrays/fnc_insert.sqf
+++ b/addons/arrays/fnc_insert.sqf
@@ -1,0 +1,39 @@
+/* ----------------------------------------------------------------------------
+Function: CBA_fnc_insert
+
+Description:
+    Inserts an element to the specified index of given array.
+
+    Modifies the original array (Thanks BaerMitUmlaut)
+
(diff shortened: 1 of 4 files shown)


BaerMitUmlaut:
You need to add it to CfgFunctions too.

654wak654:
:picard:^2

addons/arrays/fnc_insert.sqf
```diff
@@ -0,0 +1,39 @@
...
+#include "script_component.hpp"
+SCRIPT(insert);
+
+params ["_array", "_index", "_element"];
+
+private _size = count _array;
+if (_index >= _size || {_index < 0}) exitWith {[]};
+
+_right = _array select [_index, _size];
+_array deleteRange [_index, _size];
```
commy2: `_array resize [_index - 1];`

addons/arrays/fnc_insert.sqf
```diff
@@ -0,0 +1,39 @@
...
+---------------------------------------------------------------------------- */
+#include "script_component.hpp"
+SCRIPT(insert);
+
+params ["_array", "_index", "_element"];
+
+private _size = count _array;
+if (_index >= _size || {_index < 0}) exitWith {[]};
+
+_right = _array select [_index, _size];
```
commy2: missing `private`

addons/arrays/fnc_insert.sqf
```diff
@@ -0,0 +1,39 @@
...
+Author:
+    654wak654
+---------------------------------------------------------------------------- */
+#include "script_component.hpp"
+SCRIPT(insert);
+
+params ["_array", "_index", "_element"];
+
+private _size = count _array;
+if (_index >= _size || {_index < 0}) exitWith {[]};
```
commy2: `[[], 1, "test"] call CBA_fnc_insert`
should report
`[nil, "test"]`
imo

654wak654: So just `exitWith {}` or `exitWith {nil}`?

commy2:
If we'd replace `_element` with an array to insert, all we would have to change is one `pushBack` to `append`.
If you wanted to insert a single element, instead of
`[_arr, 2, "c"] call CBA_fnc_insert;`
you would write:
`[_arr, 2, ["c"]] call CBA_fnc_insert;`
, but you would still have the option to insert multiple elements at once.

commy2:
There exists `BIS_fnc_arrayInsert`

```sqf
/************************************************************
	Array Insert
	Author: Andrew Barron, optimised by Killzone_Kid

Parameters: [base array, insert array, index]
Returs: [array]

Inserts the elements of one array into another, at a specified
index.

Neither arrays are touched by reference, a new array is returned.

Example1: [[0,1,2,3,4], ["a","b","c"], 1] call BIS_fnc_arrayInsert
Returns: [0,"a","b","c",1,2,3,4]
Example2: [[1,2],[3,4],5] call BIS_fnc_arrayInsert
Returns: [1,2,<null>,<null>,<null>,3,4]
************************************************************/

/// --- validate general input
#include "..\paramsCheck.inc"
#define arr [[],[],0]
paramsCheck(_this,isEqualTypeParams,arr)

params ["_arr1", "_arr2", "_index"];

if (_index < 0) exitWith {_arr1};

private _size1 = count _arr1;
if (_index >= _size1) then {
	_arr1 = +_arr1;
	_arr1 resize _index;
};

(_arr1 select [0, _index]) + _arr2 + (_arr1 select [_index, _size1])
```

Differences:
- has input validation
- creates new array instead of modifying it
- slightly different order of parameters

We now have to think about:
- how does our variant perform compared to the (KK rewritten) BI function?
- do we need this CBA function (usefull feature vs. bloat?)

commy2:
@654wak654 Sorry for the delay. Can you add a unit test file for this? Would've to be added here:
https://github.com/CBATeam/CBA_A3/blob/master/addons/arrays/test.sqf

PR #543, opened 2016-11-06
Optimize Functions by Using 1.55 Select Syntax

654wak654:
- Make `CBA_fnc_select` and `CBA_fnc_reject` smaller by completely replacing them with the identical syntax of the select command.

### DIFF
--- a/addons/arrays/fnc_reject.sqf
+++ b/addons/arrays/fnc_reject.sqf
@@ -5,11 +5,11 @@ Description:
     Reject array elements for which the block returns true
 
 Parameters:
-    _array - Input Array [Array]
-    _block - Block [Code]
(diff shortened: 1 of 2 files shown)


commy2:
This means that you no longer can use `_this` and more importantly `_forEachIndex` with these functions. I think we should just deprecate them and leave them as they are.

654wak654:
You still have `_this`, didn't consider `_forEachIndex`.

commy2:
Yes you can. But why would you? Idk. I still think:
>we should just deprecate them and leave them as they are.

654wak654:
You should, doubt many addons are using it anyway.

654wak654:
Remember to deprecate 'em.

PR #544, opened 2016-11-08
JPN translation for v3.1.1

classicarma:
this translation is based on v3.1.1

### DIFF
--- a/addons/ai/stringtable.xml
+++ b/addons/ai/stringtable.xml
@@ -3,6 +3,7 @@
     <Package name="AI">
         <Key ID="STR_CBA_AI_Component">
             <English>Community Base Addons - Artificial Intelligence</English>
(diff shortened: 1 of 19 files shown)


commy2:
thank you

anyone knows what this means?
```
-<?xml version="1.0" encoding="utf-8"?>
```

bux578:
@commy2 That's probably just a different line ending, if I'm not mistaken.

jonpas:
@bux578 is probably correct. They should all be normalized though into LF, I'd double check entire file is in LF ending before merging.

classicarma:
@commy2 I don't know is this mean, but I saw red mark at the head of <?xml version="1.0" encoding="utf-8"?> in editor screen. I was do delete it.

Killswitch00:
Is this still up to date, now that 1.66 is out?

PR #548, opened 2016-11-16
Use new entities syntax to improve xeh speed

dedmen:
new 0.0429 ms 
`private _entities = entities [[], [], true, true]`
old 3.02719 ms
`_entities = entities "" + allUnits;_entities arrayIntersect _entities;`

I'm bored u know.
Was commy's Idea and I made a PR before he could.


PabstMirror:
Requires Arma 1.66

PR #553, opened 2016-11-27
Update fnc_initDisplay.sqf

AGPHeaddikilla:
**Changed a single syntax error**
Now the inventory (lbData)-data will be loaded properly if CBA/xeh is used.

### DIFF
--- a/addons/xeh/fnc_initDisplay.sqf
+++ b/addons/xeh/fnc_initDisplay.sqf
@@ -1,7 +1,7 @@
 #include "script_component.hpp"
 
 call {
-    #include "\a3\functions_f\GUI\fn_initDisplay.sqf";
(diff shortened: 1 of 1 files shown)


commy2:
There are no semi colon after macro instructions.

AGPHeaddikilla:
@commy2 What do you mean?

commy2:
In SQF, statements are seperated with `;`, but the preprocessor instructions terminate on newlines. It's like SQS in that case and there you don't put semi-colons at the end of the line.

AGPHeaddikilla:
Yeah, i know. I was just irritated.

commy2:
I don't understand what the initial problem is. Please elaborate on that. I don't see any list boxes here.

jonpas:
Additionally, in general pre-processor directives never use semi-colons.

Issue #513, opened 2016-09-30
MRT Accessory Functions


madpat3:
Ever thought about integrating mrt accessory functions?

for you already integrated JR.
MRT would take it to the next level, from my point of view.

dixon13:
Would need to probably get approval from the author of MRT Accessory Functions and then he or someone else would need to integrate it.

madpat3:
Robalo is one of the authors of that mod, or at least he wrote some configs for it. Maybe worth a discussion with him?

robalo:
I already asked da12thMonkey, the main author of MRT Accfncs. He allowed me to integrate it and further develop as needed into CBA.

sargken:
I am willing to help if you need it @robalo

commy2:
PR #555 is merged.

PR #555, opened 2016-11-27
Add MRT Accessory Functions #513

robalo:
- Integrate MRT Accessory Functions https://forums.bistudio.com/topic/171270-mrt-accessory-functions/
- Ref. #513


jonpas:
I feel like this could use a better component name.

robalo:
@jonpas me too, feel free to suggest a better name for it

commy2:
should use:
```sqf
private _unit = call CBA_fnc_currentUnit;
```
and then `_unit` instead of `player`, so it works when remote controlling units with Zeus.

Killswitch00:
Yeah, about that "accfcns" name.... How about just "accessory" (1 char longer, but more readable)?

robalo:
I like it.

PR #560, opened 2016-11-30
Add Timeout CBA_fnc_waitUntilAndExecute

654wak654:
- Title
- Add `_timeout` and `_timeoutCode` params, put calls that use them in a new `GVAR(waitUntilAndExecTimeoutArray)` array to not add any overhead to the usual `GVAR(waitUntilAndExecArray)`.

Separating the calls with timeout and without timeout into different arrays seemed like a better option then to add an overhead to everything that currently uses CBA's main PFH.

### DIFF
--- a/addons/common/fnc_waitUntilAndExecute.sqf
+++ b/addons/common/fnc_waitUntilAndExecute.sqf
@@ -3,7 +3,7 @@ Function: CBA_fnc_waitUntilAndExecute
 
 Description:
     Executes a code once in unscheduled environment after a condition is true.
-    It's also possible to add a timeout, in which case a difference code is executed.
+    It's also possible to add a timeout, in which case a different code is executed.
 
 Parameters:
     _condition   - The function to evaluate as condition. <CODE>
--- a/addons/common/fnc_waitUntilAndExecute.sqf
+++ b/addons/common/fnc_waitUntilAndExecute.sqf
@@ -22,18 +22,41 @@ Examples:
     (begin example)
         [{(_this select 0) == vehicle (_this select 0)}, {(_this select 0) setDamage 1;}, [player]] call CBA_fnc_waitUntilAndExecute;
     (end)
+    (begin example)
+        [{backpackCargo _this isEqualTo []}, {
+            deleteVehicle _this;
+        }, _holder, 5, {hint backpackCargo _this;}] call CBA_fnc_waitUntilAndExecute;
+    (end)
 
 Author:
     joko // Jonas, donated from ACE3
 ---------------------------------------------------------------------------- */
 #include "script_component.hpp"
 
-params [["_condition", {}, [{}]], ["_statement", {}, [{}]], ["_args", []], ["_timeout", 1E9, [0]], ["_timeoutCode", {}, [{}]]];
+params [
+    ["_condition", {}, [{}]],
+    ["_statement", {}, [{}]],
+    ["_args", []],
+    ["_timeout", 0, [0]],
+    ["_timeoutCode", {}, [{}]]
+];
 
-if (_timeout == 1E9) then {
+if (_timeout == 0) then {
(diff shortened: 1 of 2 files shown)


addons/common/init_perFrameHandler.sqf
```diff
@@ -80,6 +80,22 @@ GVAR(waitUntilAndExecArray) = [];
     if (_delete) then {
         GVAR(waitUntilAndExecArray) = GVAR(waitUntilAndExecArray) - [objNull];
     };
+
+    // Execute the waitUntilAndExecTimeout functions:
+    _delete = false;
```
commy2: could move this line after L81, so the flag is only reset if it ever where set. I think the same can be done above that once too.

addons/common/fnc_waitUntilAndExecute.sqf
```diff
@@ -25,8 +28,12 @@ Author:
 ---------------------------------------------------------------------------- */
 #include "script_component.hpp"
 
-params [["_condition", {}, [{}]], ["_statement", {}, [{}]], ["_args", []]];
+params [["_condition", {}, [{}]], ["_statement", {}, [{}]], ["_args", []], ["_timeout", 1E9, [0]], ["_timeoutCode", {}, [{}]]];
 
-GVAR(waitUntilAndExecArray) pushBack [_condition, _statement, _args];
+if (_timeout == 1E9) then {
```
commy2: maybe use `isNil "_timeout"` and have no default value / default value `nil` in `params`?
No one will ever use `1E9`, but having a random high number seems weird. (Especially an integer above single point floating precision.)

addons/common/fnc_waitUntilAndExecute.sqf
```diff
@@ -25,8 +28,12 @@ Author:
...
+    GVAR(waitUntilAndExecArray) pushBack [_condition, _statement, _args];
+} else {
+    GVAR(waitUntilAndExecTimeoutArray) pushBack [_condition, _statement, _args, _timeout, _timeoutCode, time];
```
PabstMirror: I think we want `CBA_missionTime` instead of `time`

addons/common/fnc_waitUntilAndExecute.sqf
PabstMirror: How about default timeout value of -1,
and only use time if `_timeout >= 0`

PabstMirror:
Using this wrapper, we can use the existing `waitUntilAndExecArray`, and make no changes to init_perFrameHandler.sqf.  So there would be no per-frame overhead.

```
if (_timeout == 1E9) then {
    GVAR(waitUntilAndExecArray) pushBack [_condition, _statement, _args];
} else {
    GVAR(waitUntilAndExecArray) pushBack [{
        params ["_condition", "_statement", "_args", "_timeout", "_timeoutCode"];

        if (CBA_missionTime > _timeout) exitWith {
            systemChat "timeout";
            _args call _timeoutCode;
            true
        };
        if (_args call _condition) exitWith {
            systemChat "condition";
            _args call _statement;
            true
        };
        false
    }, {}, [_condition, _statement, _args, _timeout, _timeoutCode, CBA_missionTime]];
};
```

addons/common/fnc_waitUntilAndExecute.sqf
654wak654: Could make default `0` and go for `_timeout > 0`, since 0 whould mean there is an immediate timeout.

addons/common/fnc_waitUntilAndExecute.sqf
654wak654: I used time because it's to measure relative time on one machine, but I can change it.

654wak654:
The timeout condition is `CBA_missionTime - _startTime > _timeout`, but other than that your code works great Pabst. Also change default timeout to 0 and did that """optimization""" with `_delete`.

commy2:
lgtm

PR #568, opened 2016-12-22
New package called "modFunctions"

AGPHeaddikilla:
So. I dont know why, but recently i needed this feature.
This package will simply parse and compile functions declared in the (new) config CfgModFunctions.cpp.
Benefits of this idea:
 - the possibility to select either compile or compileFinal (which is good for a development state of a package)
 - the opportunity to add more parameters to the config-compiling

It would be possible that im wrong that the BIS cfgFunctions do not accept compile/compileFinal states. But for now, i did not find a feature like this available.


jonpas:
Isn't that exactly what `PREP` does?

Dorbedo:
If i remember right, the CfgFunctions does support recompiling.
They even have a function, where you can recompile all functions on the fly: [BIS_fnc_recompile](https://community.bistudio.com/wiki/BIS_fnc_recompile)

The CfgFunctions also supports recompiling via Function attribute: [possible attributes](https://community.bistudio.com/wiki/Functions_Library_(Arma_3)#Attributes)

Usually for all functions: ```comileFinale``` is used. Only in the editor and if you set one parameter, this is switched into ```compile```.

AGPHeaddikilla:
@Dorbedo The Functions BIS_fnc_recompile recompiles (i think so) all available functions in CfgFunctions

Dorbedo:
It should be: [Wiki](https://community.bistudio.com/wiki/Functions_Library_(Arma_3)#Recompiling)

Unfortunately I can't test this these days. So i have to rely on this wiki. Maybe someone else can confirm it.

AGPHeaddikilla:
@Dorbedo but the thing is, that you cant change the code parts, if the function is compiled with compileFinal. So the idea of ("" call BIS_fnc_recompile) do unfortunately not work.

Dorbedo:
You have to set ```allowFunctionsRecompile = 1;``` via description.ext
This allows the recompiling. See ```initfunctions.sqf```(Version 3.0) Line 295

Dorbedo:
Nevertheless:
Your Idea, to add recompiling is a good one. I think this feature is missing. I could imagine that a recompile of separate functions or even whole components.
e.g.:
```
[] call CBA_fnc_recompile; // recompiles all functions
["myMod"] call CBA_fnc_recompile; // recompiles all functions with the specified PREFIX
["myMod","myCrazyComponent"] call CBA_fnc_recompile;  // recompiles component
["myMod","myCrazyComponent","greatFunction"] call CBA_fnc_recompile; // recompiles single function
```

commy2:
> Sure it does but it is (i think) in a better notation than the macro PREP() (and more easy to use ;) ).
Oh god no! `PREP()` is actually an actually usable method, because you can pick your individual components you want to recompile every mission. You also don't have to rely on config changes, so you can switch which component should be recompiled every time!
`CfgFunctions` recompiling takes for ever with a decent amount of functions like ACE (~20 seconds for me).
With the individual component recompile support `PREP()` offers it's an unnoticable delay.

Issue #551, opened 2016-11-21
CBA_fnc_taskDefend AI "Dance"


50KAL0V:
**Arma 3 Version:** `1.64` (stable)
**CBA Version:** `3.1.2` (stable)

**Mods:**
- `@CBA_A3`

**Description:**
CBA_fnc_taskDefend causing AI to "dance" when fortified in buildings.

https://www.youtube.com/watch?v=H5dgnWkk8Fw&

**Steps to reproduce:**
using "[this] call CBA_fnc_taskDefend" in group leader init or using the CBA modules

**Where did the issue occur?**
Happens on SP, Multi and dedicated server. New missions, old missions etc.

**RPT log file:**
http://pastebin.com/feStzbQb - RPT

SilentSpike:
So, using the defend zues module in ACE (which calls this function) I tested this code out in our unit's weekly game last night. Could not reproducing the dancing effect, it must be due to some more specific circumstances.

50KAL0V:
As per the description this is just using the CBA mod, nothing to do with ACE zeus modules. Issue occurs when using a module synced to the group leader or putting the call in the group leaders init.

I can provide a test mission if required.

commy2:
Please do that.

SilentSpike:
Would be helpful (and to clarify, the ace zeus module was just a mechanism of calling the taskDefend function on a group leader during actual gameplay). I can't reproduce it in the editor either as you describe and figured trying it under actual game conditions was worth a shot.

50KAL0V:
Understood, I just know the first thing i would get asked if i was testing with ACE would be "what happens when testing without ACE" =]

Just to clarify even on the example mission, this seems to take place about 70% of the time.

I'm pretty sure some others manged to replicate this straight away with the same tests when i was chatting on the ACE Slack a couple weeks ago. 

PSB

https://dl.dropboxusercontent.com/u/68881349/cbaDEFEND.Stratis.zip

SilentSpike:
Managed to narrow this down, it only seems to happen when they're in combat mode. Hopefully this isn't just an arma bug of `disableAI "PATH"` otherwise we might have to switch back to using `doStop`.

I'll do some testing with removal of the `setUnitPos` and stuff to see if that fixes it.

SilentSpike:
Yeah so after some testing, I think this is the result of the combination of `disableAI "PATH"` and `setUnitPos "UP"`.

[...]

PR #570, opened 2016-12-23
Fix dancing AI (#551)

SilentSpike:
~~Some sort of conflict between `disableAI "PATH"` and `setUnitPos "UP"`~~
A bug in `setUnitPos "UP"` causes AI to dance when engaged (repeatedly trying to lower their profile but can't I assume)

### DIFF
--- a/addons/ai/fnc_taskDefend.sqf
+++ b/addons/ai/fnc_taskDefend.sqf
@@ -4,7 +4,7 @@ Function: CBA_fnc_taskDefend
 Description:
     A function for a group to defend a parsed location.
 
-    Groups will mount nearby static machine guns and bunker in nearby buildings.
+    Units will mount nearby static machine guns and bunker in nearby buildings.
     They may also patrol the radius unless otherwise specified.
 
 Parameters:
@@ -29,7 +29,13 @@ Author:
 
 ---------------------------------------------------------------------------- */
 
-params ["_group", ["_position",[]], ["_radius",50,[0]], ["_threshold",2,[0]], ["_patrol",true,[true]]];
+params [
+    ["_group",grpNull,[grpNull,objNull]],
+    ["_position",[],[[],objNull,grpNull,locationNull],3],
+    ["_radius",50,[0]],
+    ["_threshold",2,[0]],
+    ["_patrol",true,[true]]
+];
 
 _group = _group call CBA_fnc_getGroup;
 if !(local _group) exitWith {}; // Don't create waypoints on each machine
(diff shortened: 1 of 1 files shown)


commy2:
Honestly, I liked `doStop` more. It didn't have permanent effects, so the AI could be given other waypoints / orders after they garrisoned a building.

SilentSpike:
I was under the impression `doStop` did have permanent effects unless you executed specific script commands on the AI afterwards? (just like `disableAI "PATH"` but much more hacky).

SilentSpike:
Decided to look into `doStop` for you, it actually causes the same dancing bug when used with `setUnitPos` :cry:

commy2:
Hmm. Only alternative is https://community.bistudio.com/wiki/setUnitPosWeak, but that will be overruled by the AI FSM, so there is not much point in using it.

SilentSpike:
It could work because they'd be detached from their formation with `doStop`, however if it did work I have a feeling it would cause the same problem

ShizCalev:
Not sure if this is a CBA issue or an ACE one. With this change, when issuing the defend area command to units, once the patrolling units hit their cycle waypoint, all the units whom were in buildings abandon their posts and rejoin formation to take up patrol.

commy2:
I think that is the expected behavior though. If a waypoint is finished, the whole group should begin with the next waypoint.

ShizCalev:
Previous behavior was that issuing Defend Area would have units garrison buildings in the area, with a 7% chance of a unit going onto a patrol around the area. Once the patrol group finished through all waypoints and hit the cycle, they would repeat the same patrol with the same select units, and the rest would stay garrisoned (since pathing was disabled on the ones in the buildings.)

Now the behavior is that units garrison buildings, 7% chance at patrolling, however once the patrol group reaches the final waypoint and hit the cycle, all units leave their buildings and start patrolling the area in a formation.

commy2:
That wasn't the case before @SilentSpike changed the taskDefend module / waypoint in the first place.
We cannot use `disableAI "PATH"` as it's broken.
I think placing waypoints after any of these taskX waypoints is undefined behavior.

ShizCalev:
May I ask how `disableAI "PATH"` is broken? I know `setUnitPos "UP"` is, however this is the only I've heard about disableAI (and there are a couple other mods that I use that that information would be relevant for.)

I tired out replacing `doStop _unit;` with `disableAI "PATH"` and the AI seem to stay in the buildings as intended, and shoot freely as desired.

commy2:
Yeah, my bad. `setUnitPos` is broken. However, units being stuck after a taskDefend waypoint is not intentional either.

PR #573, opened 2017-01-10
Add Garrison/Patrol ratio parameter

50KAL0V:
- Slight increase to the default, but this way users can tweak to allow for more of a "Defence" rather then just another garrison style script.

- Added the new parameter, description and edited text above its use.

### DIFF
--- a/addons/ai/fnc_taskDefend.sqf
+++ b/addons/ai/fnc_taskDefend.sqf
@@ -15,6 +15,7 @@ Optional:
     - Defend Radius (Scalar)
     - Building Size Threshold (Integer, default 2)
     - Can patrol (boolean)
+    - Ratio of Garrison/Patroling units. Higher means more garrison (Float, default 0.8)
 
 Example:
     (begin example)
@@ -34,7 +35,8 @@ params [
     ["_position",[],[[],objNull,grpNull,locationNull],3],
     ["_radius",50,[0]],
     ["_threshold",2,[0]],
-    ["_patrol",true,[true]]
+    ["_patrol",true,[true]],
+    ["_garrisonratio",0.8,[0]]
 ];
 
 _group = _group call CBA_fnc_getGroup;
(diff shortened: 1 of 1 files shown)


commy2:
Should probably do the following as well:

- Add this as parameter to the module:
https://github.com/CBATeam/CBA_A3/blob/master/addons/modules/CfgVehicles.hpp#L102
https://github.com/CBATeam/CBA_A3/blob/master/addons/modules/fnc_moduleDefend.sqf

- Change the default for the waypoint version to 80% too:
https://github.com/CBATeam/CBA_A3/blob/master/addons/ai/fnc_waypointGarrison.sqf#L34

SilentSpike:
I feel like this is slightly misleading, since it's not actually a ratio (0.8 != 80% of units will garrison).

SilentSpike:
It would probably be better to make the `_patrol` parameter accept either boolean or a percentage chance value for an implementation such as this.

commy2:
>it's not actually a ratio (0.8 != 80% of units will garrison).

It is for large amount of units.

SilentSpike:
On average true (assuming `random` is uniform distribution which I've never checked), but you could in practise get the exact opposite ratio than expected

commy2:
The function header should state that it's a probability per unit (if it is, not sure).

50KAL0V:
Before i break it all horribly not being much of a code and github guru

Whats the best way for me to submit those other changes? should i edit in the branch its created or create new pull requests?

Also regarding the modules, will the "typeName" be number?

commy2:
Edit your branch and push to your repository. It will retrieve the new changes here automatically.

commy2:
>Also regarding the modules, will the "typeName" be number? Or still called a Float???
Probably "NUMBER"

50KAL0V:
cheers, will get on that

commy2:
Status?

SilentSpike:
I had another idea along these lines recently: a parameter could be added for the percentage of units who will `disableAI "PATH"` (default 0%) that way people who want that behaviour can enable it too

Killswitch00:
This still needs the changes in other files that commy2 noted on Jan 13

SilentSpike:
I'd be happy to re-implement this if @50KAL0V has gone AWOL 😆

Killswitch00:
That'll work. Also, it was never explained why the default probability was lowered from 0.93 to 0.8.

Killswitch00:
Re-implemented in #702

PR #576, opened 2017-01-10
fix RETNIL and RETDEF throwing script errors in scheduled environment

### LINKED ISSUE #566: RETNIL and RETDEF throwing script errors
**CBA Version:** `3.1.2` (stable)

**Mods:**
- `@CBA_A3`

**Description:**

the new RETNIL and RETDEF macros are throwing errors if the value is not defined (only sheduled environment)

```
_apple = nil;
_juice = [_apple] param [0]; // throws error
```
old version:
```
_cherry = nil;
_juice = If (isNil{_cherry}) then {nil}else{_cherry}; // no error
```

commy2:
- title
- fix #566
- also adds unit test
```sqf
execVM "\x\cba\addons\common\test_ret.sqf"
```


Dorbedo:
Aren't the macros part of the main component?
If not, maybe the test_macro_is_x.sqf should be moved to common too.

commy2:
The problem is that the ruby script is set up in a way were `test_x` is not included. The test.sqf in main is only a hub to call the other test functions. `test_macro_is_x.sqf` is never used atm.
Tbh, I think it is easier to just `execVM` the individual tests, since `params` throws errors now and a bunch of other tests fail every time (sometimes due to a lacking `player` when tested in the 2d editor).

commy2:
There. Moved. No guarantee that the test still works though.

PR #577, opened 2017-01-17
Fix RETNIL macro for code

PabstMirror:
#576 broke TRACE macros such as 
`TRACE_2("",_unit,local _unit);` - due to `isNil "local _unit"` is true.
and
`TRACE_1("",QUOTE(ADDON));` - quotation problems

Using the `{}` version of `isNil` seems to work just as well and fixes these problems.


commy2:
BEEP. Reminder to myself to mute the machine when running it overnight.

This is right. This is how it was before we screwed it up by trying to use `param`. Thanks.

654wak654:
😢 whoops

PR #578, opened 2017-01-17
Fix entities returning duplicates

### LINKED ISSUE #567: XEH initPost runs twice when added retroactively via function
**CBA Version:** `3.1.2` (dev 311d8d7128f4f39679e984164736f34f9d42c6f3)

**Mods:**
- `\x\cba`
- `\z\ace`
- `@devcon_a3` (in-house dev console that I always run while developing)

**Description:**
- When using function `CBA_fnc_addClassEventHandler` to add an `"initPost"` XEH to a vehicle with the retroactive parameter set to true, the event will run twice for existing instances of that vehicle.

**Steps to reproduce:**
Run the following repro mission:
[EmptyMission.VR.zip](https://github.com/CBATeam/CBA_A3/files/666150/EmptyMission.VR.zip)

You can see I've added an initPost EH to `CAManBase` (player unit) both via the `CBA_fnc_addClassEventHandler` function and via the `Extended_InitPost_EventHandlers` in description.ext

The config version runs only once, while the scripted version runs twice.

- Editor (Singleplayer)

**RPT log file:**
- [Arma3_2016-12-21_11-23-18.txt](https://github.com/CBATeam/CBA_A3/files/666195/Arma3_2016-12-21_11-23-18.txt)

PabstMirror:
Fix #567

@commy2 
Not sure if there is a better way to do this.
`arrayIntersect` is expensive on very large arrays, but addClassEH is probably only going to be called at init.

### DIFF
--- a/addons/xeh/fnc_addClassEventHandler.sqf
+++ b/addons/xeh/fnc_addClassEventHandler.sqf
@@ -57,6 +57,9 @@ if (_applyInitRetroactively && {!(_eventName in ["init", "initpost"])}) then {
 // add events to already existing objects
 private _eventVarName = format [QGVAR(%1), _eventName];
 
+private _entities = (entities [[], [], true, true]);
+_entities = _entities arrayIntersect _entities; // entities can return duplicates at postInit - #567
+
 {
     if (_x isKindOf _className) then {
         private _unit = _x;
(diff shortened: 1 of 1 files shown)


commy2:
`arrayIntersect` is fine once on addClassEventHandler. We are doing this in the release version already with `entities + allUnits`, so it's still an improvement over all after the update.

PR #582, opened 2017-01-29
Add CBA_switchAttachment event for attachment switch

PabstMirror:
Adds a local event on attachment switch


commy2:
"CBA_attachmentSwitched" ?

PR #583, opened 2017-02-08
Expliclty add modifyer settings for Shift/Ctrl/Alt

PabstMirror:
Ref https://github.com/acemod/ACE3/issues/4822
ACE was adding a keybind of `[29, [false, false, false]]` // 29 = CTRL/STRG key
But CBA keyhandler expects the control param bool to also be true `[29, [false, true, false]]`

The same problem also exists in the code that handles rebinding. 
It would only add the base key with all modfiers set to false.

That code calls fnc_addKeybind, so we can solve both problems with this fix.

### DIFF
--- a/addons/keybinding/fnc_addKeybind.sqf
+++ b/addons/keybinding/fnc_addKeybind.sqf
@@ -71,6 +71,13 @@ if (count _defaultKeybind == 4) then {
     _defaultKeybind = [_defaultKeybind select 0, _modifiers];
 };
 
+// Make sure modifer is set to true, if base key is a modifier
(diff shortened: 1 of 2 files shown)


commy2:
I know the rest of CBA (or ACE) usually doesn't do it that way, but I think instead of using the magic numbers for the DIK codes, we should use the _enums_ from BI.

`#include "\a3\ui_f\hpp\defineDIKCodes.inc"` in this components `script_component.hpp` and then:
```
DIK_LSHIFT
DIK_RSHIFT
DIK_LCONTROL
DIK_RCONTROL
DIK_LMENU
DIK_RMENU
```

PR #585, opened 2017-02-09
randPos - Don't modify input array

### LINKED ISSUE #584: CBA_fnc_getPos returning array reference instead of an array-copy
CBA_fnc_getPos is returning the feeded array reference instead of an array-copy, 
which f.e. then bugs out CBA_fnc_RandPos which will alter the originally passed position instead of only returning a new one. The latter related to this commit, where the array-copy has been removed:

https://github.com/CBATeam/CBA_A3/commit/8a4d6c21d8fd700c0337dcf964bf087ba36a9fdc#diff-bee87607846a78814c6fe5729a46a521L24 - line 24

In our case (ALiVE), the usage of the function also causes positions beeing altered in our hashes data.

PabstMirror:
Fix #584

Looks like we removed the copy at one point because we switched to `vectorAdd`, 
but went back to `set` because it needs to work with 2d arrays.

### DIFF
--- a/addons/common/fnc_randPos.sqf
+++ b/addons/common/fnc_randPos.sqf
@@ -27,7 +27,7 @@ params [
     ["_radius", 0, [0]]
 ];
 
-private _position = _entity call CBA_fnc_getPos;
+private _position = +(_entity call CBA_fnc_getPos); // copy so we don't modify the input array
 
 _position set [0, (_position select 0) + (_radius - 2 * random _radius)];
 _position set [1, (_position select 1) + (_radius - 2 * random _radius)];


addons/common/fnc_randPos.sqf
```diff
@@ -27,7 +27,7 @@ params [
     ["_radius", 0, [0]]
 ];
 
-private _position = _entity call CBA_fnc_getPos;
+private _position = +(_entity call CBA_fnc_getPos); // copy so we don't modify the input array
 
 _position set [0, (_position select 0) + (_radius - 2 * random _radius)];
 _position set [1, (_position select 1) + (_radius - 2 * random _radius)];
```
commy2:
```
_position getPos [random _radius, random 360]
```
Both are AGL.
Btw., this does currently not place you inside a random circle, like _radius_ would suggest. You're actually placed inside a square, with the NW, NE, SW and SE edges being the extremes. It should be a circle, because it makes no sense that certain cardinal directions are preferred.

I also agree that `_position call CBA_fnc_getPos` should report a copy and not a pointer.

PabstMirror:
Not needed with PR #586

addons/common/fnc_randPos.sqf
PabstMirror: To get a uniform distribution:
`_position getPos [_radius * sqrt (random 1), random 360]`
See http://mathworld.wolfram.com/DiskPointPicking.html

PR #587, opened 2017-02-11
Fix random position calc. of CBA_fnc_randPos

commy2:
- title
- performance is pretty poor and idk why. Might be just the sum of everything.


SilentSpike:
:+1:

PR #588, opened 2017-02-13
pass short functions to addXEventHandler to reduce compile times

commy2:
- According to @dedmen 's findings, functions passed directly to any `addXEventHandler` are stored internally as strings and have to be compiled every time the event handler is executed.
- This takes longer than the additional `call` command, so it's worth to replace them with short `{call <FUNCTION>}` functions.
- I don't use `_this call`, because the unary syntax of `call` doesn't overwrite `_this` anyway, so it's the same result, but even faster while only being slightly less readable (arguably).

### DIFF
--- a/addons/common/XEH_postInit.sqf
+++ b/addons/common/XEH_postInit.sqf
@@ -2,7 +2,7 @@
 #include "script_component.hpp"
 
 //Install PFEH:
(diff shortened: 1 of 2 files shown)


dedmen:
For reference: https://feedback.bistudio.com/T123355

PR #595, opened 2017-02-17
Fix for bugged PlayerViewChanged event

PabstMirror:
`PlayerViewChanged` event does not always fire on dedicated server when entering vehicles.
Readd the vehicle change code to the eachFrame polling section.

@commy2 - I left vehicle checking in the PVC event;
no matter what it will only be called once.
So it should be fine??


commy2:
The one that checks _player is still missing. Or is that working correctly?

PabstMirror:
On dedicated with remote units
Zeus RC and `selectPlayer` both worked correctly

PR #597, opened 2017-02-22
Music functionality

fishykins:
When merged this Pull will add music based functions to CBA. 
-A very generic collection of functions, mainly used to help simplify finding, playing and monitoring music.
-Elaborates on the basic BIS config for music to add support for class "tags" and "type". 
While this may seem trivial, many mod and mission makers do rely on music in their works. CBA could help standardize the way this is done!


fishykins:
Absolute rookie here, but very keen to get it right. What are the indent rules, seeing as what I've done clearly isn't winning any medals! Thanks

commy2:
Soft tabs with 4 char length.

commy2:
ref: #598

PR #606, opened 2017-02-26
improve settings

### LINKED ISSUE #580: Global settings forced on server instead of in mission by default
**CBA Version:** `3.1.2` (dev `9b824dc1b497b2a549b28c4b7a016d5f692df7e0`)

**Mods:**
- `@CBA_A3`
- `@acre2`

**Description:**
- Global settings are forced on server by default instead of in mission (requiring a server admin to un-force the setting on server to allow mission maker to set it in mission).
- Should probably be forced in mission by default and only forced on server if set so.
- @PabstMirror pointed out a possible bug in https://github.com/CBATeam/CBA_A3/blob/d6ef65be9d78bf7989de1e5b69540eeb673cb1aa/addons/settings/fnc_isForced.sqf#L26

**Steps to reproduce:**
- Load some default global settings.
- See they are forced on server instead of in mission.

- Dedicated

**RPT log file:**
- Not applicable

Point of observation is "Force" column:

commy2:
- adds optional `cba_settings.sqf` file for missions and optional `userconfig` as well as template for a `userconfig` addon that doesn't require `filePatching`.
- reworked ui
- adds EDITBOX type setting (strings)
- adds checkbox to overwrite mission by server settings (no longer default), which:
- fixes #580
- bumps required version to 1.68
- probably fixes a bunch of smaller issues
- reworked [documentation](https://github.com/CBATeam/CBA_A3/wiki/CBA-Settings-System)

_____

notes:

`@todo` eventually:
- define in description.ext the location of the `cba_settings.sqf` in description.ext and if not found default to `MISSION_SETTINGS_FILE`


jonpas:
Notes from Slack:
> magnetar and I were also discussing mission settings, how you can currently only set it in Eden or manually editing .sqm (not even sure if you can do that)
with the same logic as reading userconfig or static settings pbo, it would be possible to have a file eg. `cba_settings.sqf` in root of mission, that could also be used to set the settings
with higher priority as Eden settings as it can be edited outside of the editor
not sure if it would be best idea to hardcode a file name, but I like that more than telling people how to call specific function (which has all sorts of problems due to stupid init delays people do and whatever)

nateberkopec:
> `cba_settings.sqf`

Yes please!

Dorbedo:
It would be great, if the cba-settings would support the settingtype "STRING" (edit-box).

commy2:
```
priority    CLIENT                  MISSION                     SERVER
0           default                 ignore this setting         ignore this setting
1           don't use suggestions   suggest this setting        suggest this setting, but mission suggestion would overwrite
2           N/A                     Overwrite client            Overwrite client, but not mission
3           N/A                     N/A                         Overwrite client and mission
```

commy2:
```
priority    CLIENT                  MISSION                     SERVER
1           default                 ignore this setting         ignore this setting
2           N/A                     Overwrite client            Overwrite client, but not mission
3           N/A                     N/A                         Overwrite client and mission
```

commy2:
@Dorbedo 
Is this good enough?

commy2:
ref: #580

commy2:
- Export disabled in MP
https://community.bistudio.com/wiki/copyToClipboard
- Info

Neviothr:
This looks **very** useful, thank you @commy2.

commy2:
WIP updated doc
https://gist.github.com/commy2/4db0602cda2f93e2b7ede61c1657b357

ViperMaul:
Help us test the release canditate:
[CBA v3.3.0 RC1](https://www.dropbox.com/s/evnxm5uqv239t9j/CBA_A3_v3.3.0-RC1a.zip?dl=0)

Release Notes:
https://github.com/CBATeam/CBA_A3/milestone/21?closed=1

PR #608, opened 2017-03-01
add CBA_fnc_allNamespaces

commy2:
- adds function to report all created pseudo namespaces

### DIFF
--- a/addons/common/fnc_allNamespaces.sqf
+++ b/addons/common/fnc_allNamespaces.sqf
@@ -8,7 +8,7 @@ Parameters:
     None
 
 Returns:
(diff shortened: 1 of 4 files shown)


addons/common/fnc_allNamespaces.sqf
```diff
@@ -0,0 +1,24 @@
...
+Function: CBA_fnc_allNamespaces
+
+Description:
+    Reports namespaces created with CBA_fnc_createNamespace.
+
+Parameters:
+    None
+
+Returns:
+    _namespace - all custom namespaces <ARRAY>
```
Killswitch00: Clarify what data types the elements in the returned array are

PR #609, opened 2017-03-01
add CBA_fnc_(de)serializeNamespace

commy2:
- title

I already use this idea for the settings stuff, but this is put inside functions. The one downside of the pseudo namespaces is, that LOCATION's and OBJECT's don't survive the missions end. CBA hashes are just arrays, so they do. By "serializing" you can essentially store a copy of the pseudo namespace in the `profileNamespace` or even in the mission (using `s`/`get3DENMissionAttributes`) by converting them to a CBA hash first.

### DIFF
--- a/addons/hashes/CfgFunctions.hpp
+++ b/addons/hashes/CfgFunctions.hpp
@@ -80,6 +80,8 @@ class CfgFunctions
                 description = "Parses a YAML file into a nested array/Hash structure.";
                 file = "\x\cba\addons\hashes\fnc_parseYAML.sqf";
             };
+            PATHTO_FNC(serializeNamespace);
+            PATHTO_FNC(deserializeNamespace);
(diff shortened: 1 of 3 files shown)


PabstMirror:
Not sure if it's a problem, but I think with the XEH fallback running there will be some extra setVars on the object namespace that we might want to filter out.

commy2:
Does that even run for CBA_namespaceDummy CfgVehicles?

commy2:
```sqf
allVariables (true call CBA_fnc_createNamespace)
[]
```

xiadaxiaogao:
What is Object CBA_namespacedummy???

xiadaxiaogao:
Cannot open object cba_namespacedummy. Configure addons menu is empty and ace option is empty too.

PR #612, opened 2017-03-04
Improve initbox

commy2:
- doubles the size of the initbox (10 lines instead of only 5)
- optionally removes code validation which enables using local variables and return values

Seems like the game actually can handle those just fine, but the code validation blocks you from doing that. ~~The only downside is having no code validation, but if there is an error in the scripting, it will be shown at mission start like any other script error. Worth it imo.~~ Optional now.


commy2:
As per suggestion, I made it optional by changing a checkbox.

PR #614, opened 2017-03-06
Sperate Settings File Load logic

jonpas:
- Title
- Shorten optional and template setting PBOs code
- Change `CBA_settings_fnc_init` to public function


addons/settings/fnc_readFile.sqf
```diff
@@ -0,0 +1,32 @@
+/* ----------------------------------------------------------------------------
+Function: CBA_settings_fnc_readFile
+
+Description:
+    Reads and applies settings from a file.
+    Only considered as public API from a static settings addon!
```
jonpas: Unsure how to make this clear. In theory it is public API because people use it in their own static settings PBOs, but we also don't want anyone to use outside of that I assume.

nateberkopec: Perhaps you should rename the func to something less generic, like `CBA_settings_fnc_readStaticSettingsFile`

commy2: Should stay an internal function. Loading the userconfig file could be some kind of ingame setting itself.

jonpas: Sure, I just tried to somehow make it it clear it is semi-public because every with a static setting addon would call it for example.

PR #617, opened 2017-03-10
Fix "unit" changed PlayerEH not working reliably

### LINKED ISSUE #615: Unit changed PlayerEH not called when leaving Zeus RemoteControl
**Arma 3 Version:** `1.66` (stable / rc / dev)
**CBA Version:** `3.2.1` (stable / dev + commit hash)

**Mods:**
```
- CBA_A3
- TFAR
```

**Description:**
When leaving a Zeus Remote Control by Pressing the `Z` button the [currentUnit changed Event](https://github.com/CBATeam/CBA_A3/blob/master/addons/events/fnc_addPlayerEventHandler.sqf#L173) is not executed. It only runs after the Zeus Interface was completly closed by Pressing Z again and going back into the Zeus Players eyeCamera

This is probably because the Arma PlayerViewChanged EH doesn't trigger when Zeus leaves Remote Control and goes back into Zeus Interface Camera

Reference: https://github.com/michail-nikolaev/task-force-arma-3-radio/issues/1211

**Steps to reproduce:**
- Use this:
```
["unit", {
    systemChat str ["playerChanged",_this];
},true] call CBA_fnc_addPlayerEventHandler;
```
- Go Into Zeus Interface and Remote Control a unit
- Press `Z` to go back into Zeus Interface
- Press `Z` again to close Zeus Interface
- You'll see that the Eventhandler is only called after you exit the Zeus Interface. Although after you quit remote Control and went back into Zeus Interface `CBA_fnc_currentUnit` report's player already.

- Dedicated / Self-Hosted Multiplayer

commy2:
- "unit"  and "vehicle" player eventhandlers seem to not work the same as the BI playerViewChanged after all, so switch back to the old polling method to avoid bugs.

### DIFF
--- a/addons/events/fnc_addPlayerEventHandler.sqf
+++ b/addons/events/fnc_addPlayerEventHandler.sqf
@@ -40,49 +40,49 @@ params [["_type", "", [""]], ["_function", {}, [{}]], ["_applyRetroactively", fa
 _type = toLower _type;
 
 private _id = switch (_type) do {
-case ("unit"): {
+case "unit": {
     if (_applyRetroactively && {!isNull (missionNamespace getVariable [QGVAR(oldUnit), objNull])}) then {
         [GVAR(oldUnit), objNull] call _function;
     };
     [QGVAR(unitEvent), _function] call CBA_fnc_addEventHandler // return id
 };
-case ("weapon"): {
+case "weapon": {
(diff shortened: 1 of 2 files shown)


PabstMirror:
Can confirm this is needed.  "unit" is giving problems with `selectPlayer` on dedicated

PR #619, opened 2017-03-17
improve performance of namespace

commy2:
- After 1.68 update, the performance cost of rendering local CBA namespaces is higher if they are `drawType = "name";` than they are with `drawType = "area";`, so we can change this to have a very slight performance boost with the map opened (becomes apparent with very large number of namespaces).

- Map - FPS no local namespaces: 62-65 FPS
- Map - 50k local namespaces before this PR: 14-15 FPS
- Map - 50k local namespaces after this PR: 50 FPS

Additionally I now ~disable the simulation of the global namespaces via `init` eventhandler~ use simple objects for the global namespaces. (possible since they are OBJECT and not LOCATION).

- No map, no namespaces: 80 ~72-75 FPS (65 on map)~
- 50k global namespaces before this PR: ~ 32 FPS (same on map)
- 50k global namespaces after this PR: 77 ~59 FPS (same on map)~

Local namespaces are _drawn_ (LOCATION) while the map is opened, while global namespaces are always _simulated_ (OBJECT). This explains why local namespaces have (almost) no effect on FPS when the map is closed, while they have a bigger one if the map is shown.

- Code used to create 50k local namespaces:
```sqf
for "_i" from 1 to 50000 do {
    private _newHash = call CBA_fnc_createNamespace;
    TEST pushBack _newHash;
};
```

- Code used to create 50k global namespaces:
```sqf
for "_i" from 1 to 50000 do {
    private _newHash = true call CBA_fnc_createNamespace;
    TEST pushBack _newHash;
};
```

Additionally, ACRE is using `setText` on the local namespace for no apparent reason. I suggest that they stop doing that, since that cuts the FPS in half with 50k local namespaces (14->7 before this PR, 50->24 after this PR).

### DIFF
--- a/addons/common/CfgLocationTypes.hpp
+++ b/addons/common/CfgLocationTypes.hpp
@@ -2,7 +2,7 @@
 class CfgLocationTypes {
     class CBA_NamespaceDummy {
         name = "";
-        drawStyle = "name";
+        drawStyle = "area";
         texture = "";
         color[] = {0,0,0,0};
         size = 0;
--- a/addons/common/CfgVehicles.hpp
+++ b/addons/common/CfgVehicles.hpp
@@ -4,9 +4,5 @@ class CfgVehicles {
     class CBA_NamespaceDummy: Static {
         scope = 1;
         displayName = "";
-
-        class EventHandlers {
-            init = "(_this select 0) enableSimulation false";
-        };
     };
 };
--- a/addons/common/CfgVehicles.hpp
+++ b/addons/common/CfgVehicles.hpp
@@ -4,5 +4,9 @@ class CfgVehicles {
     class CBA_NamespaceDummy: Static {
         scope = 1;
         displayName = "";
+
+        class EventHandlers {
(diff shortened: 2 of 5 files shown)


jonpas:
> Additionally, ACRE is using setText on the local namespace for no apparent reason. I suggest that they stop doing that, since that cuts the FPS in half with 50k local namespaces (14->7 before this PR, 50->24 after this PR).

Going to be in ACRE 2.4.1.

PabstMirror:
Would `createSimpleObject` work?

commy2:
80 -> 77, so yeah, it helps a lot. Not usable for local namespaces though, since simple objects are always transferred over network.

commy2:
Had to also adjust `CBA_fnc_allNamespaces`. `createSimpleObject` uses ASL, not AGL and does not support type filters. Since we cannot use a type filter anymore, I also changed the position to make it unlikely that another object is mistakenly reported by the function as namespace.

commy2:
`typeOf` still works, so we can use that to ensure it's a global namespace.

PR #622, opened 2017-03-20
addClassEH - Handle respawning units

PabstMirror:
Ref https://github.com/acemod/ACE3/issues/5005

Add class event handler fails if the unit is dead and in the process of respawning, 

To reproduce:
```
addMissionEventHandler ["EntityKilled", {
    ["CAManBase", "fired", {systemChat str _this}] call CBA_fnc_addClassEventHandler;
}];
```
Put in init.sqf and hit respawn in MP.

### DIFF
--- a/addons/xeh/fnc_addClassEventHandler.sqf
+++ b/addons/xeh/fnc_addClassEventHandler.sqf
@@ -57,7 +57,7 @@ if (_applyInitRetroactively && {!(_eventName in ["init", "initpost"])}) then {
 // add events to already existing objects
 private _eventVarName = format [QGVAR(%1), _eventName];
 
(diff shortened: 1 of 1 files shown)


commy2:
But if the unit respawns, it does not trigger the init eventhandler again. How does this fix anything?
And why does it fail in the first place?

commy2:
`respawnTime = 0;` or >0 ?

commy2:
Found the problem:
```
addMissionEventHandler ["EntityKilled", {
    params ["_corpse"]; 
    _entities = entities [[], [], true, true];
    hint str (_corpse in _entities); 
}]; 
```
Reports false.
Apprently the fourth argument, flag is `excludeDead`, not "`includeDead`". Oops.

PR #625, opened 2017-04-01
cleanup extended debug console

commy2:
- fixed PREV / NEXT function not being case sensitive (annoying/confusing when debugging case sensitive code)
- fixed using debug console for the first time would enable PREV button (very minor bug)
- fixed position in previous statements after closing and reopening the game (uiNamespace var had to be profileNamespace var)
- fixed pressing shift + enter (newline) would save expression as if enter was pressed
- cache function in ui namespace so pause menu opens slightly faster
- added some GVAR for global variables
- added missing `private`
- removed `disableSerialization` (not needed in unscheduled function)
- macros for UI grids
- macros for IDC's
- some simplifications, less global vars, descriptive variable names, removed unused code


commy2:
- added translations for the button and changed title to "Extended Debug Console"
- also record expression when code performance button is used

commy2:
- moves debug console down one grid, so it doesn't kiss the top edge of the screen, but uses free space at the bottom instead

Cuel:
I've noticed doing `["Open", true] call bis_fnc_arsenal` is not saved into current statements after exiting, however you can press `previous statement` and then `next statement` to get it to show correctly. Minor bug?

PR #628, opened 2017-04-06
delay initPost eventhandler of objects created mid mission

commy2:
This "solves" this issue:

```sqf
["B_MRAP_01_F", "init", {(_this select 0) setVariable ["test1", true, true]}] call CBA_fnc_addClassEventHandler;
_v = "B_MRAP_01_F" createVehicle position player;
_v setVariable ["test2", true, true];
missionNamespace setVariable ["v", _v, true];
```

>On local machine: test1 and test2 are defined
>On remote machine: test1 undefined, test2 defined
>-> setVariable public doesn't work inside the init eventhandler
>We have initPost, which solves this for editor placed objects (init artificially delayed until postInit eventhandler), but it fails for objects created mid mission, which is a pretty nasty inconsistency.
>I plan to fix this in the next CBA update by delaying initPost one frame artifically if postInit already happened.

### DIFF
--- a/addons/xeh/fnc_init.sqf
+++ b/addons/xeh/fnc_init.sqf
@@ -34,7 +34,7 @@ if !(ISINITIALIZED(_this)) then {
     // run InitPost or put on stack
     if (SLX_XEH_MACHINE select 8) then {
         {
-            [_this] call _x;
+            [_x, [_this]] call CBA_fnc_execNextFrame;
         } forEach (_this getVariable QGVAR(initPost));
     } else {
         GVAR(initPostStack) pushBack _this;


PabstMirror:
@commy2 - We got the following error a few times on a normal mission
```
  Error Reserved variable in expression
File z\ace\addons\interact_menu\functions\fnc_compileMenu.sqf, line 117
```
Can reproduce with 
```
x3 = "B_Heli_Transport_01_F" createVehicle [0,0,0];
deleteVehicle x3;
```

ACE throws an error because the object is null (so typeOf = "" and the setVar errors out)
Can easily fix in ace, but I'm wondering if we want a more generic solution in CBA with an isNull check?
Something like:
```
[{
    if (isNull _this) exitWith {};
    {
        [_this] call _x;
    } forEach (_this getVariable QGVAR(initPost));
}, _this] call CBA_fnc_execNextFrame;
```

commy2:
On the one hand, you expect the eventhandler to run when an object is created, on the other it's not very useful with objNull.
If this check is added, it should also be added here:
https://github.com/CBATeam/CBA_A3/blob/master/addons/xeh/fnc_postInit_unscheduled.sqf#L36

commy2:
Nvm, not really needed, as `null getVariable` reports `nil` and therefore the `forEach` is ignored. The `isNull` check hence isn't needed.
I like this for doing only one execNextFrame per object instead of per initPost eventhandler, so it should be done for optimization alone.

PR #631, opened 2017-04-08
keybind translations

commy2:
- use `keyName` instead of static array names for displaying the keybinds, which means that Y vs. Z is now respected on German keyboards etc. and the keys like "Home" are translated correctly.
- also translate the modifier keys
- affects menu and diary
- fixed a bug where duplicated keys would be shown as keybind title + tooltip in red instead of just the title.
- macros for XBOX controller keys
- added GVARs and replaced some magic numbers with macros
- rewrote a contrived logic to generate the readable keybinds for the diary

todos:
- test unbound key


SilentSpike:
Is there a reason the `keyName` command isn't used?

jonpas:
> Problem is that there is no* way to detect via SQF the language of the keyboard...

https://community.bistudio.com/wiki/language

commy2:
@SilentSpike Good question, I will take a look.
@jonpas That is not the keyboard language. The game language could be easily tested by making a dummy stringtable entry:
```
<English>English</English>
<German>German</German>
```

PabstMirror:
What about using a location hash instead of a oversized array?
`GVAR(keyHash) getVariable (str _x)`?

commy2:
One problem with `keyName` is, that it reports the keys as weird "string in string".

```sqf
keyName 0x15
-> """Z"""
```

Solution:
```sqf
_s = keyName 0x15; _s select [1, count _s -2]
```

commy2:
`keyName` is an excellent solution that even works for the "controller" keys:

```sqf
keyName 327684
"""XBox Up"""
```

commy2:
I use the location now like Pabst suggested. I don't use `keyName` directly with what is given by the script, because that would break the whole "unknown keycode: 123" thing and might even report wrong stuff (like "2xC", which is not supported).
Since namespaces are based on LOCATION's and are deleted at every mission start, I generate a CBA hash instead at game start (and take care of the duplicated ") and "deserialize" that with the function I written a few weeks ago. :)

PR #632, opened 2017-04-09
Add fnc_unicodeToChar.sqf to CBA_strings

AGPHeaddikilla:
This pull request will add the function CBA_fnc_unicodeToChar to CBA_strings.

Changes:
- changed CfgFunctions.hpp
- added file fnc_unicodeToChar.sqf


commy2:
Nifty.
```sqf
private _unicodeNumber = param [0, 0, [0]];
```
Could be:
```
params [["_unicodeNumber", 0, [0]]];
```

AGPHeaddikilla:
Thanks. I just changed the header. Although i liked the old format :)

commy2:
I adjusted the header for the ruby script that creates the doc on the wiki, but was unable to target this branch for whatever reason. #633 has it fixed  and has all commits from this branch.

commy2:
I searched how other languages name this function to maybe find a better function name, but only Java of all things was kinda helpful.

Python
`print u'\u1234'`

C++
`char b = '\u0444';`

Java
`String symbol = Character.toString("\u1234")`

`CBA_fnc_unicodeNumberToChar` is a bit long. Maybe shorten it to `CBA_fnc_unicodeToChar`? "unicode" already implies a "number" imo.

AGPHeaddikilla:
Well, i would take CBA_fnc_unicodeToChar. I just named it "char" because the unicode (number) only return single characters.
I think CBA_fnc_unicodeToString would imply the 'number to string'-thing.
For example:
```sqf
_inputNumber = 1;
_string = (str _inputNumber); //just string' it -> "1"
```

commy2:
I just noticed that this command exists:
https://community.bistudio.com/wiki/toString
>Converts the supplied Array of Numbers into a String. 
>The numbers in the array to be converted are the decimal Unicode representations of characters.

Is there any specific difference between the native command and this function?

AGPHeaddikilla:
@commy2 Good point. I thought about toString, but I havent checked if toString uses unicode numbers as "ids".
I think to use this "alternative" function is better because code will be better to read if using single unicodes. Maybe we could replace the `str parseText <x>;` with `toString [<x>];`

commy2:
`toString` is definitely better, because it works with more characters:

```sqf
str parseText format ["&#%1;", 0xe4];
-> " " // wrong
```

```sqf
toString [0xe4];
-> "ä" // correct
```

I just wonder if having a function is worth it if it's just a less powerful copy (only one char at a time) of the already existing `toString` command.

AGPHeaddikilla:
I just changed the function to both single and multiple character usage.

AGPHeaddikilla:
I just did a little study about different code parts:

```sqf
myNumbers = [(round random 999), ...];
//Converting tests with 1000 random numbers to unicode chars
["_string = ''; { _string = _string + (str parseText format['&#%1;', _x]); } forEach myNumbers;"] call BIS_fnc_codePerformance; //9.06306 ms
["_string = ''; { _string = _string + (toString [_x]); } forEach myNumbers;"] call BIS_fnc_codePerformance; //3.4878 ms
["_string = (toString myNumbers);"] call BIS_fnc_codePerformance; //0.0108 ms
["_string = ([myNumbers, true] call CBA_fnc_unicodeToChar);"] call BIS_fnc_codePerformance; //0.0192 ms
["_string = ([myNumbers, false] call CBA_fnc_unicodeToChar);"] call BIS_fnc_codePerformance; //0.191314 ms
```
And yes, it is slower to use single unicode convertions. Nevertheless this function would make scripts more clearly, since it have both string and array support.

jonpas:
A separate function would make more sense if you want different return values.

Dorbedo:
How about a Macro?

addons/strings/fnc_unicodeToChar.sqf
```diff
@@ -0,0 +1,62 @@
...
+    given unicode char <STRING>
+
+Examples:
+    (begin example)
+        _unicodeNumber = 49;
+        _unicodeChar = "";
+        _unicodeChar = ([_unicodeNumber] call CBA_fnc_unicodeToChar);
+        //_unicodeChar will return "1"
+    (end)
+    (begin example)
```
Killswitch00: repetition

addons/strings/fnc_unicodeToChar.sqf
```diff
@@ -0,0 +1,62 @@
...
+        _unicodeNumber = 49;
+        _unicodeChar = "";
+        _unicodeChar = ([_unicodeNumber, true] call CBA_fnc_unicodeToChar);
```
Killswitch00: Undocumented second, boolean parameter

addons/strings/fnc_unicodeToChar.sqf
```diff
@@ -0,0 +1,62 @@
...
+        //_unicdeChar will return ["1"]
```
Killswitch00: "_unicodeChar"

addons/strings/fnc_unicodeToChar.sqf
```diff
@@ -0,0 +1,62 @@
...
+
+Description:
+    Gets the unicode char of the given unicode number
+    For more information to unicodes visit: https://unicode-table.com/
+
+Parameters:
+    _unicodeNumber: the unicode number <NUMBER>
+
+Returns:
+    given unicode char <STRING>
```
Killswitch00: "unicode char"? Which of the many possible encodings?

Does this function map from a unicode code point number to a UTF-8 string representation of that code point?

Killswitch00:
Redundant. As mentioned, https://community.bistudio.com/wiki/toString already solves this problem.

commy2:
Close? I really don't think this is necessary and I fear that CBA could become (more) bloated.

AGPHeaddikilla:
@commy2 can be closed.

commy2:
O.K. Thanks anyway.

PR #634, opened 2017-04-10
function to add buttons to pause menu

commy2:
It's also compatible with the `replaceAbortButton` description.ext entry which contrary to it's name, doesn't replace the abort button, but adds a custom "Camp Maxwell" button and is apparently used in the campaign.


commy2:
```sqf
["Game Options", "RscDisplayGameOptions"] call CBA_fnc_addPauseMenuOption;
```
The scripting of the display can be done via it's onLoad eventhandler.

PR #635, opened 2017-04-10
Update some array functions

Dorbedo:
* Usage of the new selectMin / selectMax commands
* ~~select update~~
* TODO for shuffle done

PR #636, opened 2017-04-10
new test for shuffle

Dorbedo:
commys wish from #635 (depending on this pull request)

### DIFF
--- a/addons/arrays/test_shuffle.sqf
+++ b/addons/arrays/test_shuffle.sqf
@@ -21,14 +21,17 @@ TEST_OP(count _result,==,count _original,_fn);
     TEST_OP(_x,in,_original,_fn);
 } forEach _result;
 
-// test modify original
 _original = [1, 2, 3];
-_result = [_original,true] call CBA_fnc_shuffle;
-_result set [0,100];
-TEST_OP((_original select 0),==,100,_fn);
+_result = _original call CBA_fnc_shuffle;
+TEST_OP(count _result,==,1,_fn);
 
 _original = [];
 _result = [_original] call CBA_fnc_shuffle;
 TEST_OP(count _result,==,count _original,_fn);
+
+_original = [1, 2, 3];
+_result = [_original,true] call CBA_fnc_shuffle;
+_result set [0,100];
+TEST_OP((_original select 0),==,100,_fn);
 
 nil;
--- a/addons/arrays/test_shuffle.sqf
+++ b/addons/arrays/test_shuffle.sqf
@@ -21,18 +21,14 @@ TEST_OP(count _result,==,count _original,_fn);
     TEST_OP(_x,in,_original,_fn);
 } forEach _result;
 
-// Test depecated version.
+// test modify original
 _original = [1, 2, 3];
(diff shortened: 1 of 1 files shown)


addons/arrays/test_shuffle.sqf
```diff
@@ -21,18 +21,14 @@ TEST_OP(count _result,==,count _original,_fn);
     TEST_OP(_x,in,_original,_fn);
 } forEach _result;
 
-// Test depecated version.
+// test modify original
```
Killswitch00: Don't remove what the old test checked. *Add* a new one that modifies the input array.

Dorbedo: It has to be removed, since the behavior of the function has changed. Otherwise there will always be an error.

Killswitch00: Refactorise the old test to do what it did before and _add_ a new one that checks if the original array is modified.

PabstMirror: `[1,2,3,4,5,6,7] call CBA_fnc_shuffle; `
Old version returns: [4,7,2,1,3,5,6]
New version returns: []

Old version threw a warning to rpt, but still worked.
I am a little worried some people would use it like that.

Killswitch00: That aswell - use of the "direct array" parameter form is discouraged, but the function still works. Why break it?

Dorbedo: The warning has been in there since 2009.
My personal opinion is, that the CBA-users should have changed their functions after 8 years of warning.

Dorbedo: @Killswitch00 the new feature is not compatible with the old deprecated behavior

Killswitch00: Fair enough. The unit test refactorisation issue remains.

Dorbedo: I'm sorry, but I don't understand, what you want from me.
Could you please be so kind and point it out for me.

Killswitch00: Look at the last unit test you altered (line 31). That one is refactorised for the new parameter format. The first two tests in the original version should be kept, but updated in the same way. Then add the new test you added for the "modify input array" option.

addons/arrays/test_shuffle.sqf
```diff
@@ -21,18 +21,17 @@ TEST_OP(count _result,==,count _original,_fn);
     TEST_OP(_x,in,_original,_fn);
 } forEach _result;
 
-// Test depecated version.
 _original = [1, 2, 3];
 _result = _original call CBA_fnc_shuffle;
```
Killswitch00: No. Update the way the function is called and then do the same two checks that the original unit test performs.

Dorbedo: But it would be the same test like in line 16. So you run the test twice

Killswitch00: Didn't see that one. Good. All cases are covered.

PR #645, opened 2017-04-17
enable XEH for classes used on map (but not terrain objects)

commy2:
- enable XEH / `CBA_fnc_addClassEventHandler` for many additional classes, mainly stuff like modules, game logics, furniture, pop up targets...
- terrain objects (that belong to the map) are still disabled. this is done already by the scripting, because `typeOf` reports `""` apparently for them.
- this does not enable for `class Static` objects (houses, bushes, trees, rocks, road signs, flagpoles, hesco barriers ...)

Will probably make a ton of XEH not supported for X appear with other mods (e.g. ACE), because, well they don't, and were peviously disabled. Nothing that can't be fixed by the authors though.


addons/xeh/CfgVehicles.hpp
```diff
@@ -137,7 +213,7 @@ class CfgVehicles {
         XEH_ENABLED;
     };
 
-    // backwards comp, inert
+    // backwards comp, inertx
```
Killswitch00: "inertx" ?

commy2: lol
I think this was a failed attempt at ctrl + x.

Killswitch00:
This will need thorough testing with both current and older missions and addons.

commy2:
Thanks to #663, terrain objects are now explicitly skipped.

commy2:
Merge for RC? Can always undo _if_ it causes problems.

Killswitch00:
Yes, let's merge for the RC.

PR #647, opened 2017-04-18
Keybinding internals overhaul

commy2:
goal:
- redo the keyhandler code using new commands / techniques, and hopefully no longer use 5+ indents...
- add support for mouse buttons

requires redoing CBA_fnc_addKeybind too, but have to keep full bwc


commy2:
replaced by #651

PR #654, opened 2017-04-23
fix strange modifier key behavior

### LINKED ISSUE #213: Strange behaviour of modifier key.
two actions:
1. bound to any key (L)
2. bound to the same key, but with modifier: (Ctrl + L)

bugs:
- Pressing Ctrl + L and then releasing Ctrl first, triggers the action only bound to L.
- Even when releasing the modifier key, it still seems to be active for a short while, causing this input:
  `press and release Ctrl` -> `press L`
  to trigger the second action, despite ctrl not being pressed.

Will add some debug script later.

commy2:
- fixes #213

The shift/control/alt parameters passed by the keyDown/Up events are really crappy. When you released a key, say _shift_, the "shift" parameter will still report true for all keyDown/Up events that happen up to about half a second later, even though the keyUp event for "shift" already occurred.
With this PR we look for the modifier keys themselves and store their states in variables. This makes the behavior way more intuitive and predictable.
By doing this, this PR also adds modifier support to the mouse wheel.

### DIFF
--- a/addons/events/XEH_preInit.sqf
+++ b/addons/events/XEH_preInit.sqf
@@ -79,3 +79,7 @@ GVAR(keyHoldTimers) = call CBA_fnc_createNamespace;
 
     false
 }] call CBA_fnc_compileFinal;
+
+GVAR(shift) = false;
+GVAR(control) = false;
+GVAR(alt) = false;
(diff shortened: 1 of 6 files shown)


commy2:
^ should the key be released in the same frame, the keyUp would happen after the keyDown. The "true" state is unreliable as described above, but if the state is false, the key is guaranteed not held, so we can make the modifier key release frame precise.

PR #655, opened 2017-04-24
delay whole loop instead of each initPost event

commy2:
- minor optimization that also resolves a script error that would happen in ACE after 3.3

see: https://github.com/CBATeam/CBA_A3/pull/628#issuecomment-296521143

### DIFF
--- a/addons/xeh/fnc_init.sqf
+++ b/addons/xeh/fnc_init.sqf
@@ -33,9 +33,11 @@ if !(ISINITIALIZED(_this)) then {
 
     // run InitPost or put on stack
     if (SLX_XEH_MACHINE select 8) then {
-        {
-            [_x, [_this]] call CBA_fnc_execNextFrame;
-        } forEach (_this getVariable QGVAR(initPost));
+        [{
+            {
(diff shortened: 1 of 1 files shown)


commy2:
repro:
```sqf
["All", "init", {
    systemChat str [typeOf (_this select 0), "init"];

["All", "initPost", {
    systemChat str [typeOf (_this select 0), "initPos"];
```
```sqf
x3 = "B_Heli_Transport_01_F" createVehicle [0,0,0];
x3 = "O_Heli_Transport_04_F" createVehicle [0,0,0];
deleteVehicle x3;
```
without PR:
```sqf
["B_Heli_Transport_01_F","init"]
["O_Heli_Transport_04_F","init"]
["B_Heli_Transport_01_F","initPos"]
["","initPos"]
```
with PR:
```sqf
["B_Heli_Transport_01_F","init"]
["O_Heli_Transport_04_F","init"]
["B_Heli_Transport_01_F","initPos"]
```

Not executing for null objects makes it consistent with what would happen would the object be deleted before postInit.

PR #658, opened 2017-05-01
Debug watching for statements that run on a remote machine

PabstMirror:
Adds addition statement watchers to the debug menu that run on the server
They exec on the server and return the results back to the client.

Requires `EnableTargetDebug= 1;` in description.ext or 3den

image


commy2:
Any reason why `GVAR(watchOutput)` is an RscEdit? You cannot edit the output lines.

PabstMirror:
Allows you to copy the text

commy2:
Ah! Nice.

commy2:
Doesn't work so good with "Large" and "Very Large" ui size.

commy2:
Works great! Tested in local hosted MP. Amazing feature!

dedmen:
Just need another remoteExec button to execute code on a specific client

jonpas:
> Just need another remoteExec button to execute code on a specific client

I agree, could replace the "Server Exec" with "Target Exec" as well, since server can be selected in the drop-down as well.

commy2:
This line is the gem of this PR:
```sqf
_this = ([nil] apply compile _this) select 0;
```

It solves this annoying problem:
```sqf
_blub = {_blah = 1}; 
_blob = [nil] apply _blub; // generic error in expression
systemChat str [_blob];
```

```sqf
_blub = {_blah = 1}; 
_blob = [nil] apply _blub select 0;
systemChat str [_blob]; // [<null>]
```

PR #659, opened 2017-05-02
Debug watching for any client

commy2:
- Combobox to select target for the "server watch" thing, so you can peak on any client.


commy2:
ref: #658

PR #662, opened 2017-05-03
abort init events during savegame load

### LINKED ISSUE #661: Crash loading savegame
```
19:52:51 ["buffer a",[{
{
[_this] call _x;
} forEach (_this getVariable "cba_xeh_initPost");
},2010fdac100# 1569015: powerpolewooden_l_f.p3d]]
19:52:51 Error loading/saving variable cba_common_nextframebuffera
```

Loading a save game crashes, because the init eventhandler puts the initPost event function on the next frame buffer. It seems that the issue is, that you cannot update an array while the game is loading the serialized value.
Identifier is `GVAR(nextFrameBufferA)`.
This only happens when at least one map object using the init eventhandler is present. On Altis this is `Land_PowerPoleWooden_L_F`, which makes this crash map dependent.

commy2:
- fix #661

These seem to only happen for map objects that are created during the loading of a savegame. We don't want XEH for these objects anyway, so aborting is good enough.

```sqf
GVAR(LoadingCheck) = 0 spawn {
    disableSerialization;
    sleep 1E9;
};
```
is the best way I know of to detect this state.

### DIFF
--- a/addons/xeh/fnc_init.sqf
+++ b/addons/xeh/fnc_init.sqf
@@ -19,9 +19,15 @@ Examples:
 Author:
     commy2
 ---------------------------------------------------------------------------- */
+#define DEBUG_SYNCHRONOUS
 #include "script_component.hpp"
 
 params ["_this"];
+
+// fix for https://github.com/CBATeam/CBA_A3/issues/661
+if (scriptDone GVAR(LoadingCheck)) exitWith {
+    INFO_1("Abort init event during savegame load. Class: %1.",typeOf _this);
+};
 
 if !(ISINITIALIZED(_this)) then {
     SETINITIALIZED(_this);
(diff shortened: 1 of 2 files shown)


commy2:
Will print:
```
22:37:31 [CBA] (xeh) INFO: Abort init event during savegame load. Class: Land_PowerPoleWooden_L_F.
```
x222 on Altis, but I want to keep this for now for debugging reasons.

Btw., these objects only fire their init event during the savegame loading on the server machine.

Cdeeks:
Hi commy2 thanks again for working so quick on this. I'm new to Github not sure how it works. do you release a new version of CBA or do we take that code above and put it somewhere?

commy2:
You will have to wait for the next update, but @ViperMaul surely will release that ASAP after this pull request is approved and merged into the `release` branch.

commy2:
We probably want the other PR, as the fix there is more explicit.

PR #664, opened 2017-05-03
replace SERVER EXEC button with TARGET EXEC button

commy2:
- title

`TARGET EXEC` button to execute the "Execute" script on the machine selected for "Target Watch".


commy2:
ref: #658

PR #665, opened 2017-05-04
fix assignment operator return value bug for directCall

commy2:
- title

before PR:
```sqf
_a = {_b = 1} call CBA_fnc_directCall; // generic error
isNil "_a"
```

after PR:
```sqf
_a = {_b = 1} call CBA_fnc_directCall;
isNil "_a"
-> true
```

### DIFF
--- a/addons/common/fnc_directCall.sqf
+++ b/addons/common/fnc_directCall.sqf
@@ -22,12 +22,12 @@ Author:
 ---------------------------------------------------------------------------- */
 #include "script_component.hpp"
 
-params [["_CBA_code", {}, [{}]], ["_CBA_arguments", []]];
+params [["_CBA_code", {}, [{}]], ["_this", []]];
 
 private "_CBA_return";
 
 isNil {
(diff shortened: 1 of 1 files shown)


addons/common/fnc_directCall.sqf
```diff
@@ -22,12 +22,12 @@ Author:
...
 #include "script_component.hpp"
 
-params [["_CBA_code", {}, [{}]], ["_CBA_arguments", []]];
+params [["_CBA_code", {}, [{}]], ["_this", []]];
 
 private "_CBA_return";
 
 isNil {
-    _CBA_return = _CBA_arguments call _CBA_code;
+    _CBA_return = [_x] apply _CBA_code select 0;
```
PabstMirror: What is the point of `_x`?

commy2: So `_x` doesn't become undefined in `_CBA_code`, but stays `_x`.

PR #667, opened 2017-05-05
add ability to name groups in lobby

commy2:
- see title

They said it's impossible and it kinda is, except for this method. The only information that can be send to a client in the lobby is the role descriptions of the slots. `remoteExec` doesn't work and the description.ext is only available to the server machine.
By hiding our custom group name inside the role name of the first slot of the group, we can replace the default group name with anything using listbox fuckery.

This is how it looks:

Using "Team Leader@Team 1", the result will be a slot named "Team Leader" and the group will be named "Group 1". `@` is the magic separator.

Result:

Since this doesn't require any information from the description.ext (slot names are send anyway to anyone in the lobby) this does work for clients.

Negatives:
All old slot names using `@` will be cut off on the first `@`, but I doubt anyone uses this.
The slot has to be named for a group name to be chosen and it has to be manually entered.
If no respawn is used in the mission, not only the group leader, but all units have to add the group name. This is because only the first unit of the group is checked and dead units are removed from the group permanently including their slot from the lobby.

### DIFF
--- a/addons/network/CfgEventHandlers.hpp
+++ b/addons/network/CfgEventHandlers.hpp
@@ -1,0 +1,6 @@
+
+class Extended_DisplayLoad_EventHandlers {
+    class RscDisplayMultiplayerSetup {
+        ADDON = QUOTE(_this call COMPILE_FILE(fnc_initDisplayMultiplayerSetup));
+    };
+};
--- a/addons/network/CfgEventHandlers.hpp
+++ b/addons/network/CfgEventHandlers.hpp
@@ -1,6 +1,0 @@
-
-class Extended_DisplayLoad_EventHandlers {
-    class RscDisplayMultiplayerSetup {
-        ADDON = QUOTE(_this call COMPILE_FILE(fnc_initDisplayMultiplayerSetup));
-    };
-};
--- a/addons/network/config.cpp
+++ b/addons/network/config.cpp
@@ -14,7 +14,6 @@ class CfgPatches {
     };
 };
 
-#include "CfgEventHandlers.hpp"
 #include "CfgFunctions.hpp"
 
 #include "CfgSettings.hpp"
--- a/addons/network/config.cpp
+++ b/addons/network/config.cpp
@@ -14,6 +14,7 @@ class CfgPatches {
     };
 };
 
+#include "CfgEventHandlers.hpp"
 #include "CfgFunctions.hpp"
 
 #include "CfgSettings.hpp"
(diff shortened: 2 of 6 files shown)


jonpas:
I think it would be neat if this would automatically do `setGroupId` on player init so it would show in-game as well. As right now, you'd have to do `setGroupId` and also write it here to make it the same in-game and in-lobby.

commy2:
@jonpas You can already change the "Callsign" attribute of the group to do this. This is broken in MP in vanilla, but also fixed in CBA (unrelated to this).
One difference between the "Role Description" of a unit and the "Callsign" of a group is, that the former exclusively affects the lobby, while the latter is used ingame as well in a bunch of places and uses the `setGroupID` command instead of being purely a mission.sqm entry.
I know it sucks that you basically have to enter the same information twice, but I don't want that the "Role Description" entry is used as a replacement for the "Callsign". Maybe someone wants the callsign to be different from the lobby description too, who knows.

commy2:
Also, as @rebelvg pointed out, you can use the same lobby name for multiple groups on the same side twice, which the Callsign attribute doesn't allow for.

Drofseh:
> All old slot names using | will be cut off on the first |, but I doubt anyone uses this.

commy and I have discussed this a little on slack, but both the communities I am part of use | in slot naming to separate platoon, squad, and team designations.

for example
https://i.gyazo.com/76bf3a84266adb1ad358d3e2a118ec5d.png
and it looks even cleaner in game than it does in this screenshot.

As such we would definitely prefer a different character be used for the splitString.

I would suggest ^, if possible, as a reminder that what follows is being moved up to the group name

for example: 
1 Platoon | 2 Squad | Squad Leader^1'2
1 Platoon | 2 Squad | Alpha Team | Team Leader
1 Platoon | 2 Squad | Alpha Team | Rifleman
etc.

jonpas:
Semi-colon `;`?

Otherwise, why not something like that: `2nd Squad{Group X}`, encapsulated in `{}` (or something else like that except `()` obviously).

commy2:
Make `\|` escape it?

Drofseh:
1 Platoon | 2 Squad | Squad Leader\|1'2
Like that?
I'd be ok with this.

Dorbedo:
How about using another symbol as separator? There are many left e.g.: ``` §$%&#```

commy2:
`@`

Dorbedo:
lgtm:
`Teamlead@Group 1`
`Rifleman (AT)@Group 3`

commy2:
Changed it to `@`.

commy2:
Doc: https://github.com/CBATeam/CBA_A3/wiki/Name-Groups-in-Lobby

nomisum:
you are a fucking god :1st_place_medal:

PR #668, opened 2017-05-05
Optimize CBA_fnc_players

SilentSpike:
- Use `allDeadMen` instead of `allDead`

`allDead` includes destroyed vehicles.

### DIFF
--- a/addons/common/fnc_players.sqf
+++ b/addons/common/fnc_players.sqf
@@ -23,4 +23,4 @@ Author:
 #include "script_component.hpp"
 SCRIPT(players);
 
-(allPlayers) select { !(_x isKindOf "HeadlessClient_F") }
+(allUnits + allDeadMen) select {isPlayer _x && {!(_x isKindOf "HeadlessClient_F")}}
--- a/addons/common/fnc_players.sqf
+++ b/addons/common/fnc_players.sqf
@@ -23,4 +23,4 @@ Author:
 #include "script_component.hpp"
 SCRIPT(players);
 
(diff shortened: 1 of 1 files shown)


commy2:
https://github.com/CBATeam/CBA_A3/issues/353#issue-158313664

But I never reproduced this. Will try right now.

PabstMirror:
>NOTE: In player hosted game, the complete array of allPlayers may get delayed at the start. Use BIS_fnc_listPlayers if you need it earlier
image

commy2:
```sqf
addMissionEventHandler ["EachFrame", {
    diag_log [diag_frameNo, allPlayers, [] call CBA_fnc_players, allPlayers isEqualTo ([] call CBA_fnc_players)];
}];
```

```
 0:39:07 [11821,[B CSCustom0:1 (Gefr. commy2)],[B CSCustom0:1 (Gefr. commy2),B CSCustom0:2 (Gefr. commy2 (2)) REMOTE],false]
```

commy2:
But why does it use `allDead` and not `allDeadMen`?

>allDead
>Return a list of all dead units including agents and destroyed vehicles. Dead units may be in vehicles.

SilentSpike:
Is it confirmed that the old way truly does return all players under the same circumstances in which `allPlayers` doesn't?

commy2:
See my test. Was local hosted.

SilentSpike:
Ah yeah, missed that :+1: Can change this to simply switch to `allDeadMen` then

commy2:
It's still an optimization, because `isPlayer` filters out the wrecks atm. But this makes it slightly faster by not listing them in the first place.

PR #671, opened 2017-05-07
Dev tool to report performance of state machines

PabstMirror:
Tag @BaerMitUmlaut 

Reports rough average runtime for each statemachine.
Includes both the statemachine's functions time + CBA's clockwork time.
`diag_tickTime` has limited timer accuracy, but with a longer run time it seems to average out well enough.

No impact on normal release, requires `#define` to work.

Example RPT:
```
 CBA State Machine Results:
 ------------------ [Time: 59.4] -------------------------
 0: [ace_medical_blood] Average: 1.071 ms [4.031s / 3763]
 1: [ace_medical_ai] Average: 0.103 ms [0.388s / 3763]
```


BaerMitUmlaut:
Two things:
- Why not save the counter on the state machine itself instead of in a GVAR?
- Personally I'd not use another prefix for the function, the name of the function is almost 50 characters long, I'd rather mention in the header that this is only for development purposes

PabstMirror:
1. Can still track removed state machine and what their performance was
2. Fixed

PR #673, opened 2017-05-08
add function to translate keys

commy2:
- title

This is used in `help` and in `keybinding`, so might as well put it in a function.

```sqf
15 call CBA_fnc_localizeKey
"Tab"

[15, true, false, false] call CBA_fnc_localizeKey
"Shift+Tab"

[15, [false, true, true]] call CBA_fnc_localizeKey
"Ctrl+Alt+Tab"
```

### DIFF
--- a/addons/keybinding/fnc_localizeKey.sqf
+++ b/addons/keybinding/fnc_localizeKey.sqf
@@ -1,0 +1,54 @@
+/* ----------------------------------------------------------------------------
+Function: CBA_fnc_localizeKey
+
+Description:
+    Translate DIK code and modifier keys to readable keybind.
+
+Parameters:
+    _key     - DIK code <NUMBER>
+    _shift   - Shift pressed? <BOOL>
+    _control - Control pressed? <BOOL>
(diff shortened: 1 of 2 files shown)


addons/keybinding/fnc_localizeKey.sqf
```diff
@@ -0,0 +1,54 @@
...
+// Try to convert dik code to a human key code.
+private _keyName = GVAR(keyNames) getVariable str _key;
+
+if (isNil "_keyName") then {
+    _keyName = format [localize LSTRING(unkownKey), _key];
+};
+
+// Build the full key combination name.
+if (_alt && {!(_key in [DIK_LMENU, DIK_RMENU])}) then {
+    _keyName = localize "str_dik_alt" + "+" + _keyName;
```
SilentSpike: Totally minor, but is `[localize "bla", _keyName] joinString "+"` a faster method for these modifier keys?

SilentSpike: Could also append them all to an array and then `joinString` at the end

PR #678, opened 2017-05-11
CfgFunctions PATHTO_FNC macro for keybinding

commy2:


PabstMirror:
Would this be a problem?
```
-// Automatically generated by 'functions_config.rb'
 -// DO NOT MANUALLY EDIT THIS FILE!
```

commy2:
No one uses that ruby script anymore. It would pick up the functions for PREP too and add them to CfgFunctions.

PR #685, opened 2017-05-17
Always re-run preInit on mission load/new

### LINKED ISSUE #684: CBA settings in editor dont work since 3.3.1
**Arma 3 Version:** `1.70` stable
**CBA Version:** `3.3.1` stable

**Mods:**
```
- CBA_A3 only
```

**Description:**
When a mission.sqm is loaded into the editor,the CBA settings stoped working

**Steps to reproduce:**
https://www.youtube.com/watch?v=HOyEquBoRpc

-  Editor 

**RPT log file:**
https://gist.github.com/Radeni/9eb3c5a8630f19dc796dcc2f70efcb9e

PabstMirror:
Should fix #684

Seems like in 1.70, missionnamespace alwayse gets cleared out
(which seems good, just causes some problems)

### DIFF
--- a/addons/xeh/fnc_initDisplay3DEN.sqf
+++ b/addons/xeh/fnc_initDisplay3DEN.sqf
@@ -1,12 +1,13 @@
 #include "script_component.hpp"
 
-// fix for preInit = 1 functions not being executed when entering 3den from main menu
-[] call CBA_fnc_preInit;
+params ["_display"];
 
-// since 1.60, preInit = 1 functions aren't executed when returning from a preview either ...
-add3DENEventHandler ["OnMissionPreviewEnd", {[] call CBA_fnc_preInit}];
+private _fnc_watchDog = {
+    if (!ISPROCESSED(missionNamespace)) then {
+        diag_log text format ["XEH: missionNamespace processed [%1]", ISPROCESSED(missionNamespace)];
+        [] call CBA_fnc_preInit;
+    };
+};
 
-// switching terrains in 3den will reset missionNamespace
-add3DENEventHandler ["OnTerrainNew", {[] call CBA_fnc_preInit}];
-add3DENEventHandler ["OnMissionNew", {[] call CBA_fnc_preInit}];
-add3DENEventHandler ["OnMissionLoad", {[] call CBA_fnc_preInit}];
+_display displayAddEventHandler ["MouseMoving", _fnc_watchDog];
+_display displayAddEventHandler ["MouseHolding", _fnc_watchDog];
--- a/addons/xeh/fnc_initDisplay3DEN.sqf
+++ b/addons/xeh/fnc_initDisplay3DEN.sqf
@@ -8,14 +8,5 @@ add3DENEventHandler ["OnMissionPreviewEnd", {[] call CBA_fnc_preInit}];
 
 // switching terrains in 3den will reset missionNamespace
 add3DENEventHandler ["OnTerrainNew", {[] call CBA_fnc_preInit}];
-
-private _fnc_preInitMissionConfig = {
-    {
-        if (_x select 0 == "" && {_x select 1 == "preInit"}) then {
-            [] call (_x select 2);
-        };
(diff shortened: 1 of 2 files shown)


addons/xeh/fnc_initDisplay3DEN.sqf
```diff
@@ -8,14 +8,5 @@ add3DENEventHandler ["OnMissionPreviewEnd", {[] call CBA_fnc_preInit}];
...
-    {
-        if (_x select 0 == "" && {_x select 1 == "preInit"}) then {
-            [] call (_x select 2);
-        };
-    } forEach (missionConfigFile call CBA_fnc_compileEventHandlers);
-};
-
-add3DENEventHandler ["OnMissionNew", _fnc_preInitMissionConfig];
-add3DENEventHandler ["OnMissionLoad", _fnc_preInitMissionConfig];
+add3DENEventHandler ["OnMissionNew", {[] call CBA_fnc_preInit}];
```
commy2: Is OnTerrainNew already covered by OnMissionNew and OnMissionLoad?

PabstMirror: Going from one terrain to another
```
"OnMissionNew"
XEH: missionNamespace processed [false]
[83008,343.676,0,"XEH: PreInit started. v3.3.1.170504"]
[83008,344.812,0,"XEH: PreInit finished."]
"OnTerrainNew"
XEH: missionNamespace processed [false]
83008,346.107,0,"XEH: PreInit started. v3.3.1.170504"]
```
both events happen on the same frame, but missionNamespace is wiped between them
`OnTerrainNew` seems to be last, so if we don't add the event, nothing will be defined

PabstMirror:
- Major problem is this doesn't work when loading of a mission fails.
e.g. You load a mission that has requiredAddons you aren't running.
Mission namespace gets wiped, but no `onMissionLoad` event.

- Minor problem, we run preInit more than we have to, worst case seems to be going from one map to loading another, preInit runs 3 times:
```
"OnMissionNew"
XEH: missionNamespace processed [false]
"OnTerrainNew"
XEH: missionNamespace processed [false]
"OnMissionLoad"
XEH: missionNamespace processed [false]
```

addons/xeh/fnc_initDisplay3DEN.sqf
commy2: OnTerrainNew is still needed as it fires after OnMissionLoad and the missionNamespace get's wiped inbetween them for whatever reason.

commy2:
To the 2nd point, doesn't really matter imo. preInit is mainly copying over stuff from ui namespace via PREP.

1st one. Why would loading a mission fail?

PabstMirror:
`Warning Message: mission.sqm/Mission/Entities/Item2/Entities/Item0.type: Vehicle class CUP_O_TK_Crew no longer exists`

image

commy2:
And that opens an empty mission on the same map as selected, right?

PabstMirror:
yea

commy2:
I don't like how convoluted this has become. All 3 events are needed, but they all fire, because BI keeps wiping the namespace.

commy2:
Yeah, I like this more.

PabstMirror:
This also fixes the multiple running of preInit

addons/xeh/fnc_initDisplay3DEN.sqf
```diff
@@ -1,21 +1,13 @@
...
-
-private _fnc_preInitMissionConfig = {
-    {
-        if (_x select 0 == "" && {_x select 1 == "preInit"}) then {
-            [] call (_x select 2);
-        };
-    } forEach (missionConfigFile call CBA_fnc_compileEventHandlers);
+private _fnc_watchDog = {
+    if (!ISPROCESSED(missionNamespace)) then {
+        diag_log text format ["XEH: missionNamespace processed [%1]", ISPROCESSED(missionNamespace)];
```
commy2: `ISPROCESSED(missionNamespace)` is always false, because that is the condition of the `if`.

commy2:
Works and fixes #684.

PR #688, opened 2017-05-23
addKeybind - Handle empty strings to indicated no code

PabstMirror:
`  _downCode       - Code for down event, empty string for no code. <CODE>`

ACRE uses this:
`["ACRE2", "CycleRadio", (localize LSTRING(CycleRadio)), { [1] call FUNC(cycleRadios) }, "", [58, [true, false, true]]] call cba_fnc_addKeybind;`

### DIFF
--- a/addons/keybinding/fnc_addKeybind.sqf
+++ b/addons/keybinding/fnc_addKeybind.sqf
@@ -88,9 +88,15 @@ if (_defaultKeybind isEqualTypeParams [0, false, false, false]) then {
     _defaultKeybind = [_defaultKey, [_defaultShift, _defaultControl, _defaultAlt]];
 };
 
-// Handle empty strings to indicated no code
-if (_downCode isEqualType "") then {_downCode = {};};
-if (_upCode isEqualType "") then {_upCode = {};};
+// Handle non-code type being passed for up/down code args
+if (!(_downCode isEqualType {})) then {
+    if (!(_downCode isEqualTo "")) then {WARNING_1("_downCode: Non code type [%1] will be ignored",_downCode);};
+    _downCode = {};
+};
+if (!(_upCode isEqualType {})) then {
+    if (!(_upCode isEqualTo "")) then {WARNING_1("_upCode: Non code type [%1] will be ignored",_upCode);};
+    _upCode = {};
+};
 
 // Make sure modifer is set to true, if base key is a modifier
 _defaultKeybind params [["_defaultKey", 0, [0]], ["_defaultModifiers", [], [[]]]];
--- a/addons/keybinding/fnc_addKeybind.sqf
+++ b/addons/keybinding/fnc_addKeybind.sqf
@@ -71,8 +71,8 @@ params [
     ["_addon", "", [""]],
     ["_addonAction", "", [""]],
     ["_title", "", ["", []]],
-    ["_downCode", {}, [{}]],
-    ["_upCode", {}, [{}]],
+    ["_downCode", {}, [{}, ""]],
+    ["_upCode", {}, [{}, ""]],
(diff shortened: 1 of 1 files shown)


jonpas:
Won't this automatically be done by `params`? As it only accepts code `{}` as input, if it's not code it will use the default value, which is code `{}` anyways.

addons/keybinding/fnc_addKeybind.sqf
```diff
@@ -88,6 +88,10 @@ if (_defaultKeybind isEqualTypeParams [0, false, false, false]) then {
     _defaultKeybind = [_defaultKey, [_defaultShift, _defaultControl, _defaultAlt]];
 };
 
+// Handle empty strings to indicated no code
+if (_downCode isEqualType "") then {_downCode = {};};
+if (_upCode isEqualType "") then {_upCode = {};};
```
commy2: Why not go the extra mile and do:
```sqf
+if (_downCode isEqualType "") then {
    _downCode = compile _downCode;
};
 +if (_upCode isEqualType "") then {
    _upCode = compile _upCode;
};
```

Would make more sense to me.

commy2:
@jonpas True, but it also throws an error. From what I understand, STRING used to work as replacement for CODE just like how you can use STRING inside `addEventHandler`.

jonpas:
Ah, you are correct indeed, totally forgot about the error.

addons/keybinding/fnc_addKeybind.sqf
PabstMirror: The old method is consistent with the old way which only checked if it was a string
```
    if(IS_CODE(_downCode)) then {
        [_hashDown, "keydown"] call cba_fnc_removeKeyHandler;
```

Compiling could cause problems if they pass something like `"#nothing"`.
I figured it's deprecated anyway as there is no reason to not pass `{}`

commy2: I know what you mean, but then this isn't limited to STRING either, but any type.
And `"#nothing"` is a bad example, because that actually doesn't throw an error when compiling or calling the result, because `#` is treated like `#LINE 1` and other instructions that survive preprocessing even though it does nothing.

addons/keybinding/fnc_addKeybind.sqf
```diff
@@ -88,6 +88,10 @@ if (_defaultKeybind isEqualTypeParams [0, false, false, false]) then {
     _defaultKeybind = [_defaultKey, [_defaultShift, _defaultControl, _defaultAlt]];
 };
 
+// Handle empty strings to indicated no code
```
Killswitch00: This results in any string, empty or not, to be regarded as an empty code block

addons/keybinding/fnc_addKeybind.sqf
```diff
@@ -88,6 +88,16 @@ if (_defaultKeybind isEqualTypeParams [0, false, false, false]) then {
...
 
+// Handle non-code type being passed for up/down code args
+if (!(_downCode isEqualType {})) then {
+    if (!(_downCode isEqualTo "")) then {WARNING_1("_downCode: Non code type [%1] will be ignored",_downCode);};
+    _downCode = {};
+};
+if (!(_upCode isEqualType {})) then {
+    if (!(_upCode isEqualTo "")) then {WARNING_1("_upCode: Non code type [%1] will be ignored",_upCode);};
+    _upCode = {};
+};
```
commy2: `params` already guarantees that the type is either CODE or STRING, so `!( isEqualType {})` and `!( isEqualType "")` is tautologous.
I really don't see whats wrong with just compiling the "code" if it's STRING type.

PR #690, opened 2017-06-03
DirectCall - prevent problems with exitWith

### LINKED ISSUE #689: directCall problems with exitWith
Ref https://github.com/acemod/ACE3/issues/5240
```
[] spawn {
    [{
        systemChat str 1234;
        if (true) exitWith {
            true
        };
    }, []] call CBA_fnc_directCall;
};
```
```
14:50:11 Error in expression <il {
_CBA_return = [_x] apply _CBA_code select 0;
};

if (!isNil "_CBA_return") >
14:50:11   Error position: <select 0;
};

if (!isNil "_CBA_return") >
14:50:11   Error select: Type Bool, expected Array,String,Config entry
14:50:11 File \x\cba\addons\common\fnc_directCall.sqf [CBA_fnc_directCall], line 1747
```

PabstMirror:
Fixes #689

### DIFF
--- a/addons/common/fnc_directCall.sqf
+++ b/addons/common/fnc_directCall.sqf
@@ -27,7 +27,8 @@ params [["_CBA_code", {}, [{}]], ["_this", []]];
 private "_CBA_return";
 
 isNil {
-    _CBA_return = [_x] apply _CBA_code select 0;
+    // Wrap the _CBA_code in an extra call block to prevent problems with exitWith and apply
+    _CBA_return = ([nil] apply {call _CBA_code}) select 0;
 };
 
 if (!isNil "_CBA_return") then {_CBA_return};
--- a/addons/common/fnc_directCall.sqf
+++ b/addons/common/fnc_directCall.sqf
@@ -28,7 +28,7 @@ private "_CBA_return";
 
 isNil {
     // Wrap the _CBA_code in an extra call block to prevent problems with exitWith and apply
-    _CBA_return = ([nil] apply {call _CBA_code}) select 0;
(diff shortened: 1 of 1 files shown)


addons/common/fnc_directCall.sqf
```diff
@@ -27,7 +27,8 @@ params [["_CBA_code", {}, [{}]], ["_this", []]];
 private "_CBA_return";
 
 isNil {
-    _CBA_return = [_x] apply _CBA_code select 0;
+    // Wrap the _CBA_code in an extra call block to prevent problems with exitWith and apply
+    _CBA_return = ([nil] apply {call _CBA_code}) select 0;
```
commy2: This would mean that the `_x` variable is overwritten.

PabstMirror:
```
_x = 5; 
x3 = [{_x = 4; systemChat str _x; }, 4] call cba_fnc_directCall;
systemChat str _x;
```
prints 4, 5. 
  I think the `_x` is privated from the `apply`, not sure?

commy2:
I think it should just be:
```sqf
- _CBA_return = [_x] apply _CBA_code select 0;
+ _CBA_return = [_x] apply _CBA_code param [0];
```

PabstMirror:
That would fail if something returns an array:
```
_code = {num = random 1; if (num < 0.5) exitWith {[true, num]}; [false, num]};
```
`[false,0.914843]`
`true`

commy2:
Then I think your solution is best. But with `[_x]` and not `[nil]`.

PR #697, opened 2017-06-07
JR for pistol optics

robalo:
- Add 2 new JR classes for mounting handgun optics" a common mount and a picatinny rail.


robalo:
ignore, this branch has extra unrelated changes

PR #699, opened 2017-06-09
Add ability to pass args to chat commands

PabstMirror:
@jonpas requested this

### DIFF
--- a/addons/events/fnc_registerChatCommand.sqf
+++ b/addons/events/fnc_registerChatCommand.sqf
@@ -8,7 +8,7 @@ Parameters:
     _command      - Chat command <STRING>
     _code         - Code to execute after command was entered. <CODE>
     _availableFor - "all", "admin" or "adminLogged" (optional, default: "admin") <STRING>
-    _thisArgs     - Variables available to the function <ANY>
+    _thisArgs     - Arguments to pass to event chat handler code <ANY>
 
 Returns:
     _return - true: Success, false: Error <BOOLEAN>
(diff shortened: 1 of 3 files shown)


jonpas:
And tested with https://github.com/Theseus-Aegis/tac-a3-core/commit/fca81726424ce9cdd7f3f19f0cd2c28b521c7506.

addons/events/fnc_registerChatCommand.sqf
```diff
@@ -8,6 +8,7 @@ Parameters:
     _command      - Chat command <STRING>
     _code         - Code to execute after command was entered. <CODE>
     _availableFor - "all", "admin" or "adminLogged" (optional, default: "admin") <STRING>
+    _thisArgs - Variables available to the function <ANY>
```
jonpas: Spacing like the other arguments?

addons/events/fnc_registerChatCommand.sqf
```diff
@@ -8,6 +8,7 @@ Parameters:
     _command      - Chat command <STRING>
     _code         - Code to execute after command was entered. <CODE>
     _availableFor - "all", "admin" or "adminLogged" (optional, default: "admin") <STRING>
+    _thisArgs     - Variables available to the function <ANY>
```
Killswitch00: "Variables" ought to be "Argument(s)" and "the function" is not clear - does it refer to "_code"?

(Put yourself in the shoes of a user reading the documentation for this function for the first time)

Also, add an example.

PR #702, opened 2017-06-11
Implement taskDefend ratio arguments

SilentSpike:
- Update function header
- Change threshold to work on a `>=` comparison as it is more intuative as
  an argument
- Make patrol parameter accept a percentage chance to provide better
  control over units that will patrol
- Add a hold parameter to allow the function to force units to hold their
  positions inside buildings and not react to combat

This is the re-implementation of #573


SilentSpike:
Is there a reason the waypoint doesn't make use of the function? Seems like a lot of duplicate code

Issue #693, opened 2017-06-05
Request player event for joining / leaving group


Crazy538:
Is there scope for a "player joined group" and "player left group" event in your player events framework? Currently, there is no way of detecting these events without polling (please correct me if I'm wrong) and seeing as you are already doing the polling with the intention of reducing the need elsewhere I figured I'd ask. The main use I can think of is restricting group sizes in missions and game modes but it could be used in all sorts of situations I imagine.

commy2:
All the currently available events are used in ACE or at least were used at some point. Sure we could add a check for the players group, but then it does check for that every frame. Is it worth it? I'm not convinced it is.

Crazy538:
Ah, are they every frame? Yea, possibly not worth it at that level. Thought I'd ask is all. If you guys don't' see a use for this then fair enough :)

Killswitch00:
Support added to CBA 3.4.0

PR #703, opened 2017-06-11
add Group and Leader Changed Player Event

jokoho48:
- Event when group player changes
- Event when the player current leader changes
- #693


addons/events/fnc_addPlayerEventHandler.sqf
```diff
@@ -114,6 +130,18 @@ if (_id != -1) then {
                 GVAR(oldUnit) = _player;
             };
 
+            private _data = group _player;
```
dedmen: Why re-private `_data` ? Line 151 is not doing that. In the old version only the first `_data` assignment was privated

PR #708, opened 2017-06-22
Add XEH to missing classes (modules)

PabstMirror:
Fix:
```
[XEH]: Module_F does not support Extended Event Handlers! Addon: curator
[XEH]: Land_Laptop_02_unfolded_F does not support Extended Event Handlers! Addon: argo
```

Ace adds XEH to "All" so it runs the fallback loop:
`One or more children of class All do not support Extended Event Handlers. Fall back to loop.`


commy2:
Thanks.

PR #710, opened 2017-06-24
Quick fix for getArea

fishykins:
somewhere along the line "triggerArea" started returning z values, making the return array size 6. This causes problems for other functions using the area format so this change simply removes the 6th element of the array.

### DIFF
--- a/addons/common/fnc_getArea.sqf
+++ b/addons/common/fnc_getArea.sqf
@@ -1,6 +1,8 @@
 /* ----------------------------------------------------------------------------
 Function: CBA_fnc_getArea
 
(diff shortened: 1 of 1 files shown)


PabstMirror:
https://community.bistudio.com/wiki/triggerArea
>NOTE: Since Arma 3 v1.59.135137, triggerArea returns 3rd dimension for the monitored area. If height is not set, the value for it would be -1;

SilentSpike:
Worth noting there is now a BI function which does this too: https://community.bistudio.com/wiki/BIS_fnc_getArea

Perhaps the CBA version should be deprecated

Killswitch00:
Yes, this looks like it could be gently deprecated. Please add a deprecation comment in the header (see arrays/fnc_join.sqf for the format)

PR #711, opened 2017-06-24
remove useless privates in player Events

jokoho48:
Changes after Dedmens comment

https://github.com/CBATeam/CBA_A3/pull/703/files/52f988ba478e78acc659bea94b37b57e2d3ca7bc#r123883063


dedmen:
privates are not really necessary anyway as there are no variables in any upper scope. But because there is an upper scope because the EventHandler recompile crap, private still provides a performance improvement.

commy2:
The variable is reused from L133.

PR #717, opened 2017-07-01
Add a Slower PFH that only checks every 5 secs to Player Events

jokoho48:
- Add a slower PFH to player EHs
- move Loadout, group and leader changed to the slower PFH

### DIFF
--- a/addons/events/fnc_addPlayerEventHandler.sqf
+++ b/addons/events/fnc_addPlayerEventHandler.sqf
@@ -121,6 +121,7 @@ if (_id != -1) then {
         GVAR(oldVisionMode) = -1;
         GVAR(oldCameraView) = "";
         GVAR(oldVisibleMap) = false;
+        GVAR(nextSlowCheck) = 0;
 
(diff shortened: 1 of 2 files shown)


PabstMirror:
What about if we still use the same EachFrame eh instead of adding the overhead of a new PFEH
```
// Do Fast Checks
if (GVAR(nextSlowCheck) < CBA_missionTime) then {
    GVAR(nextSlowCheck) = GVAR(nextSlowCheck) + 5;
    // Do Slow Checks
};
```
5 second seems a little long for the loadout/inventory check IMHO
e.g. ace_hearing, put on a helmet with hearing protection and you won't get audible feedback right away

commy2:
I don't like this. I think it's not needed at all.

SilentSpike:
I agree. If anything does this not add complexity to the PFH? Has performance difference been tested?

jokoho48:
i made this because commy2 noticed in the slack that dont like it that group and leader gets checked everyframe. that is why i changed it so that its only checked 1 a sec

Glowbal:
Is there a performance problem? If so, how much and why change these 3 events and not others? 

If not, I feel this risks the responsiveness of dependent listeners and affect gameplay in a negative manner.

PR #720, opened 2017-07-05
Fix #719 Invalid result for CBA_fnc_getNearest

### LINKED ISSUE #719: CBA_fnc_getNearest invalid results
**Arma 3 Version:** `1.72` stable
**CBA Version:** `3.4.0.170627` stable 

**Mods:**
```
- CBA_A3
```

**Description:**
- getNearest returns invalid results

**Steps to reproduce:**
result = [[0,0,0], [[10,10,0],[1,1,0], [5,5,0]]] call CBA_fnc_getNearest;
result => [10,10,0,[1,1,0],[5,5,0]]
expected => [1,1,0]

nilsms:
Fix #719 Invalid result for CBA_fnc_getNearest
res = [[0,0,0], [[10,10,0],[1,1,0], [5,5,0]]] call CBA_fnc_getNearest;
result before fix => [10,10,0,[1,1,0],[5,5,0]]
result after fix => [1,1,0]

### DIFF
--- a/addons/common/fnc_getNearest.sqf
+++ b/addons/common/fnc_getNearest.sqf
@@ -41,8 +41,7 @@ private _return = [[], objNull] select (isNil {param [2]});
 
     if (_distance < _radius) then {
         if !(call _code) exitWith {}; // don't move up. condition has to return false, vs. has to return true. Can be nil!
-
-        if !(_return isEqualType objNull) then {
+        if (count _this > 2) then {
             _return pushBack _x;
         } else {
             _radius = _distance;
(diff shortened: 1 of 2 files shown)


PabstMirror:
getNearest has 2 return types based on the presence of the radius param

So I think we want something like this
```
        if (count _this > 2) then {
            _return pushBack _x;
        } else {
            _radius = _distance;
            _return = _x;
        };
```
from https://github.com/nilsms/CBA_A3/blob/8d3519bf9e6ddf83a97a6a2247cb122ae96b4a2f/addons/common/fnc_getNearest.sqf

`isEqualType objNull` is the problem, because it doesn't work with arrayPositions as inputs

nilsms:
Your right, I missed the "or List of entities within given distance" part.

nilsms:
Quick rewrite of the patch to support returning multiple entities. Tested in game with both array positions and vehicles. Also added test case with radius.

Rewrote the branch commit history, old habit from other projects, not sure what you prefer.

PR #723, opened 2017-07-08
Add settings pbo prefix file

Dystopian:


commy2:
What for? This is exactly the same as if you'd use no PBOPREFIX.

Dystopian:
Really, I rechecked, makepbo without -N key (which I used by default) uses `Prefix = cba_settings_userconfig`. So if other tools you know do the same without PBOPREFIX plese let me know or just close this PR.

commy2:
I fail to see the difference between what you wrote and what it does after this PR. If a tool requires PBOPREFIX, the tool is incomplete.

Dystopian:
it's OK. Another time bad documentation made its bad thing.

PR #724, opened 2017-07-09
Add Target Debug run with addon config

### LINKED ISSUE #712: Add addon activation of extended MP debug console
**Arma 3 Version:** `1.73` (dev)
**CBA Version:** `3.3.1`

**Mods:**
```
- CBA_A3
```

**Description:**
- On arma 1.73 dev branch the debug console can now be made active when an addon is loaded with `enableDebugConsole = 2` in the config root. The extended version added in #658 should have a similar method of activation so that (for example) admins could run an addon to always have this available.

Dystopian:
Fixes #712.
~Addon config has a greater priority than mission one.~
Also function description fixed.

- [ ] If merged, fix [Target-Debugging wiki](/CBATeam/CBA_A3/wiki/Target-Debugging) page

### DIFF
--- a/addons/diagnostic/XEH_preInit.sqf
+++ b/addons/diagnostic/XEH_preInit.sqf
@@ -18,7 +18,8 @@ GVAR(projectileTrackedUnits) = [];
 
 ADDON = true;
 
-if (getMissionConfigValue ["EnableTargetDebug", 0] isEqualTo 1) then {
+private _cfg = configFile >> "EnableTargetDebug";
+if (1 isEqualTo ([getMissionConfigValue ["EnableTargetDebug", 0], getNumber _cfg] select isNumber _cfg)) then {
     INFO("EnableTargetDebug is enabled");
 
     [QGVAR(watchVariable), {
--- a/addons/diagnostic/XEH_preInit.sqf
+++ b/addons/diagnostic/XEH_preInit.sqf
@@ -18,8 +18,7 @@ GVAR(projectileTrackedUnits) = [];
 
 ADDON = true;
 
-private _cfg = configFile >> "EnableTargetDebug";
-if (1 isEqualTo ([getMissionConfigValue ["EnableTargetDebug", 0], getNumber _cfg] select isNumber _cfg)) then {
+if (1 == getMissionConfigValue ["EnableTargetDebug", 0] || {1 == getNumber (configFile >> "EnableTargetDebug")}) then {
     INFO("EnableTargetDebug is enabled");
 
     [QGVAR(watchVariable), {
--- a/addons/diagnostic/fnc_initTargetDebugConsole.sqf
+++ b/addons/diagnostic/fnc_initTargetDebugConsole.sqf
@@ -17,8 +17,7 @@ Author:
 
 #define COUNT_WATCH_BOXES 8
 
-private _cfg = configFile >> "EnableTargetDebug";
-if !(1 isEqualTo ([getMissionConfigValue ["EnableTargetDebug", 0], getNumber _cfg] select isNumber _cfg)) exitWith {};
+if !(1 == getMissionConfigValue ["EnableTargetDebug", 0] || {1 == getNumber (configFile >> "EnableTargetDebug")}) exitWith {};
 
 params ["_display"];
 TRACE_1("adding server watch debug",_display);
--- a/addons/diagnostic/fnc_initTargetDebugConsole.sqf
+++ b/addons/diagnostic/fnc_initTargetDebugConsole.sqf
@@ -2,8 +2,8 @@
 Function: CBA_diagnostic_fnc_initTargetDebugConsole
 
 Description:
-    Adds addition watch statements that are run on a remote target and have their values returned to the client.
-    Requires `EnableTargetDebug = 1; `in description.ext
+    Adds additional watch statements that are run on a remote target and have their values returned to the client.
+    Requires `EnableTargetDebug = 1;` in addon root config or description.ext or 3den scenario attribute with the same name
 
 Author:
     (based on BIS's RscDebugConsole.sqf)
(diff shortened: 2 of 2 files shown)


jonpas:
What about the new array syntax: `enableDebugConsole[] = {"SteamID1", "SteamID2", ...}`?

Dystopian:
> Target Debugging is only available, when the Debug Console is available

So Debug console checks steam IDs if they exists. Why do you need to define different ways for both consoles? If you insist I can implement it.

jonpas:
Oh right, never mind! :P

addons/diagnostic/XEH_preInit.sqf
```diff
@@ -18,7 +18,8 @@ GVAR(projectileTrackedUnits) = [];
 
 ADDON = true;
 
-if (getMissionConfigValue ["EnableTargetDebug", 0] isEqualTo 1) then {
+private _cfg = configFile >> "EnableTargetDebug";
+if (1 isEqualTo ([getMissionConfigValue ["EnableTargetDebug", 0], getNumber _cfg] select isNumber _cfg)) then {
```
commy2: Should be:
```sqf
getMissionConfigValue ["EnableTargetDebug", getNumber (configFile >> "EnableTargetDebug")] isEqualTo 1
```
imo. There is no reason for an addon to overwrite the mission if the mission sets the value explicitly.

addons/diagnostic/fnc_initTargetDebugConsole.sqf
```diff
@@ -17,7 +17,8 @@ Author:
 
 #define COUNT_WATCH_BOXES 8
 
-if (!((getMissionConfigValue ["EnableTargetDebug", 0]) isEqualTo 1)) exitWith {};
+private _cfg = configFile >> "EnableTargetDebug";
+if !(1 isEqualTo ([getMissionConfigValue ["EnableTargetDebug", 0], getNumber _cfg] select isNumber _cfg)) exitWith {};
```
commy2: same

commy2:
>Addon config has a greater priority than mission one.

Why? Nowhere else is that the case.

Dystopian:
Target Debug is server admin/developer tool disabled by default. Security questions are solved with vanilla `enableDebugConsole`. There are no any reasons to disable it in mission config. Addon priority is for those times when you play some mission where you need to debug something right now (multiplayer missions are difficult to debug, so sometimes you have to debug in playtime) but for some reason the author disabled it (by chance or purposely).

Dorbedo:
I personally don't agree with this.
If the missiondesigner has disabled the feature on purpose, his decision should be respected and not be overwritten by an addon.
If the missiondesigner didn't use the value, the Addon specified value could be used.
Just like commys suggestion.

Dystopian:
What purpose can it be? As server admin and addon tester/developer I can't agree.

commy2:
It should work the same way as vanilla does described here: #724 
I don't know which way that is.

Dorbedo:
[BIS-Wiki Vanilla order](https://community.bistudio.com/wiki/Mission_Editor:_Debug_Console_(Arma_3))
```
1. Mission param is checked. If is not defined then...
2. description.ext param enableDebugConsole is checked. If is not defined then...
3. Eden attribute option is checked. If not defined then...
4. Global/mod param enableDebugConsole is checked.
```

commy2:
So mission config beats addon config. Just like everything else. It's a good rule to go by. It seems to me like this PR does it the opposite way though.

SilentSpike:
I sort of don't see the point in an addon being able to enable the debug console if the mission can just override it anyway 🤔

commy2:
I guess it's more convenient to be able to change the default behavior by addon.
I'd say that almost all missions that "disable" the debug console, actually just didn't bother adding the entry. And if it's disabled explicitly by the mission, then that should overrule the addon.

PabstMirror:
Yeah, I'd say if either one explicitly disables, then it should be disabled
```
private _cfg = configFile >> "EnableTargetDebug";
private _blocked = ((isNumber _cfg) && {(getNumber _cfg) isEqualTo 0}) || {(getMissionConfigValue ["EnableTargetDebug", -999]) isEqualTo 0};
if ((!_blocked) && {1 isEqualTo ([getMissionConfigValue ["EnableTargetDebug", 0], getNumber _cfg] select isNumber _cfg)}) then {
    ...
```

Dystopian:
So nobody answered my question about mission maker purpose to disable Target Debug. I gave 2 arguments for addon priority and nobody said anything against.

About `enableDebugConsole` analogue: as I saw in dev version src a week ago after [T124835](https://feedback.bistudio.com/T124835), it looks like mission value WILL NOT beat addon value. I don't know if it's changed in release version - we'll see in 1.74.

Dorbedo:
There is no good reason for overruling the missionmakers decision. If you have the permission of the missioncreator, you could modify the missionfile very easily. If not, you have to respect the decision.

Also I don't see a good reason to make the behavior different from the vanilla behavior. This makes things much more complicated.
Until the vanilla behavior is not changed, CBA should stick with mission priority over addon.

I must add, that we are talking about a very few cases. I think, there will almost never be someone allowing the debug console, but disallowing the target debugging.

Dystopian:
We have this in vanilla 1.76:
```
_chk = getMissionConfigValue ["enableDebugConsole", 0]; 
if !(_chk isEqualTo 0) exitWith {_chk};
...
_cfg = configFile >> "enableDebugConsole"; 
if (!isNull _cfg) exitWith {_cfg call _fnc_getConfigValue};
```
Mission value has priority only if > 0.

commy2:
BIS_fnc_isDebugConsoleAllowed is a mess now.

Dorbedo:
> Mission value has priority only if > 0.
No, the prio has not been changed. Please read the function properly.

Dystopian:
updates on this?

PR #732, opened 2017-08-02
fixes 7.62 silencers being available for 6.5 guns

TheAussieEnigma:
title


robalo:
asdg_MuzzleSlot_65 extends the 7.62 one and should not be used on guns directly, was only added to fix some vanilla compat issues.
They don't make 6.5 suppressors IRL.

PR #734, opened 2017-08-12
XEH compatibility for 1.76 classes

commy2:
- `A3_Props_F_Argo` is for `Land_Laptop_02_F` and was forgotten the last time...


commy2:
Requires 1.76, don't merge before.

PR #735, opened 2017-08-15
changes for 1.76 debug console

commy2:
- There is a new output line for the exec box. This PR puts it in the right position for the extended debug console.
- Uses the new idc of the LOCAL EXEC button in the script

todo:
- [ ] The TARGET EXEC button does not update the expression output line. The SERVER EXEC button does. @PabstMirror 
- [x] The LOCAL EXEC button no longer updates the 50 expressions log (prev / next). Pressing enter, the other EXEC buttons and the code performance button still work...


commy2:
Could be merged and the TARGET exec button be tweaked later.

PR #737, opened 2017-08-19
remove watch field pause

commy2:
- removes the very frustrating pause when the focus is set on the watch fields input lines, which is a thing now in 1.76


Killswitch00:
Does this work in 1.74 aswell? ("work" could be "has no effect but doesn't break anything 1.74")

commy2:
Yes, would do nothing in 1.74, but also cause no problems.

PR #739, opened 2017-08-28
Fix settings menu slow loading

SilentSpike:
- Change the settings menu so that controls are created only once their associated category is switched to (should drastically reduce menu load time when a lot of settings are present - tested using the [ace conversion branch](https://github.com/acemod/ACE3/tree/convertSettings) where the issue was discovered).

To be honest I totally fumbled my way through the code until I found the parts that created the controls and handled category changing, then just adjusted those to try and implement the desired behaviour. It appears to work, but I can't say for certain whether there are any undesired side effects of this change that I've overlooked so please review thoroughly.


commy2:
Thanks! Will take a look later.

commy2:
This looks good to me. I don't know if I should pity you for having to go through my mess or if I should be proud of myself that you found a solution that only changes a small number of lines.

PR #744, opened 2017-09-08
Add CBA_miscItems to virtual arsenal

PabstMirror:
ACE and ACRE had been setting their items as minedetectors to show up in virtual arsenal.
With Arma 1.76 they now act as real mine detectors.

This PR adds a new base CfgWeapon `CBA_MiscItem` that will be manually added to the "Misc Items" tab in arsenal.

Example use:
```
class CfgWeapons {
    class CBA_MiscItem;
    class CBA_MiscItem_ItemInfo;

    class CBA_BananaA: CBA_MiscItem {
        displayName = "Banana A";
    };
    class CBA_BananaB: CBA_MiscItem {
        displayName = "Banana B";
        scope = 2;
        class ItemInfo: CBA_MiscItem_ItemInfo {
            mass = 2;
        };
    };
};
```


addons/common/XEH_preClientInit.sqf
```diff
@@ -8,4 +8,7 @@ if (hasInterface) then {
     GVAR(actionListUpdated) = false; // Set true to force recreation of actions.
     GVAR(nextActionIndex) = 0; // Next index that will be given out.
     GVAR(actionListPFEH) = false;
+
+    // Add CBA_MiscItems to VirtualArsenal
+    [missionnamespace, "arsenalOpened", {call CBA_fnc_handleArsenalOpened}] call bis_fnc_addscriptedeventhandler;
```
Neviothr: `missionNamespace` camelCase.

Neviothr: Check camelCase capitalization in:
XEH_preClientInit - line 13.
fnc_handleArsenalOpened - lines 13, 19, 21, 30-44, 51-57 and 60.

addons/common/fnc_handleArsenalOpened.sqf
```diff
@@ -0,0 +1,66 @@
...
+    private _class = _x;
+    private _scope = if (isnumber (_class >> "scopeArsenal")) then {getnumber (_class >> "scopeArsenal")} else {getnumber (_class >> "scope")};
+    TRACE_2("",_class,_scope);
+    if (_scope == 2 && {gettext (_class >> "model") != ""}) then {
+        _cbaMiscItems pushBack (configName _class);
+    };
+} forEach (configProperties [configFile >> "CfgWeapons", "(isClass _x) && {(configName _x) isKindOf ['CBA_MiscItem', configFile >> 'CfgWeapons']}"]);
+TRACE_2("Items to add",count _cbaMiscItems,_cbaMiscItems);
+
+
```
Neviothr: Redundant whitespace?

addons/common/fnc_handleArsenalOpened.sqf
```diff
@@ -0,0 +1,66 @@
...
+// Much of this is from fn_arsenal.sqf -> "ListAdd"
+
```
Neviothr: Same?

addons/common/CfgWeapons.hpp
```diff
@@ -0,0 +1,10 @@
+class CfgWeapons {
+    class ItemCore;
+    class InventoryItem_Base_F;
+    class CBA_MiscItem_ItemInfo: InventoryItem_Base_F {
+        type = 302; // "bipod"
+    };
```
commy2: I'm wondering if this class (`CBA_MiscItem_ItemInfo`) is really needed or if the modified entries couldn't just be put in `CBA_MiscItem/ItemInfo`.

addons/common/XEH_preClientInit.sqf
commy2: I'd like this to be camel case.
```sqf
[missionNamespace, "arsenalOpened", {call CBA_fnc_handleArsenalOpened}] call bis_fnc_addScriptedEventHandler;
```

addons/common/fnc_handleArsenalOpened.sqf
```diff
@@ -0,0 +1,66 @@
+#define DEBUG_MODE_FULL
+#include "script_component.hpp"
+
+#define IDC_RSCDISPLAYARSENAL_LIST            960
+#define IDC_RSCDISPLAYARSENAL_TAB_CARGOMISC   24
+#define IDC_RSCDISPLAYARSENAL_SORT            800
```
commy2: These are already part of:
```cpp
#include "\a3\ui_f\hpp\defineResinclDesign.inc"
```
But that file cannot be included in the components script_component.hpp, because it has a double define that crashes the binarizer. It can be included in this file though.

commy2:
It's not really a miscellaneous item, is it? Making that isn't hard. It's a base class that specifically makes the item appear in the Virtual Arsenal. `CBA_Item(Virtual)Arsenal_base`? Then the name would be descriptive.

addons/common/XEH_preClientInit.sqf
commy2:
```sqf
[missionNamespace, "arsenalOpened", {call CBA_fnc_handleArsenalOpened}] call BIS_fnc_addScriptedEventHandler;
```

jonpas:
Technically it's a miscellaneous item that supports Virtual Arsenal, I'd call it "inventory miscellaneous item".

Most of the commented code above is copied from `arsenal.sqf` as @PabstMirror noted, unsure if worth updating, makes it easier to diff with original like that?

commy2:
Agreed, except the line in preInitClient.sqf^^

addons/common/CfgWeapons.hpp
PabstMirror: It makes configs easier IMHO because you just add the 2 parent defines instead of having to add the full inhertiance from the parent.
Also makes it a drop in replacement for our existing configs (just change `InventoryItem_Base_F`):
```
    class ACE_fieldDressing: ACE_ItemCore {
        class ItemInfo: InventoryItem_Base_F {
            mass = 1;
        };
    };
```

PabstMirror:
We hook into arsenal when it calls events via:
```
with missionnamespace do {
    [missionnamespace,"arsenalOpened",[_display,_toggleSpace]] call bis_fnc_callscriptedeventhandler;
};
```
With ACE items our code takes ~15ms to finish and has no problems.
ACE adds 53 items to the list, but also has 1000 hidden dogtags, which explains why this takes some amount of time.
With debuging on, the execution time went up to 100ms and it randomly started to throw script errors.
It seems like it was getting kicked out of `missionNamespace`, which the command `with` says can happen:

https://community.bistudio.com/wiki/with
>when used in a do-construct in scripts with allowed suspension, an unexpected namespace switch could happen (see note below)

Example:
```
missionNameSpace setVariable [QGVAR(arsenalDataModified), true];
TRACE_1("finished",GVAR(arsenalDataModified));
```
Reported:
`TRACE: finished: cba_common_arsenalDataModified=any`

Solution I came up with is to do
```
_this spawn {
    [{
        ...
    }, _this] call CBA_fnc_directCall;
};
```
Which isn't pretty, but does seem to prevent problems with `with` and namespaces.
This should also be safe if ACRE or other addons add thousands more items to filter and process.
I'm not real familiar with this aspect of sqf, if anyone has thoughts on this please let me know.

addons/common/init_addMiscItemsToArsenal.sqf
```diff
@@ -16,66 +16,73 @@ Author:
 
 
 [missionNameSpace, "arsenalOpened", {
```
jonpas: It's `missionNamespace` I believe.

PR #757, opened 2017-09-30
Add: Return random position in a desired direction +- an angle

Vdauphin:
Hello,
Now you can also search for a random position in a desired direction with a define amplitude.

Does it fit CBA ?

- Now you can also search for a random position in a desired direction with a define amplitude of angle (search in direction 100° with +- 25° angle of amplitude).
- `params` has been updated with optional parameter (default value search for a random position in the circle)
- `_direction - 0.5*_amplitude + random _amplitude` give the possibility to search in a direction with +- `_amplitude `
- Respect the [Submitting Content Guidelines](https://github.com/CBATeam/CBA_A3/wiki/Submitting%20content) --> tell me if anything is wrong 

Quick code to test the behavior with an exemple of the result: 
```
for "_i" from 1 to 1000 do { 
     private _direction = 100; 
     private _amplitude = 50; 
     private _pos = (getpos player) getPos [100*sqrt random 1 , _direction - 0.5*_amplitude + random _amplitude]; 
     deleteMarker str(_i); 
     private _marker = createmarker [str(_i), _pos]; 
     systemChat str(_pos); 
     str(_i) setMarkerSize [1,1]; 
     str(_i) setMarkerType "mil_dot"; 
     str(_i) setMarkerColor "ColorBlue"; 
     str(_i) setMarkerShape "ICON"; 
};
```

### DIFF
--- a/addons/common/fnc_randPos.sqf
+++ b/addons/common/fnc_randPos.sqf
@@ -5,8 +5,10 @@ Description:
     A function used to randomize a position around a given center
 
 Parameters:
-    _position - <MARKER, OBJECT, LOCATION, GROUP, TASK or POSITION>
-    _radius   - random Radius <NUMBER>
+    _position  - <MARKER, OBJECT, LOCATION, GROUP, TASK or POSITION>
+    _radius    - random Radius <NUMBER>
+    _direction - randomization direction (optional, default: 0) <NUMBER>
+    _spreading - spreading around direction (optional, default: 360) <NUMBER>
 
 Example:
     (begin example)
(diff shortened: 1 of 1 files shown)


addons/common/fnc_randPos.sqf
```diff
@@ -24,16 +24,18 @@ SCRIPT(randPos);
...
 
-_position = _position getPos [_radius * sqrt random 1, random 360];
+_position = _position getPos [_radius * sqrt random 1, _direction - 0.5*_amplitude + random _amplitude];
 
 if (_doResize) then {
     _position resize 2;
 };
 
-_position
+_position
```
commy2: Newline should stay.

commy2:
>Does it fit CBA ?

Sure.

Is "amplitude" the right word for this? Can't really think of anything. "Spread"?
These parameters should be explained in the functions header.

addons/common/fnc_randPos.sqf
```diff
@@ -5,8 +5,10 @@ Description:
     A function used to randomize a position around a given center
 
 Parameters:
-    _position - <MARKER, OBJECT, LOCATION, GROUP, TASK or POSITION>
-    _radius   - random Radius <NUMBER>
+    _position  - <MARKER, OBJECT, LOCATION, GROUP, TASK or POSITION>
+    _radius    - random Radius <NUMBER>
+    _direction - randomization direction (optional, default: 0) <NUMBER>
+    _spreading - spreading around direction (optional, default: 360) <NUMBER>
```
Killswitch00: "Angle" is the term for this - the angle of the circular arc in which the random position will end up.

Vdauphin: Sound better, I am agree. So header like this is okay ? : 
``` _angle - the angle of the circular arc in which the random position will end up. (optional, default: 360) <NUMBER> ```

Killswitch00: Yes

PR #762, opened 2017-10-16
Fix versioning error on dev branch

PabstMirror:
Dumb error on dev branch and it will probably be fixed before an actual release
but I don't see any harm in changing config now

Using build.py

`getArray (configfile >> "CfgSettings" >> "CBA" >> "Versioning" >> "cba" >> "dependencies" >> "CBA")`
["cba_main",[1,0,0],"(true)"] // Before
["cba_main",[1,0,"0 "],"(true)"] // After

No idea where the quotes are from, causes old version warning.

### DIFF
--- a/addons/main_a3/config.cpp
+++ b/addons/main_a3/config.cpp
@@ -19,7 +19,7 @@ class CfgSettings {
         class Versioning {
             class PREFIX {
                 class Dependencies {
-                    CBA[] = {"cba_main", { 1,0,0 },"(true)"};
+                    CBA[] = {"cba_main", { 1,0,0},"(true)"};
                 };
             };
         };


jonpas:
What does the actual config say in-game? It's not something pboProject magically converts, but game manages to still read it correctly?

PabstMirror:
unbinned pbos from build.py (file patching)
image

but it's fine with make.py / pboProject builds

commy2:
Probably thinks the space after the 1 makes the subarray element a string. Question is if this happens for flat arrays too. If so, then this will break all kinds of things most likely.

addons/main_a3/config.cpp
```diff
@@ -19,7 +19,7 @@ class CfgSettings {
         class Versioning {
             class PREFIX {
                 class Dependencies {
-                    CBA[] = {"cba_main", { 1,0,0 },"(true)"};
+                    CBA[] = {"cba_main", { 1,0,0},"(true)"};
```
Killswitch00: Ugh.... how about `{1, 0, 0}` to make it less hurtful on the eyes?

jonpas: Or `{1,0,0}` so there are no spaces anywhere.

PabstMirror:
Fixed in dev branch (didn't see anything in changelog)

PR #771, opened 2017-10-22
Statemachine cleanup

Neviothr:
- do some clean up in statemachine component.


addons/statemachine/fnc_clockwork.sqf
```diff
@@ -18,7 +18,7 @@ SCRIPT(clockwork);
 
 {
     #ifdef STATEMACHINE_PERFORMANCE_COUNTERS
-    private _perfStartTime = diag_tickTime;
+        private _perfStartTime = diag_tickTime;
```
Killswitch00: `_perfStartTime` belongs to the same scope as `_stateMachine` and `_list` below and should be indented the same amount. Conditional preprocessing should not change the indentation level.

addons/statemachine/fnc_clockwork.sqf
```diff
@@ -98,8 +98,8 @@ SCRIPT(clockwork);
     };
 
     #ifdef STATEMACHINE_PERFORMANCE_COUNTERS
-    private _perfRunTime = diag_tickTime - _perfStartTime;
-    (GVAR(performanceCounters) select _id) pushBack _perfRunTime;
+        private _perfRunTime = diag_tickTime - _perfStartTime;
```
Killswitch00: Same here and in all other places where this is done.

addons/statemachine/fnc_clockwork.sqf
```diff
@@ -103,4 +103,4 @@ SCRIPT(clockwork);
     #endif
 
     false
-} count GVAR(stateMachines);
+} count GVAR(stateMachines);
```
commy2: newline eof

Neviothr:
@commy2 - git claims no newline :confused:

jonpas:
Uhm no, commy added it, red line was removed, green one added. GitHub won't show an empty newline as an actual empty new line, but if there isn't one there will be a red marker.

Neviothr:
I'm :turtle:

commy2:
Sorry. I complained about it in a comment and then just fixed it myself afterwards.

PR #791, opened 2017-10-30
Added some of Portuguese translation

JuhBass:
Added some of Brazilian Portuguese translation

**When merged this pull request will:**
- Describe what this pull request will do
- Each change in a separate line
- Respect the [Submitting Content Guidelines](https://github.com/CBATeam/CBA_A3/wiki/Submitting%20content)

PR #792, opened 2017-10-31
Some Portuguese translation

caiqueff:
- Add some Portuguese translation


commy2:
@caiqueff What about #791?

commy2:
Backlog, because it conflicts with (almost ?) all changes from #791.

commy2:
Dead.

PR #802, opened 2017-11-21
add optional parameter to editbox setting to obscure passwords, close #763

### LINKED ISSUE #763: Request - Settings EditBox text not visible on clients
At ACRE2 we will have TeamSpeak 3 Channel Password in CBA Settings (for auto-move) and it would be nice if clients couldn't just go into the settings dialog and be able to read it.

commy2:
- if you can't edit them, replace them with stars
- closes #763

### DIFF
--- a/addons/settings/fnc_init.sqf
+++ b/addons/settings/fnc_init.sqf
@@ -89,6 +89,7 @@ switch (toUpper _settingType) do {
     };
     case "EDITBOX": {
         _defaultValue = _valueInfo param [0, "", [""]]; // don't use params - we want these variables to be private to the main scope
+        _settingData pushBack (_valueInfo param [1, false]);
     };
     case "LIST": {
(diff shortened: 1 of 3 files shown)


addons/settings/fnc_init.sqf
```diff
@@ -89,6 +89,7 @@ switch (toUpper _settingType) do {
     };
     case "EDITBOX": {
         _defaultValue = _valueInfo param [0, "", [""]]; // don't use params - we want these variables to be private to the main scope
+        _settingData pushBack (_valueInfo param [1, false]);
```
PabstMirror: Just to be safe, maybe add type checking `param [1, false, [false]]`

commy2: I already have `isEqualTo true` check in the place where I read this one.

PR #806, opened 2017-11-23
Add CBA_loadingScreenDone event

commy2:
- title

I sometimes get asked if there is a way to execute code when Arma is done with the loading screen at the start of the mission. This PR adds a cba event for that by hooking into BI's internal framework for loading screens (`BIS_fnc_start/endLoadingScreen`).

There are up to 3 loading screens that happen at the start of the mission. `bis_fnc_preload`, `bis_fnc_initRespawn` and `bis_fnc_initFunctions`, named after the scripts that initiates them.
`bis_fnc_preload` is started before even preInit runs and it essentially makes clients wait for the server to load the mission first. It waits for a dummy public global variable to be synched.
`bis_fnc_initRespawn` sets up some kind of respawn screen, which is done even in SP surprisingly, and is for some insane reason done in scheduled environment (hence the need for a loading screen I guess).
`bis_fnc_initFunctions` makes the client wait for all postInit functions to be finished. But it only appears on dedicated clients, which is a very strange choice, because it means that someone could make an eternal loop in postInit stalling all other postInit scripts forever in SP. But it also means that the mission still ends the loading screen in SP if a script error occured which is helpful for debugging in some way - no need to kill the game (reminder that postInit CfgFunctions is scheduled).
There is no way to tell which of these finishes first, so we just wait for all 3.

During the mission there can be many more of course, but this event is only supposed to fire once. Perfect for mission intros etc.

I can't decide if the demand is there or if it would be bloat.

Note that the debug beep sound seems to happen right before the loading screen disappears. This is because the `endLoadingScreen` command BI uses makes it fade out slowly. The event is triggered when the loading screen starts fading.

WIP, because it has debug enabled.

### DIFF
--- a/addons/xeh/CfgFunctions.hpp
+++ b/addons/xeh/CfgFunctions.hpp
@@ -66,5 +66,13 @@ class CfgFunctions {
                 file = PATHTOF(fnc_initDisplay.sqf);
             };
         };
+        class Misc {
+            class startLoadingScreen {
+                file = PATHTOF(fnc_startLoadingScreen.sqf);
+            };
+            class endLoadingScreen {
+                file = PATHTOF(fnc_endLoadingScreen.sqf);
+            };
+        };
     };
 };
--- a/addons/xeh/fnc_endLoadingScreen.sqf
+++ b/addons/xeh/fnc_endLoadingScreen.sqf
@@ -5,7 +5,7 @@ private _return = call {
 };
 
 isNil {
-    ["CBA_LoadingScreenEnded", _this] call CBA_fnc_localEvent;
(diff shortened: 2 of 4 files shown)


commy2:
oops, misclicked. Still not done with my explanation in the OP, but w/e.

Cuel:
This is very useful.

commy2:
Updated OP with my findings.

Cyruz143:
I'd make use of this @commy2, good for doing text overlays at the start of the mission etc. Sometimes they'll fire before the client can actually see them!

commy2:
Usage:
```sqf
// init.sqf
["CBA_loadingScreenDone", {
    // blah
```

PR #811, opened 2017-11-30
fix v1.78 enum popup on xeh incompatible objects by fallback loop

commy2:
- Adding events that don't exist doesn't matter, except that it does now by creating that enum error. This PR makes it so we don't.

### DIFF
--- a/addons/xeh/fnc_initEvents.sqf
+++ b/addons/xeh/fnc_initEvents.sqf
@@ -36,7 +36,7 @@ if !(ISPROCESSED(_unit)) then {
     if (!isClass _eventClass) then {
         {
             _unit addEventHandler [_x, format ['call FUNC(%1)', _x]];
-        } forEach [XEH_EVENTS];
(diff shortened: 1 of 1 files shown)


commy2:
```
16:59:26 Error in expression <f (!isClass _eventClass) then {
{
_unit addEventHandler [_x, format ['call cba_x>
16:59:26   Error position: <addEventHandler [_x, format ['call cba_x>
16:59:26   Error Foreign error: Unknown enum value: "FiredBis"
16:59:26 File \x\cba\addons\xeh\fnc_initEvents.sqf [CBA_fnc_initEvents], line 1946
16:59:26 Error in expression <f (!isClass _eventClass) then {
{
_unit addEventHandler [_x, format ['call cba_x>
16:59:26   Error position: <addEventHandler [_x, format ['call cba_x>
16:59:26   Error Foreign error: Unknown enum value: "InitPost"
16:59:26 File \x\cba\addons\xeh\fnc_initEvents.sqf [CBA_fnc_initEvents], line 1946
```

PR #813, opened 2017-12-02
pause game in SP when Addon Options menu is shown

commy2:
- title

Noticed this while debugging a parachute related setting for ACE. The vanilla options menu pauses the game, while ours does not (it even resumes, because the esc menu itself pauses it).


jonpas:
Milestone?

commy2:
3.5.1 or 3.6 as long as no other issue with RC2 is found.

PR #814, opened 2017-12-03
fix issue with translated setting categories

commy2:
- fixes:

Bad logic broke settings in categories using stringkeys. The category would be translated **in the wrong place**, which causes it to appear untranslated in the ui and the setting would not be recognized either.


jonpas:
3.5.0 or 3.5.1?

commy2:
3.5.1 or 3.6 as long as no other issue with RC2 is found.

:grin:

PR #816, opened 2017-12-05
fix setting direction of remote units via zeus

commy2:
- title

It has been years. They will never fix this. This work around is very simple.

### DIFF
--- a/addons/common/XEH_preInit.sqf
+++ b/addons/common/XEH_preInit.sqf
@@ -37,4 +37,22 @@ GVAR(addons) = _addons;
 // BWC
 #include "backwards_comp.sqf"
 
+// fix changing direction of remote units not working with zeus
+["ModuleCurator_F", "init", {
+    params ["_logic"];
+
+    _logic addEventHandler ["CuratorObjectEdited", {
+        params ["", "_unit"];
+
+        if (!local _unit) then {
+            private _dir = getDir _unit;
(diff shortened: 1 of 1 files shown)


PabstMirror:
maybe add 
`if (missionNamespace getVariable [QGVAR(disableZeusDirFix), false]) exitWith {};`
Just in case this causes problems or is undesired?

commy2:
We could store the ehid in a object namespace variable on the logic.

addons/common/XEH_preInit.sqf
```diff
@@ -37,4 +37,22 @@ GVAR(addons) = _addons;
...
+// fix changing direction of remote units not working with zeus
+["ModuleCurator_F", "init", {
+    params ["_logic"];
+
+    _logic addEventHandler ["CuratorObjectEdited", {
+        params ["", "_unit"];
+
+        if (!local _unit) then {
+            private _dir = getDir _unit;
+            [_unit, _dir] remoteExec ["setDir", _unit];
```
shukari: With doWatch you can also let tanks turn the gun. Would be nice to have ;D

Cuel: The gun turns when you rotate them anyway?

commy2: The turret is rotating fine with this fix without having to do doWatch.

PR #823, opened 2017-12-07
remove obsolete ui_helper component

commy2:
- remove a component that does nothing in A3

This component was meant for TKOH to work around some kind of initialization problem. Since CBA_A3 is not compatible with TKOH, this can be removed.
Otherwise it's used by the fleximenu to #include the dik codes. That's not needed either, because they can just be included directly.


PabstMirror:
If we do drop a PBO, we need to be very clear in the next update that people need to delete or there will be old pbo key problems

commy2:
Steam does delete it correctly, or not?

PabstMirror:
I think so

commy2:
Then a warning with the changelog / release notes should be good.

addons/ui/config.cpp
```diff
@@ -11,6 +11,9 @@ class CfgPatches {
         version = VERSION;
         authors[] = {"Dr Eyeball", "commy2"};
     };
+
+    // Backwards compatibility
+    class cba_ui_helper: ADDON {author = "";};
```
jonpas: Why not just inherit author? It's set to CBA anyways.

commy2: This prevents the credits line stuff from the help component to show up this removed component from time to time. Same is done for other bwc CfgPatches entries like the four aliases of XEH.

jonpas: Oh I see, time to do that in ACRE... :P
Thanks!

PR #828, opened 2017-12-09
add userXX to keybinding menu

### LINKED ISSUE #825: Ability to bind to Joysticks
Is there a possibility that we are able to bind different controls to buttons/switches on a HOTAS (for me its the Thrustmaster Warthog), this is mainly for custom eject scripts that we are using along with the ability to bind TFAR Radio controls to it.

Not sure if the Arma 3 framework would allow this to work, or if it has been implemented and I am just having a Dumb Dumb moment (or a couple of dozen)

commy2:
### Documentation:
https://github.com/CBATeam/CBA_A3/wiki/Bind-Addon-Key-Actions-to-Joystick-Buttons

- close #825

This would kind of allow joystick buttons to be usable with CBA keybinding.
The idea is to poll the `inputAction` state of all 20 "User Action" keys and to trigger our scripted keyhandler event with special "dik codes" when they change. The special "dik codes" are then also added to the drag and drop part of the keybinding menu.


commy2:
```sqf
["Debug Category", "DebugKey", "Debug Key", {
    systemChat str true;
}, {
    systemChat str false;
}, [0, [false, false, false]]] call CBA_fnc_addKeybind;
```

addons/events/fnc_userKeyHandler.sqf
```diff
@@ -9,7 +9,7 @@ Author:
 ---------------------------------------------------------------------------- */
 #include "script_component.hpp"
 
-if (!GVAR(checkUserActions)) exitWith {};
+if (GVAR(skipCheckingUserActions)) exitWith {};
```
commy2: lol

commy2: @jonpas Check the commit info. Can't comment on commits anymore.

jonpas: Oh, wow.

PR #829, opened 2017-12-09
add weapon eventhandler framework

### LINKED ISSUE #82: Integrate Weapon Eventhandler Framework
Add Weapon Eventhandler Framework 
Re: http://forums.bistudio.com/showthread.php?192780-Weapon-Eventhandler-Framework

commy2:
### Documentation:
https://github.com/CBATeam/CBA_A3/wiki/Weapon-Animations-Framework

- close #82

Helped a guy on discord to implement this the other day, so might as well make it a function in CBA.

### DIFF
--- a/addons/events/fnc_weaponEvents.sqf
+++ b/addons/events/fnc_weaponEvents.sqf
@@ -51,7 +51,7 @@ private _fnc_soundSource = {
     _soundSource
 };
 
-if (isEmpty) then {
+if (_isEmpty) then {
     private _sound = getText (_config >> "soundEmpty");
     private _soundLocation = getText (_config >> "soundLocationEmpty");
 
--- a/addons/events/fnc_weaponEvents.sqf
+++ b/addons/events/fnc_weaponEvents.sqf
@@ -42,10 +42,11 @@ private _fnc_soundSource = {
     private _soundSourceName = format [QGVAR(soundSource_%1), _soundLocation];
     private _soundSource = _unit getVariable [_soundSourceName, objNull];
 
+    if (isNull _soundSource) then {
+        _soundSource = "#particlesource" createVehicleLocal [0,0,0];
+    };
+
     if !(_soundSource in attachedObjects _unit) then {
-        if (isNull _soundSource) then {
-            _soundSource = "#particlesource" createVehicleLocal [0,0,0];
-        };
         _soundSource attachTo [_unit, [0,0,0], _soundLocation];
         _unit setVariable [_soundSourceName, _soundSource];
     };
--- a/addons/events/fnc_weaponEvents.sqf
+++ b/addons/events/fnc_weaponEvents.sqf
@@ -68,8 +68,13 @@ if (!_isEmpty || _onEmpty) then {
     private _soundLocation = getText (_config >> "soundLocation");
     private _delay = getNumber (_config >> "delay");
 
+    private _expectedMagazineCount = count magazines _unit;
+
     [{
-        params ["_unit", "_handAction", "_sound", "_soundSource"];
+        params ["_unit", "_handAction", "_sound", "_soundSource", "_expectedMagazineCount"];
+
+        // exit if unit started reloading
+        if (count magazines _unit != _expectedMagazineCount) exitWith {};
 
         if (local _unit) then {
             _unit playAction _handAction;
@@ -78,5 +83,5 @@ if (!_isEmpty || _onEmpty) then {
         if (_sound != "") then {
             _soundSource say3D _sound;
         };
-    }, [_unit, _handAction, _sound, call _fnc_soundSource], _delay] call CBA_fnc_waitAndExecute;
+    }, [_unit, _handAction, _sound, call _fnc_soundSource, _expectedMagazineCount], _delay] call CBA_fnc_waitAndExecute;
 };
--- a/addons/events/fnc_weaponEvents.sqf
+++ b/addons/events/fnc_weaponEvents.sqf
@@ -43,7 +43,9 @@ private _fnc_soundSource = {
     private _soundSource = _unit getVariable [_soundSourceName, objNull];
 
     if !(_soundSource in attachedObjects _unit) then {
-        _soundSource = "#particlesource" createVehicleLocal [0,0,0];
+        if (isNull _soundSource) then {
+            _soundSource = "#particlesource" createVehicleLocal [0,0,0];
+        };
         _soundSource attachTo [_unit, [0,0,0], _soundLocation];
         _unit setVariable [_soundSourceName, _soundSource];
     };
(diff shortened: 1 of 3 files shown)


addons/events/fnc_weaponEvents.sqf
```diff
@@ -0,0 +1,80 @@
...
+    if !(_soundSource in attachedObjects _unit) then {
+        _soundSource = "Building" createVehicleLocal [0,0,0];
+        _soundSource attachTo [_unit, [0,0,0], _soundLocation];
+        _unit setVariable [_soundSourceName, _soundSource];
+    };
+
+    _soundSource
+};
+
+if (isEmpty) then {
```
jokoho48: missing _

addons/events/fnc_weaponEvents.sqf
```diff
@@ -0,0 +1,80 @@
...
+if (isNumber (_config >> "onEmpty")) then {
+    _onEmpty = getNumber (_config >> "onEmpty") == 1;
+};
+
+private _fnc_soundSource = {
+    private _soundSourceName = format [QGVAR(soundSource_%1), _soundLocation];
+    private _soundSource = _unit getVariable [_soundSourceName, objNull];
+
+    if !(_soundSource in attachedObjects _unit) then {
+        _soundSource = "Building" createVehicleLocal [0,0,0];
```
jokoho48: using #particlesource is faster
```Result:
0.0093 ms

Cycles:
10000/10000

Code:
JK_test pushBack ("#particlesource" createVehicleLocal [0,0,0]);
```
VS
```
Result:
0.25208 ms

Cycles:
3967/10000

Code:
JK_test pushBack ("Building" createVehicleLocal [0,0,0]);```

File: addons/events/fnc_weaponEvents.sqf [hunk 3]
```diff
@@ -0,0 +1,80 @@
...
+
+    [{
+        params ["_unit", "_handAction", "_sound", "_soundSource"];
+
+        if (local _unit) then {
+            _unit playAction _handAction;
+        };
+
+        if (_sound != "") then {
+            _soundSource say3D _sound;
```
jokoho48: the sound object dont get delete anywhere so they would stack up. or did i miss something?

addons/events/fnc_weaponEvents.sqf
commy2: How long does it live?

addons/events/fnc_weaponEvents.sqf
commy2: No point in deleting it, if you can recycle it instead.

commy2: See L45

addons/events/fnc_weaponEvents.sqf
jokoho48: until you delete it

addons/events/fnc_weaponEvents.sqf
jokoho48: aha i oversaw that sorry :D

addons/events/fnc_weaponEvents.sqf
commy2: Perfect, thanks!

addons/events/fnc_weaponEvents.sqf
jokoho48: but maybe delete it when the unit dies? because else when some units would use that and die they would eather way stack up.

addons/events/fnc_weaponEvents.sqf
commy2: Any reason to not use `#soundsource`? You know, because it's supposed to be a soundsource?

addons/events/fnc_weaponEvents.sqf
commy2: Question is if the gain would be worth the effort. I have my doubts.

jokoho48: i think it would. else you could take a look at what i did within the ammo cookoff. Object Pooling would be the perfect thing for this i think.

addons/events/fnc_weaponEvents.sqf
jokoho48:
```
Result:
0.0087 ms

Cycles:
10000/10000

Code:
JK_test pushBack ("#particlesource" createVehicleLocal [0,0,0]); 
```
vs 
```
Result:
0.135655 ms

Cycles:
7379/10000

Code:
JK_test pushBack ("#soundsource" createVehicleLocal [0,0,0]); 
```

commy2: Nvm, seems like those don't work with `createVehicleLocal`.

jokoho48: what does not work with createVehicleLocal?

addons/events/fnc_weaponEvents.sqf
```diff
@@ -0,0 +1,80 @@
...
+
+params ["_unit", "_weapon"];
+
+private _config = configFile >> "CfgWeapons" >> _weapon >> "CBA_WeaponEvents";
+
+private _isEmpty = _unit ammo _weapon == 0;
+private _onEmpty = true;
+
+if (isNumber (_config >> "onEmpty")) then {
+    _onEmpty = getNumber (_config >> "onEmpty") == 1;
```
PabstMirror: default returns 0, so can probably skip `isNumber`?

commy2: But I want 1/true to be the default.

Dystopian: why not `CBA_fnc_getConfigEntry`?

commy2: Overhead.

addons/events/fnc_weaponEvents.sqf
commy2: I changed it to reuse the sound source from the corpse. But I still don't want to delete them after use.

commy2:
Doc: https://github.com/CBATeam/CBA_A3/wiki/Weapon-Animations-Framework

commy2:
For `handAction`s with delay, this ^ prevents the action from interrupting the reload gesture, destroying the magazine.

commy2:
^ a null object is attached to the player in the arsenal, so we have to make sure we don't check `objNull in [<null>]`.

commy2:
>Ok, so don't reload round as long as you look down sights.

commy2:
- [ ] hasOptic config entry

madpat3:
why not make it a standart animation for the "M320 LRR .408"?

Dystopian:
Not CBA goal I think. ACE3 could do it.

madpat3:
ah, i see.

PR #830, opened 2017-12-09
Don't check inputAction if we have none bound

PabstMirror:
For #828 

Checking `cba_events_fnc_userKeyHandler` isn't that bad, but it still is 20x calls to inputAction
This skips those checks if we don't have any keys bound to userActions and makes that func call about 20 times faster


commy2:
You can remove unbound keys or bind them twice and it just fails silently, so I don't trust just counting them. This checks all 20 is one of them was added or removed.

commy2:
Merge was into the other PR.

PR #832, opened 2017-12-10
add dialog to export and import settings text files

commy2:
- fixes a bug where settings are not displayed correctly in ui after  #"import"
- `copyFrom/ToClipboard` has stupid limitations in MP, so let the user press Ctrl+C/V on a edit box instead

It scrolls for more than 40 settings and can display thousands.


Dorbedo:
I think, it would be an advantage, if the settings would be exported sorted by category. Maybe even with a comment.
``` 
// category 1
setting_category1_A = 1;
force setting_category1_B = 2;
// category 2
setting_category2_A = ABC;
```
I recognized this feature as missing while exporting the ace-settings

Dystopian:
it would be cool if only changed settings could be exported

commy2:
^ fixes https://www.youtube.com/watch?v=F-whgpGviRU&feature=youtu.be

commy2:
^ sorted by category

commy2:
Done and done.

addons/settings/fnc_export.sqf
```diff
@@ -15,9 +15,10 @@ Author:
 ---------------------------------------------------------------------------- */
 #include "script_component.hpp"
 
-params [["_source", "client", [""]]];
+params [["_source", "client", [""]], ["_skipDefault", false, [false]]];
```
Dystopian: add `_skipDefault` to header?

Dystopian:
Tested, works great.
If have few non-default settings and uncheck `Show default values`, comments are really superfluous. It would be good to have checkbox to disable comments (and maybe double newline between categories).

commy2:
I think the comments are great and a checkbox just to disable them feels like "too much".

PR #833, opened 2017-12-14
settings select client tab as logged admin by default

commy2:
- title

If you're a logged in admin on a dedicated server, the client tab will be selected by default, not the server tab. This is so that @cyruz doesn't mess with everyone's view distance by accident mid mission.


Cyruz143:
Thanks commy!

PR #834, opened 2017-12-15
Fix sliders for whole numbers

PabstMirror:
`[3.14, 1, -2] call cba_fnc_formatNumber` = "003"
Could also fix in formatNumber?
Or we could switch to `toFixed` (`3.14 toFixed -1` = "3")

### DIFF
--- a/addons/settings/fnc_gui_settingSlider.sqf
+++ b/addons/settings/fnc_gui_settingSlider.sqf
@@ -19,7 +19,7 @@ _ctrlSlider ctrlAddEventHandler ["SliderPosChanged", {
 
     private _controlsGroup = ctrlParentControlsGroup _ctrlSlider;
     private _ctrlSliderEdit = _controlsGroup controlsGroupCtrl IDC_SETTING_SLIDER_EDIT;
-    _ctrlSliderEdit ctrlSetText ([_value, 1, _trailingDecimals] call CBA_fnc_formatNumber);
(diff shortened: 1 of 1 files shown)


commy2:
-2 meaning 3 forced digits with leading zeros is intended by CBA_fnc_format.
Why does the setting use _trailingDecimals = -2 in the first place?

PabstMirror:
From https://github.com/CBATeam/CBA_A3/pull/510

FormatNumber:
```
_decimalPlaces - Number of decimal places, padded with trailing 0s,
        if necessary [Number: >= 0, defaults to 0]
```
I guess the call would be `[3.14, 1, -1] call cba_fnc_formatNumber` but that's also "003"

commy2:
Is the -1 supposed to be "ignore this"?

>[3.14, 1, -1] call cba_fnc_formatNumber [is] also "003"

That's a bug then.

PabstMirror:
-1 is for whole numbers only,
but passing -1 to format number is undefined, not really an error

I think we should just do the `_trailingDecimals max 0` as that fixes the output

commy2:
>-1 is for whole numbers only,

But that is already what 0 does. Zero places after the decimal dot.

PabstMirror:
See https://github.com/CBATeam/CBA_A3/pull/510
0 doesn't display decimal places, but they still exist in the final setting.

commy2:
I get it now.

PR #843, opened 2017-12-22
add description.ext entry to enable debug console in SP

commy2:
- Adds `enableDebugConsoleSP` mission config entry to work around the broken isDebugConsoleEnabled script always reporting false in single player
- Used in https://github.com/acemod/ACE3/pull/5953


commy2:
^ this commit is just from a different PR that I forked from this one, oops.

PR #851, opened 2017-12-26
Fixed missing waypointGarrison CfgFunctions entry

dedmen:
CBA_fnc_waypointGarrison was missing in cfgFunctions.

https://forums.bistudio.com/forums/topic/168277-cba-community-base-addons-arma-3/?page=46&tab=comments#comment-3258805

I don't see a reason to hide it from normal use as a function and have it be waypoint exclusive.

### DIFF
--- a/addons/ai/CfgFunctions.hpp
+++ b/addons/ai/CfgFunctions.hpp
@@ -8,6 +8,7 @@ class CfgFunctions {
             PATHTO_FNC(taskDefend);
             PATHTO_FNC(taskPatrol);
             PATHTO_FNC(taskSearchArea);
+            PATHTO_FNC(waypointGarrison);
         };
     };
 };


commy2:
Don't like the idea. Why not use:
```sqf
_group addWaypoint [_position, 0] setWaypointScript "\x\cba\addons\ai\fnc_waypointGarrison.sqf []";
```
or:
```sqf
[_group, _position] call compile preprocessFileLineNumbers "\x\cba\addons\ai\fnc_waypointGarrison.sqf";
```
This could be added to the function header though.

dedmen:
Ask the guy on BIF asking why it doesn't work

commy2:
Guess I'll have to sign up with a new account there then.

commy2:
>I also don't see why it would hurt to have it accessible.

It's not meant to be a function. It wasn't designed as one. And therefore it will have flaws when used in that environment. That's very general, but it also shows in practice in this function. It isn't actually one, because it requires suspension (waitUntil). It's named "fnc_" for consistency with the BI stuff.

SilentSpike:
Shouldn't the garrison function just be making use of the taskDefend function anyway?

commy2:
It can't. The taskDefend is an actual function that does not need to be spawned. This one is a script that has to return true suspended once all troops reached their destinations, so that the next waypoint could be started.

PR #855, opened 2017-12-26
Settings - Clarify logging for userconfig loading

PabstMirror:


commy2:
Very good! Testing...

addons/settings/fnc_initDisplayMain.sqf
```diff
@@ -29,8 +28,13 @@ if (_file != "") then {
     };
 
     if (_fileExists) then {
+        INFO_1("Userconfig: Loading file [%1]",_file);
         _userconfig = preprocessFile _file;
+    } else {
+        INFO_1("Userconfig: Could not open file [%1]",_file);
```
commy2: Works, but technically the file could also be just empty. Probably not worth mentioning it.

jonpas: `Could not open file [%1] (empty or <other reason>)`?

commy2: File not found or empty.
This also reminds me that the log comments should have periods at the end imo.

PR #859, opened 2017-12-27
read mission settings file every menu load

commy2:
- in 3den, when opening the settings menu, read the settings file again and recreate the missionConfig
- refresh all mission scope settings using the new missionConfig
- clarified some logging akin to Pabst's PR
- simplified a check and mute some excessive logging

### DIFF
--- a/addons/settings/fnc_gui_export.sqf
+++ b/addons/settings/fnc_gui_export.sqf
@@ -14,7 +14,7 @@ private _ctrlToggleDefaultText = _display displayCtrl IDC_EXPORT_TOGGLE_DEFAULT_
 
 // --- scripted buttons
 if (_mode == "import") then {
-    _ctrlTitle ctrlSetText localize LSTRING(ButtonImport);
+    _ctrlTitle ctrlSetText format ["%1 (%2)", localize LSTRING(ButtonImport), localize STR_SOURCE];
 
     _ctrlToggleDefault ctrlEnable false;
     _ctrlToggleDefault ctrlShow false;
@@ -31,7 +31,7 @@ if (_mode == "import") then {
         _display closeDisplay IDC_OK;
     }];
 } else {
-    _ctrlTitle ctrlSetText localize LSTRING(ButtonExport);
+    _ctrlTitle ctrlSetText format ["%1 (%2)", localize LSTRING(ButtonExport), localize STR_SOURCE];
 
     _ctrlToggleDefault cbSetChecked (uiNamespace getVariable [QGVAR(showDefault), true]);
     _ctrlToggleDefault ctrlAddEventHandler ["CheckedChanged", {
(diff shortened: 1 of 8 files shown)


commy2:
Done and works with ACE.

bux:
Does it only read a mission's setting file or the one in userconfig too (e.g. when using/testing a local dedicated server)?

commy2:
Userconfig would need a server restart. You could load the file if one was present at the first start I think, but removing it or checking it when there wasn't one at game start makes it crash to the mission select screen.

commy2:
I like the idea though. Will look into it later this night.

bux:
Please, not the one from userconfig.

jonpas:
Why not?

bux:
>  (e.g. when testing a local dedicated server)?

One wouldn't want to double check each time that the userconfig file is removed when building missions.

jonpas:
I am confused, it's read anyways on first load.

commy2:
I'm confused too. It's not really needed, so I guess we should leave it until someone comes up with a good reason why one should have this. Servers are restarted fairly quickly.

bux:
I thought this PR is about the settings when editing a mission.

My concern is: If one is working with a local dedi server (be it for dev or testing purposes) one might choose to use a "cba_settings.sqf" file in the user config instead of saving settings in the profile. This file could hold the weirdest combinations of settings (for dev or testing purposes).
Now imagine this person is trying to build a mission and saving settings just for this mission. Without knowing in detail how cba settings are implemented, my concern is that those "weird" settings could be taken into account when opening the addon options menu and exporting them and making them wrongly persist in a mission settings file.

commy2:
I don't think that's possible, bux.

bux:
what is possible?

commy2:
That it exports or stores as preset garbage settings.

bux:
These are my assumptions:
1. A "userconfig/cba_settings.sqf" file gets read when starting the game
2. The values stored in the "userconfig/cba_settings.sqf" file get applied when opening the addon options
3. The settings get merged with a "mission/cba_settings.sqf" file (if available)
4. When exporting the settings of the "userconfig/cba_settings.sqf" are ignored (based on your comment https://github.com/CBATeam/CBA_A3/pull/859#issuecomment-354264242)

commy2:
Whatever gets exported depends on the selected source (SERVER, MISSION, CLIENT), and a "broken" (wrong setting values) server config will not change the mission settings export.

bux:
Aaah, this clarifies A LOT!
maybe add the source to the export title bar?

commy2:
Sure, that sounds helpful!

commy2:
Done.

commy2:
After considering how to implement this, I decided not to reload the userconfig. In SP this could still work, but in MP you'd have to:
- make the admin load the file on the server, and therefore delay the menu until the new config (global namespace) with all variables is synched (dedicated server only)
- send around a rather huge file to everyone every time the menu is opened by at least the admin

And I don't think that is worth it.

PR #862, opened 2017-12-29
make starting a mission reload the userconfig settings file

commy2:
- Same as #859, but for the userconfig, and (only) every mission restart


commy2:
Remind myself that this doesn't need to be on the changelog, because it's a PR into another PR.

PR #866, opened 2017-12-31
Add optional mod to skip missing Mod check

dedmen:
Provides the option to disable CBA versioning's missing Mod check which would alert everyone on clientside if client is missing a Mod that uses CBA versioning.

Allows Servers to use CBA (PFH, local events, stuff) in serverside only scripts without annoying each client with a useless message.


addons/versioning/XEH_postInitServer.sqf
```diff
@@ -20,6 +20,9 @@ FUNC(paranoid) = {
 
 QGVAR(versions_serv) addPublicVariableEventHandler { (_this select 1) call FUNC(paranoid) };
 
+// Skip missing mod check if it is disabled.
+if (isNumber (configFile >> "CBA_skipMissingModCheck") && {getNumber (configFile >> "CBA_skipMissingModCheck") == 1}) exitWith {};
```
commy2: Why the `isNumber` check?

dedmen: Not sure if the getNumber could cause errors. But as I am writing this right now I remember the talk from a couple days ago. Fix incoming

optionals/no_missing_mod_check/config.cpp
```diff
@@ -0,0 +1,17 @@
...
+
+class CfgPatches {
+    class ADDON {
+        author = "$STR_CBA_Author";
+        name = ECSTRING(Optional,Component);
+        url = "$STR_CBA_URL";
+        units[] = {};
+        weapons[] = {};
+        requiredVersion = REQUIRED_VERSION;
+        requiredAddons[] = {"CBA_Extended_EventHandlers", "CBA_Versioning"};
```
jonpas: Should be lower-case really.

optionals/no_missing_mod_check/script_component.hpp
```diff
@@ -0,0 +1,16 @@
...
+#ifdef DEBUG_ENABLED_CACHE_DISABLE
+	#define DEBUG_MODE_FULL
+#endif
+
+#ifdef DEBUG_SETTINGS_CACHE_DISABLE
+	#define DEBUG_SETTINGS DEBUG_SETTINGS_CACHE_DISABLE
+#endif
+
+#undef REQUIRED_VERSION
+#define REQUIRED_VERSION 1.00
```
jonpas: Pointless, this optional requires 2 other components which have required version set from `main\script_mod.hpp` anyways.

optionals/no_missing_mod_check/script_component.hpp
```diff
@@ -0,0 +1,16 @@
+#define COMPONENT no_missing_mod_check
+#include "\x\cba\addons\main\script_mod.hpp"
+
+
+#ifdef DEBUG_ENABLED_CACHE_DISABLE
```
jonpas: Should be `#ifdef DEBUG_ENABLED_NO_MISSING_MOD_CHECK`.

optionals/no_missing_mod_check/script_component.hpp
```diff
@@ -0,0 +1,16 @@
...
+	#define DEBUG_MODE_FULL
+#endif
+
+#ifdef DEBUG_SETTINGS_CACHE_DISABLE
```
jonpas: Should be `#ifdef DEBUG_ENABLED_NO_MISSING_MOD_CHECK`.

optionals/no_missing_mod_check/script_component.hpp
```diff
@@ -0,0 +1,16 @@
...
+	#define DEBUG_SETTINGS DEBUG_SETTINGS_CACHE_DISABLE
```
jonpas: Should be `#define DEBUG_SETTINGS DEBUG_SETTINGS_NO_MISSING_MOD_CHECK`.

optionals/no_missing_mod_check/config.cpp
dedmen: Copied from cache disable

optionals/no_missing_mod_check/script_component.hpp
dedmen: Same for cache disable then

optionals/no_missing_mod_check/config.cpp
jonpas: I figured, still wrong, components are all lower-case.

dedmen: Uppercase CBA looks good to me tho. But the uppercase Versioning already looked weird when I wrote it.

jonpas: But it's still wrong, actual component is `cba_versioning`. ¯\\\_(ツ)_/¯

dedmen: https://github.com/CBATeam/CBA_A3/blob/master/addons/settings/config.cpp#L11 consistency then

jonpas: I guess, bet there is a component that doesn't respect it. 😆

dedmen: https://github.com/CBATeam/CBA_A3/blob/master/addons/main/config.cpp#L13 *cough*

commy2:
I think the name `missing_mod_check` is weird. It's atm called "versioning".
Disabling it completely is a bit odd. Someone could still want it, but only ignore one certain addon.
Btw. which addon are we talking about? Because unlike ACE_checkPBO, this only works for addons that have it enabled via config. Shouldn't that optional addon be disabled instead?

dedmen:
I don't think I disable versioning completly. I only disable the missing mod check.
I basically want CBA on the Server without telling clients they are missing it.
So I guess this is about disabling the check for all of CBA.

commy2:
Related: #871

jonpas:
I still don't see why it has to be an optional PBO, when it could be just a config entry which people can define in `description.ext` as well. Or a setting itself as @commy2 suggested on Slack.

In short it should literally be identical to `enableTargetDebug` (which could also be a setting).

dedmen:
So what now? CBA setting or config? Or both?
I'd say CBA setting is enough. But it would be a serverside only setting. And would that be the first CBA setting in CBA? :D

dedmen:
A CBA setting for this would be "confusing" because it's only relevant on serverside. Meaning it shouldn't show up on client/mission settings which afaik isn't supported in CBA settings currently.

Also to enable the setting you would then either need a serverside userconfig for the settings or you'd have to connect with CBA enabled on clientside and then configure it. Which is a chore on a vanilla server with bisigning enabled.

Jonpas to do it the same way as `enableTargetDebug` which is a mission specific attribute.
I don't think a thing that is hidden on the server-backend should be a mission attribute that you have to set in every mission.
My concern is server-side only mods that require CBA. You shouldn't have to edit every of your missions just to run a server-side mode that might have nothing to do with the mission.

I'm still for optional-pbo. In my case I want this for server-side Intercept. And as all users already need a PBO with config to register their plugin they could also add the config entry into their pbo instead of loading the CBA optional.
The optional PBO also shows everyone that, that feature exists. And people might also want to use that outside of just serverside intercept and without having to build their own pbo.

I guess it wouldn't hurt to make it configFile or missionConfigFile. Just to let all options open.

PR #869, opened 2018-01-07
Add sub-categories to settings

PabstMirror:
Adds sorted sub-categories to base categories
ACE has a lot of settings, we don't want to create more base categories than needed.
This will allow better organization while inside of a single category.

cat

(code/ui might be a little rough, mainly just looking for feedback on the idea)
From https://github.com/acemod/ACE3/pull/3563#issuecomment-193916539
We could even attempt to autogenerate these sub-categories for setting names that follow a standard `GVAR(` pattern, but I'm leaning towards leaving it to the mod-makers to fill.


bux:
Good idea 👍 

Could you add some space above each sub category? This would help to visually separate each sub category.

SilentSpike:
I like it yeah, would agree that there should be a line's height above the title line 👍

PR #872, opened 2018-01-11
Add support for arbitrary character trimming

SilentSpike:
- Add support for arbitrary character trimming
- This will also fix the functions for strings with non-ASCII characters where the `toArray` length is different from the length of the original string

Tested and working perfectly (tried with non-ASCII too)

### DIFF
--- a/addons/strings/fnc_leftTrim.sqf
+++ b/addons/strings/fnc_leftTrim.sqf
@@ -2,12 +2,13 @@
 Function: CBA_fnc_leftTrim
 
 Description:
-    Trims white-space (space, tab, newline) from the left end of a string.
+    Trims specified characters (all whitespace by default) from the left end of a string.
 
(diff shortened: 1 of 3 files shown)


PabstMirror:
Was failing these tests:
```
_str = ["   "] call CBA_fnc_leftTrim;
TEST_OP(_str,==,"",_fn);
_str = ["   "] call CBA_fnc_rightTrim;
TEST_OP(_str,==,"",_fn);
_str = [" 	"] call CBA_fnc_rightTrim; // including tabs
TEST_OP(_str,==,"",_fn);
```
I think this will fix: https://github.com/CBATeam/CBA_A3/commit/fa245be3447acee27db05a96c463ca0047d0ddf7

commy2:
Test:
```sqf
["Test", "Test", "Test", {systemChat str _this}, {}, [0xD3, [false, false, false]]] call CBA_fnc_addKeybind;
```

PR #892, opened 2018-02-14
Settings whitelist

### LINKED ISSUE #876: Add whitelist for server settings editing
**Arma 3 Version:** `1.80` (stable)
**CBA Version:** `3.5.0` (stable)

**Mods:**
```
- CBA_A3
- ACE3
```

**Description:**
- Adding a whitelist of player uids for server settings editing could be helpful to protect the editing of server settings in large communities. In our community every member has got the server password to load missions on the fly however server settings should not be editable by them. I would like to request a security feature where the server owner can specify certain uid's allowed to change server settings.

commy2:
- close #876

Adds `cba_settings_whitelist[]` array to both addon and mission config (description.ext).
The dedicated client/player can edit server tab settings if and only if any of these apply:
- their UID (`getPlayerUID`) appears inside either array
- both arrays are either empty or undefined and the player is a logged in admin
- the "admin" wildcard was used in either array and the player is a logged in admin

If you want nobody to edit the settings, not even the logged in admin, then you can use a throwaway UID (e.g. "nobody") in either addon or mission config array.

By default, both of them are undefined/empty, so point #2 applies (logged in admin can edit the settings which is the current behavior before this PR).

In SP or as Local Host, you can edit the settings regardless of this whitelist.

This requires putting down UID's in either a global addon or the mission config, so if you're worried about spoofing, don't use it :~)

Future improvement could be made by comparing the players UID on the server instead, but I don't feel like writing a callback system for this stuff atm. This way you could use a server machine local addon to compare the UID.

### DIFF
--- a/addons/settings/fnc_whitelisted.sqf
+++ b/addons/settings/fnc_whitelisted.sqf
@@ -1,0 +1,39 @@
+/* ----------------------------------------------------------------------------
+Internal Function: CBA_settings_fnc_whitelisted
+
+Description:
+    Check if local machine can edit server settings.
+
+Parameters:
+    None.
+
+Returns:
+    _return - Can change server settings? <BOOL>
+
+Examples:
+    (begin example)
+        [] call CBA_settings_fnc_whitelisted
+    (end)
+
+Author:
+    commy2
+---------------------------------------------------------------------------- */
+#include "script_component.hpp"
+
+private _uid = getPlayerUID player;
+private _whitelist = getArray configFile/QGVAR(whitelist);
+
+private _cfgMissionList = missionConfigFile/QGVAR(whitelist);
+if (isArray _cfgMissionList) then {
+    _whitelist append getArray _cfgMissionList;
+};
+
+// if neither addon nor mission have white list, use wildcard for admin instead
+if (_whitelist isEqualTo []) then {
+    _whitelist = ["admin"];
+};
+
+// admin wildcard and local machine is logged in admin
+if ("admin" in _whitelist && {IS_ADMIN_LOGGED}) exitWith {true};
+
+_uid in _whitelist
--- a/addons/settings/XEH_PREP.sqf
+++ b/addons/settings/XEH_PREP.sqf
@@ -7,6 +7,7 @@ PREP(import);
 PREP(export);
 PREP(clear);
 PREP(priority);
+PREP(whitelisted);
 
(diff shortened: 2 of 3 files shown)


commy2:
Waiting for feedback from @chris579.

chris579:
Like it!

addons/settings/fnc_whitelisted.sqf
```diff
@@ -0,0 +1,39 @@
...
+
+// if neither addon nor mission have white list, use wildcard for admin instead
+if (_whitelist isEqualTo []) then {
+    _whitelist = ["admin"];
+};
+
+// admin wildcard and local machine is logged in admin
+if ("admin" in _whitelist && {IS_ADMIN_LOGGED}) exitWith {true};
+
+_uid in _whitelist
```
Dystopian: maybe `_whitelist` should be GVAR and set at mission init, then `FUNC(whitelisted)` become `getPlayerUID player in GVAR(whitelist)`

Dystopian: no, `"admin" in GVAR(whitelist) && {IS_ADMIN_LOGGED} || {getPlayerUID player in GVAR(whitelist)}`

commy2: But then you can't set it with addon.

dedmen: Would give the ability to dynamically add entries to the variables. Meaning admins with debug console access can just add themselves to the list. Which is exactly what the original Request wanted to prevent.

commy2: Also true. But with the debug console, you can mess with the settings anyway if you know what you're doing.

Dystopian: > But then you can't set it with addon.

why? `configFile/QGVAR(whitelist)` should exist on mission init.

dedmen: >you can mess with the settings anyway if you know what you're doing.
But not with config files.
>But then you can't set it with addon.
Think commy missunderstood. ala GVAR != config.

Dystopian: Current state doesn't give any additional security because like commy2 said you can easy manage settings manually with debug console. But ability to change whitelist during mission could be helpful in some cases.

PR #893, opened 2018-02-14
make settings ediable in unsaved mission

### LINKED ISSUE #865: CBA Settings has greyed-out MISSION tab in 3DEN
**Arma 3 Version:** `1.80` (stable)
**CBA Version:** `3.5.0` (stable)

**Mods:**
```
- CBA_A3
- ACE3 v3.12.0
```

**Description:**
- In some cases (which turns out to be most cases for me), the MISSION settings tab is greyed out, making me unable to set the mission settings - https://i.imgur.com/n33NQ5d.jpg .
- Even though the screenshot shows extra mods used, I tested it with just CBA and ACE3 and then with CBA only, with new missions (both via the "New" button in 3den as well as loading the mission via arma3.exe cmdline), with existing missions, with a clean Arma 3 profile, with and without `-world=empty -nosplash`, after a mission preview, after MP preview, and so far I cannot get a reliable reproducer - I can edit the MISSION tab in ~10% of random cases, usually when loading an existing mission via arma3.exe cmdline.

- Editor (Singleplayer) / Editor (Multiplayer)

**RPT log file:**
- https://pastebin.com/9KrZfmtV (nothing interesting even with `cba_diagnostic_enable_logging.pbo`)

commy2:
- close #865

This was a limitation I made relatively early in the development of the settings, but it seems like it's no longer needed. I tested this and works as one would expect.

### DIFF
--- a/addons/settings/script_component.hpp
+++ b/addons/settings/script_component.hpp
@@ -95,11 +95,11 @@
 
 #define CAN_SET_SERVER_SETTINGS ((isServer || {IS_ADMIN_LOGGED}) && {!isNull GVAR(server)}) // in single player, as host (local server) or as logged in (not voted) admin connected to a dedicated server
 #define CAN_SET_CLIENT_SETTINGS !isServer // in multiplayer as dedicated client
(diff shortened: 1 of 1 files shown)


commy2:
Seems like these events also handle this properly:
https://github.com/CBATeam/CBA_A3/blob/master/addons/settings/fnc_initDisplay3DEN.sqf#L3-L62

PR #894, opened 2018-02-14
indicate settings that need a mission restart

### LINKED ISSUE #879: (settings) Indicator for non live-editable settings
**Arma 3 Version:** `1.80`
**CBA Version:** `3.5.0`

**Mods:**
```
- CBA_A3
```

**Description:**
- Add a per setting indicator whether it can be live-edited mid-mission or not.
As many ACE settings and other Mods don't support live-changing it's confusing to the user that the setting doesn't take effect.
Not sure about how we should solve that with server-side settings. As they need to be saveable/editable mid-mission but the user also needs to know about it not taking effect till restart.
For Mission/Client settings I think we can just block the editing ability.

commy2:
- display new icons if the setting in the current tab applies or is overwritten, instead of only when overwritten
- adds optional flag for settings that require a restart
- display yellow icon if the setting is changed during the mission, but requires a restart
- close #879

~~needs testing >.<~~


commy2:
Seems to work.

commy2:
Is that good enough? :joy: :hammer:

addons/settings/script_component.hpp
```diff
@@ -108,8 +115,9 @@
 #define GET_TEMP_NAMESPACE_VALUE(setting,source)    (GET_TEMP_NAMESPACE(source) getVariable [setting, [nil, nil]] select 0)
 #define GET_TEMP_NAMESPACE_PRIORITY(setting,source) (GET_TEMP_NAMESPACE(source) getVariable [setting, [nil, nil]] select 1)
 
-#define SET_TEMP_NAMESPACE_VALUE(setting,value,source)       (GET_TEMP_NAMESPACE(source) setVariable [setting, [value, GET_TEMP_NAMESPACE_PRIORITY(setting,source)]])
-#define SET_TEMP_NAMESPACE_PRIORITY(setting,priority,source) (GET_TEMP_NAMESPACE(source) setVariable [setting, [GET_TEMP_NAMESPACE_VALUE(setting,source), priority]])
+#define SET_TEMP_NAMESPACE_AWAITING_RESTART(setting) if (toLower setting in GVAR(needRestart)) then {GVAR(awaitingRestartTemp) pushBackUnique toLower setting}
```
PabstMirror: Maybe we could skip showing in 3den
```
#define SET_TEMP_NAMESPACE_AWAITING_RESTART(setting) if ((toLower setting in GVAR(needRestart)) && {!is3den}) then {GVAR(awaitingRestartTemp) pushBackUnique toLower setting}
```

PabstMirror:
@commy2 - Should the icon also change when the setting is changed, not just override checkboxes?

commy2:
It should be shown when either the setting is changed, or the settings priority is changed, or both I'd say.

PR #902, opened 2018-03-19
Add getUniquePlayerItems function

dedmen:
Default settings return the same as `items _unit` but without duplicates.

iterating over items unit like in https://github.com/acemod/ACE3/blob/master/addons/tagging/functions/fnc_addTagActions.sqf#L36
Can be very expensive if you are just searching for whether the player has one item. Because it returns all items and multiple items as multiple strings.

Was a total perf killer in TFAR when someone had a backpack with ACE medical supplies. I'm sure this can be useful for many things.

Also made it configurable as you might want to exclude backpack or include weapon items.

I don't know how I could make unit tests for this. And I also don't quite like the current Name. My creativity has it's limits.

### DIFF
--- a/addons/common/fnc_getUniquePlayerItems.sqf
+++ b/addons/common/fnc_getUniquePlayerItems.sqf
@@ -1,0 +1,44 @@
+/* ----------------------------------------------------------------------------
+Function: CBA_fnc_getUniquePlayerItems
+
+Description:
+    A function used to retrieve a unique list of items in the units inventory
+
+Parameters:
+    _unit           - Unit to retrieve the items from
+    _weaponItems    - Include weapons, attachments, loaded magazines (Default: false)
+    _backpack       - Include items in backpack (Default: true)
+    _vest           - Include items in vest     (Default: true)
+    _uniform        - Include items in uniform  (Default: true)
+
+Example:
+    (begin example)
+    _allItems = [player, true, false] call CBA_fnc_getUniquePlayerItems
+    (end)
+
+Returns:
+    Array of item classnames <ARRAY>
+
+Author:
+    Dedmen
+---------------------------------------------------------------------------- */
+#include "script_component.hpp"
+SCRIPT(getUniquePlayerItems);
+
+params [["_unit", objNull, [objNull]], ["_weaponItems", true, [true]], ["_backpack", true, [true]], ["_vest", true, [true]], ["_uniform", true, [true]]];
+
+private _allItems = (assignedItems _unit);
+if (_uniform) then {_allItems append ((getItemCargo (uniformContainer _unit)) select 0);};
+if (_vest) then {_allItems append ((getItemCargo (vestContainer _unit)) select 0);};
+if (_backpack) then {_allItems append ((getItemCargo (backpackContainer _unit)) select 0);};
+
+if (_weaponItems) then {
+    _allItems append (primaryWeaponItems _unit);
+    _allItems append (secondaryWeaponItems _unit);
+    _allItems append (handgunItems _unit);
+    _allItems append [  primaryWeapon _unit, secondaryWeapon _unit, handgunWeapon _unit,
+                        primaryWeaponMagazine _unit, secondaryWeaponMagazine _unit, handgunMagazine _unit
+                     ];
+};
+
+_allItems arrayIntersect _allItems //Remove duplicates


Dystopian:
maybe `getUniqueItems` because not only players but any units?

addons/common/fnc_getUniquePlayerItems.sqf
```diff
@@ -0,0 +1,44 @@
...
+private _allItems = (assignedItems _unit);
+if (_uniform) then {_allItems append ((getItemCargo (uniformContainer _unit)) select 0);};
+if (_vest) then {_allItems append ((getItemCargo (vestContainer _unit)) select 0);};
+if (_backpack) then {_allItems append ((getItemCargo (backpackContainer _unit)) select 0);};
+
+if (_weaponItems) then {
+    _allItems append (primaryWeaponItems _unit);
+    _allItems append (secondaryWeaponItems _unit);
+    _allItems append (handgunItems _unit);
+    _allItems append [  primaryWeapon _unit, secondaryWeapon _unit, handgunWeapon _unit,
```
jonpas: Array contents to new line, one indent right.

addons/common/fnc_getUniquePlayerItems.sqf
```diff
@@ -0,0 +1,44 @@
...
+                        primaryWeaponMagazine _unit, secondaryWeaponMagazine _unit, handgunMagazine _unit
+                     ];
```
jonpas: Same indent as `_allItems append ...`.

addons/common/fnc_getUniquePlayerItems.sqf
```diff
@@ -0,0 +1,44 @@
...
+if (_uniform) then {_allItems append ((getItemCargo (uniformContainer _unit)) select 0);};
+if (_vest) then {_allItems append ((getItemCargo (vestContainer _unit)) select 0);};
+if (_backpack) then {_allItems append ((getItemCargo (backpackContainer _unit)) select 0);};
+
+if (_weaponItems) then {
+    _allItems append (primaryWeaponItems _unit);
+    _allItems append (secondaryWeaponItems _unit);
+    _allItems append (handgunItems _unit);
+    _allItems append [  primaryWeapon _unit, secondaryWeapon _unit, handgunWeapon _unit,
+                        primaryWeaponMagazine _unit, secondaryWeaponMagazine _unit, handgunMagazine _unit
```
jonpas: Only one indent right.

jonpas:
> Was a total perf killer in TFAR when someone had a backpack with ACE medical supplies. I'm sure this can be useful for many things.

Same thing in ACRE, will be beneficial.

Soldia1138:
Great idea. So maybe this can be a starter to offer a unique ID framework within CBA?
There are quite a few use cases where a unique ID are beneficial.

TheMagnetar:
And ACE BFT as well...

commy2:
How does this compare to:
```sqf
params [["_unit", objNull, [objNull]];

private _items = items _unit;
_items arrayIntersect _items // return
```
?

commy2:
Naming sheme:
- the "get" prefix may or may not be optional
- "player" vs. "unit": this works on all units (persons), so "unit" instead of "player"
- `pushBack`, `pushBackUnique`: the "unique" qualifier is used as suffix
- I'd try to make it a 3 word compound if possible (easier to remember)

So my suggestions are:
```sqf
CBA_fnc_getUnitItemsUnique
CBA_fnc_unitItemsUnique
CBA_fnc_getItemsUnique
CBA_fnc_itemsUnique
```

commy2:
This function does not report the [`binocularMagazine`](https://github.com/CBATeam/CBA_A3/blob/master/addons/common/fnc_binocularMagazine.sqf) as far as I can tell.

addons/common/fnc_getUniquePlayerItems.sqf
```diff
@@ -0,0 +1,44 @@
+/* ----------------------------------------------------------------------------
+Function: CBA_fnc_getUniquePlayerItems
+
+Description:
+    A function used to retrieve a unique list of items in the units inventory
```
commy2: ~A function used to~ [R]etrieve[s] a unique list of items in the units inventory[.]

dedmen:
>How does this compare to:

In default settings aka no weapon items it should be the same. I used it in TFAR as a replacement for exactly that.

@Soldia1138 
>So maybe this can be a starter to offer a unique ID framework within CBA?

No. Not at all.

addons/common/fnc_getUniquePlayerItems.sqf
dedmen:
```
    _allItems append [
                        primaryWeapon _unit, secondaryWeapon _unit, handgunWeapon _unit,
                        primaryWeaponMagazine _unit, secondaryWeaponMagazine _unit, handgunMagazine _unit
                     ];
```
Or
```
    _allItems append [
    primaryWeapon _unit, secondaryWeapon _unit, handgunWeapon _unit,
    primaryWeaponMagazine _unit, secondaryWeaponMagazine _unit, handgunMagazine _unit
                     ];
```
? I'm confused now.

Cuel:
CBA_fnc_uniqueUnitItems in my humble opinion

addons/common/fnc_getUniquePlayerItems.sqf
jonpas:
```sqf
_allItems append [
    primaryWeapon _unit, secondaryWeapon _unit, handgunWeapon _unit,
    primaryWeaponMagazine _unit, secondaryWeaponMagazine _unit, handgunMagazine _unit
];
```

commy2:
`CBA_fnc_itemsUnique` personally. Merger of `items` and `pbUnique`.

dedmen:
@commy2 In the end you decide anyway. Want me to just go for your variant or want me to keep arguing that it's so ambiguous because the "unit" is missing.

commy2:
It's not "arguing", but throwing out ideas. Sadly you only have one shot and are stuck with the function name later, because bwc.

dedmen:
Let's vote?
`CBA_fnc_itemsUnique` :taco: 
`CBA_fnc_uniqueUnitItems` :burrito: 
`CBA_fnc_unitItemsUnique` :pizza:

commy2:
Since we are (hobby) programmers, we should at least use instant-runoff voting instead of this archaic method barely good enough to decide who runs a country.

commy2:
Pushed back, because https://github.com/CBATeam/CBA_A3/pull/902#issuecomment-374149360 is still unaddressed.

dedmen:
@commy2 it's supposed to replace items player. Does that return binocularMagazine?

addons/common/fnc_uniqueUnitItems.sqf
```diff
@@ -0,0 +1,46 @@
...
+if (_weaponItems) then {
+    _allItems append (primaryWeaponItems _unit);
+    _allItems append (secondaryWeaponItems _unit);
+    _allItems append (handgunItems _unit);
+    _allItems append [
+        primaryWeapon _unit, secondaryWeapon _unit, handgunWeapon _unit,
+        primaryWeaponMagazine _unit, secondaryWeaponMagazine _unit, handgunMagazine _unit
+     ];
+     _allItems pushBack (_unit call CBA_fnc_binocularMagazine);
+};
```
commy2:
```sqf
if (_weaponItems) then {
    _allItems append [
        primaryWeapon _unit,
        secondaryWeapon _unit,
        handgunWeapon _unit
    ];
    _allItems append primaryWeaponItems _unit;
    _allItems append secondaryWeaponItems _unit;
    _allItems append handgunItems _unit;
    _allItems primaryWeaponMagazine _unit
    _allItems secondaryWeaponMagazine _unit
    _allItems handgunMagazine _unit
    _allItems pushBack (_unit call CBA_fnc_binocularMagazine);
};
```

commy2: Remember that `primaryWeaponMagazine`-no S report ARRAY, exactly like the XItems versions.

addons/common/fnc_uniqueUnitItems.sqf
```diff
@@ -0,0 +1,46 @@
...
+---------------------------------------------------------------------------- */
+#include "script_component.hpp"
+SCRIPT(uniqueUnitItems);
+
+params [["_unit", objNull, [objNull]], ["_weaponItems", true, [true]], ["_backpack", true, [true]], ["_vest", true, [true]], ["_uniform", true, [true]]];
+
+private _allItems = (assignedItems _unit);
+if (_uniform) then {_allItems append ((getItemCargo (uniformContainer _unit)) select 0);};
+if (_vest) then {_allItems append ((getItemCargo (vestContainer _unit)) select 0);};
+if (_backpack) then {_allItems append ((getItemCargo (backpackContainer _unit)) select 0);};
```
commy2: parenthesis aaaaa

jonpas: Perfectly fine here to quickly see the order they are executed in.

commy2: aaaaa

jonpas:
```sqf
getItemCargo uniformContainer _unit select 0
```
this is just unreadable IMO.

addons/common/fnc_uniqueUnitItems.sqf
```diff
@@ -0,0 +1,48 @@
...
+Returns:
+    Array of item classnames <ARRAY>
+
+Author:
+    Dedmen
+---------------------------------------------------------------------------- */
+#include "script_component.hpp"
+SCRIPT(uniqueUnitItems);
+
+params [["_unit", objNull, [objNull]], ["_weaponItems", true, [true]], ["_backpack", true, [true]], ["_vest", true, [true]], ["_uniform", true, [true]]];
```
dedmen: Hah!
`["_weaponItems", true, [true]]` default false.. duh..

dedmen:
Btw here is some perf data about the use-case I made this for: https://github.com/acemod/ACEX/pull/114#discussion_r187768490

PR #905, opened 2018-04-14
disable compile cache in script_macros_mission.hpp

Cuel:
- Title

Any reason one would want this enabled for missions?


commy2:
No, but since BI killed includes from addon space in description.ext by crashing the server when loading the mission, no one can use this anyway.

Cuel:
I didn't know that. Guess I'll close this then

Issue #910, opened 2018-04-20
CBA_fnc_waitUntilAndExecute should accept 0 as timeout


felixlinker:
Hi there,

I think `CBA_fnc_waitUntilAndExecute` would behave more intuitively if the value for no timeout would be `-1` or something else truly less than zero.
In my case there is a timeout that might resolve to zero and in this case I would expect `CBA_fnc_waitUntilAndExecute` to execute the code immediately.

Right now, it treats a zero as if no timeout was given: https://github.com/CBATeam/CBA_A3/blob/master/addons/common/fnc_waitUntilAndExecute.sqf#L40

felixlinker:
Or just immediately.

felixlinker:
This is what I think would be better.

PR #911, opened 2018-04-20
fnc_waitUntilAndExecute now can timeout for 0s

felixlinker:
References #910 

`CBA_fnc_waitUntilAndExecute` now adopts timeout behaviour for timeout values > 0.

### DIFF
--- a/addons/common/fnc_waitUntilAndExecute.sqf
+++ b/addons/common/fnc_waitUntilAndExecute.sqf
@@ -3,7 +3,8 @@ Function: CBA_fnc_waitUntilAndExecute
 
 Description:
     Executes a code once in unscheduled environment after a condition is true.
-    It's also possible to add a timeout, in which case a different code is executed.
+    It's also possible to add a timeout >= 0, in which case a different code is executed.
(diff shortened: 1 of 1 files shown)


commy2:
That's what I imagined, but it still wouldn't execute code immediately, but later the frame or next frame.
This makes sense to me and is the behaviour I would expect.
The PR basically inverts the behavior for 0 and -1 tiímeout. Idk what the implications of that are for third party mods using this function, though I doubt 0 or -q is used much.
I would also describe 0 and -1 to the functions header.

addons/common/fnc_waitUntilAndExecute.sqf
```diff
@@ -3,7 +3,8 @@ Function: CBA_fnc_waitUntilAndExecute
 
 Description:
     Executes a code once in unscheduled environment after a condition is true.
-    It's also possible to add a timeout, in which case a different code is executed.
+    It's also possible to add a timeout >= 0, in which case a different code is executed.
+    Then it will be waited until _timeout time has elapsed or _condition evaluates to true.
```
Killswitch00: As commy2 suggests, describe what timeout 0 and -1 will mean.

PR #912, opened 2018-04-21
add CBA_fnc_progressBar

commy2:
- title


addons/ui/fnc_progressBar.sqf
```diff
@@ -0,0 +1,156 @@
+/* ----------------------------------------------------------------------------
+Function: CBA_fnc_display
+
+Description:
+    Opens a progress bar. Closes the currently active progress bar.
+
+Parameters:
+    _condition  - Execute every frame. If reports false, close the progress bar <CODE>
```
PabstMirror: _title <string>

addons/ui/fnc_progressBar.sqf
```diff
@@ -0,0 +1,168 @@
+/* ----------------------------------------------------------------------------
+Function: CBA_fnc_display
```
Cuel: wrong name

addons/ui/fnc_progressBar.sqf
```diff
@@ -0,0 +1,168 @@
...
+
+Arguments:
+    _this - same as _arguments <ANY>
+
+Returns:
+    Nothing
+
+Examples:
+    (begin example)
+        ["progress bar", {true}, 5, {hint "done"}, {hint "aborted"}] call CBA_fnc_display;
```
Cuel: wrong name

commy2:
ref: https://github.com/acemod/ACE3/blob/master/addons/common/functions/fnc_progressBar.sqf

addons/ui/fnc_progressBar.sqf
```diff
@@ -82,7 +82,12 @@ if (!isNull _display) then {
 };
 
 // create new progress bar
-_display = (uiNamespace getVariable "RscDisplayMission") createDisplay QGVAR(ProgressBar);
+if (_blockInput) then {
+    _display = (uiNamespace getVariable "RscDisplayMission") createDisplay QGVAR(ProgressBar);
+} else {
+    QGVAR(ProgressBar) cutRsc [QGVAR(ProgressBar), "PLAIN"];
+    _display = uiNamespace getVariable QGVAR(ProgressBar);
+};
```
commy2: todo, MouseMoving, MouseHolding don't work for cutRsc.

addons/ui/fnc_progressBar.sqf
```diff
@@ -0,0 +1,172 @@
...
+
+Author:
+    commy2
+---------------------------------------------------------------------------- */
+#include "script_component.hpp"
+
+if (!hasInterface) exitWith {};
+
+if (canSuspend) exitWith {
+    isNil (uiNamespace getVariable _fnc_scriptName);
```
SilentSpike: What's this bit all about?

commy2: Run the function unscheduled every time.

commy2: Same as https://github.com/CBATeam/CBA_A3/blob/master/addons/settings/fnc_init.sqf#L44-L47

addons/ui/fnc_progressBar.sqf
```diff
@@ -0,0 +1,172 @@
...
+
+    private _onSuccess = _display getVariable QGVAR(onSuccess);
+    private _onFailure = _display getVariable QGVAR(onFailure);
+
+    if (!_continue) exitWith {
+        _display closeDisplay 0;
+        [_onFailure, [_arguments, false, _elapsedTime, _totalTime]] call CBA_fnc_execNextFrame;
+    };
+
+    if (_elapsedTime > _totalTime) exitWith {
```
PabstMirror: `_elapsedTime = (CBA_missionTime - _startTime) min _totalTime;`
so elapsed can never be greater than total

addons/ui/fnc_progressBar.sqf
dedmen: Maybe prefer fnc_directCall for clarity?

SilentSpike: Yeah I just hadn't see the `uiNamespace` stuff before, was curious as to how that worked

commy2:
Any suggestion on how to order those parameters? -__-

dedmen:
onFailure could be optional. Most people probably don't need that.
And i would swap condition and totalTime.
So that you have all code variables together.

commy2:
In truth, all params are optional.

commy2:
- [ ] handle savegames

dedmen:
Why the change from a small onDraw EH to a big, always recompiling EachFrame EH?
Yeah it doesn't matter if your fps are one lower in a progress bar. You can't do anything anyway. But why?

commy2:
Latest is using onDraw.
I changed for a short while, because
- EF cerries over from savegames (fixed by using Loaded event checking a global var that also contains the args)
- killing the display from inside onDraw ctd's (fixed by execNextFrame, race conditions?)

dedmen:
>race conditions?
The onDraw EH runs in the middle of the Drawing function of the map control, not at the end. So if you delete the map who'se function is currently running and the function then tries to continue drawing the map and access it's own variables.. boom.

commy2:
Not  what I meant. I want to make sure this stuff doesn't behave weird if you e.g. create a new progress bar at the same time the previous one closes.

commy2:
Ready now hopefully.

addons/ui/fnc_progressBar.sqf
```diff
@@ -89,4 +80,70 @@ if (!isNil QGVAR(ProgressBarParams)) then {
 
 GVAR(ProgressBarParams) = [_arguments, _condition, _onSuccess, _onFailure, CBA_missionTime, _totalTime, _blockMouse, _blockKeys, _allowClose];
```
SilentSpike: Not that it really matters, but you could set these on the display itself and that way it get's cleaned up automatically

commy2: The display dies when saving and loading a savegame.
If I use a global instead, which is serialized properly, I can create a new loading bar with the same parameters once the game is loaded:
https://github.com/CBATeam/CBA_A3/pull/912/files#diff-8ff25e6d6abfb5db687c39f11a07e5b4R7

Only thing that doesn't work is the block input display, but I just assume that this only applies to mouse enabled progress bars, since

addons/ui/RscTitles.hpp
```diff
@@ -0,0 +1,57 @@
...
+            colorBar[] = GUI_BCG_COLOR;
+            texture = "#(argb,8,8,3)color(1,1,1,0.7)";
+            x = profileNamespace getVariable ['TRIPLES(IGUI,GVAR(grid),X)', 0];
+            y = profileNamespace getVariable ['TRIPLES(IGUI,GVAR(grid),Y)', 0];
+            w = profileNamespace getVariable ['TRIPLES(IGUI,GVAR(grid),W)', 0];
+            h = profileNamespace getVariable ['TRIPLES(IGUI,GVAR(grid),H)', 0];
+        };
+
+        class Script: RscMapControl {
+            idc = IDC_PROGRESSBAR_SCRIPT
```
PabstMirror: File x\cba\addons\ui\RscTitles.hpp, line 51: '/RscTitles/cba_ui_ProgressBar/Script.idc': Missing ';' at the end of line

Killswitch00:
Guys, get me a "+1" on this one if you feel it's ready.

PR #914, opened 2018-04-23
Fix rounded Travis badge

BaerMitUmlaut:
- Make this completly broken mod usable again


commy2:
backlog = no version

PR #915, opened 2018-04-25
Fix a issue that a PFH could be skip

jokoho48:
This PR fixes an issue that a Frame handler is skipped when the PFH that got executed before is deleted.
```sqf
{
    diag_log "delete FrameHandler No 1 Frame: " + str diag_frameNo;
    (_this select 1) call CBA_fnc_removePerFrameHandler;
} call CBA_fnc_addPerFrameHandler;

{
    diag_log "FrameHandler No 2 Frame: " + str diag_frameNo;
} call CBA_fnc_addPerFrameHandler;
```
Current Output:
>delete FrameHandler No 1 Frame: 5101
FrameHandler No 2 Frame:5102
FrameHandler No 2 Frame:5103
...

Expected Output:
>delete FrameHandler No 1 Frame: 5101
FrameHandler No 2 Frame:5101
FrameHandler No 2 Frame:5102
FrameHandler No 2 Frame:5103

### DIFF
--- a/addons/common/fnc_removePerFrameHandler.sqf
+++ b/addons/common/fnc_removePerFrameHandler.sqf
@@ -33,8 +33,7 @@ if (_handle < 0 || {_handle >= count GVAR(PFHhandles)}) exitWith {};
     if (isNil "_index") exitWith {};
     GVAR(deletedPFHIndices) pushback _index;
 
-    private _oldData = GVAR(perFrameHandlerArray) select _index;
-    _oldData set [6, false];
+    (GVAR(perFrameHandlerArray) select _index) set [0, {}];
 
     GVAR(PFHhandles) set [_handle, nil];
 }, _handle] call CBA_fnc_directCall;
--- a/addons/common/fnc_removePerFrameHandler.sqf
+++ b/addons/common/fnc_removePerFrameHandler.sqf
@@ -34,9 +34,7 @@ if (_handle < 0 || {_handle >= count GVAR(PFHhandles)}) exitWith {};
     GVAR(deletedPFHIndices) pushback _index;
 
     private _oldData = GVAR(perFrameHandlerArray) select _index;
-    _oldData set [5, false];
-
-    GVAR(perFrameHandlerArray) set [_index, _oldData];
+    _oldData set [6, false];
 
     GVAR(PFHhandles) set [_handle, nil];
 }, _handle] call CBA_fnc_directCall;
--- a/addons/common/fnc_removePerFrameHandler.sqf
+++ b/addons/common/fnc_removePerFrameHandler.sqf
@@ -34,7 +34,7 @@ if (_handle < 0 || {_handle >= count GVAR(PFHhandles)}) exitWith {};
     GVAR(deletedPFHIndices) pushback _index;
 
     private _oldData = GVAR(perFrameHandlerArray) select _index;
-    _oldData set [5, true];
+    _oldData set [5, false];
 
     GVAR(perFrameHandlerArray) set [_index, _oldData];
 
--- a/addons/common/fnc_removePerFrameHandler.sqf
+++ b/addons/common/fnc_removePerFrameHandler.sqf
@@ -31,7 +31,7 @@ if (_handle < 0 || {_handle >= count GVAR(PFHhandles)}) exitWith {};
     private _index = GVAR(PFHhandles) select _handle;
 
     if (isNil "_index") exitWith {};
-    GVAR(deletedPFHIndices) pushback _index;
+    GVAR(deletedPFHIndices) pushBack _index;
 
     (GVAR(perFrameHandlerArray) select _index) set [0, {}];
 
--- a/addons/common/fnc_removePerFrameHandler.sqf
+++ b/addons/common/fnc_removePerFrameHandler.sqf
@@ -28,14 +28,17 @@ if (_handle < 0 || {_handle >= count GVAR(PFHhandles)}) exitWith {};
 
 [{
     params ["_handle"];
+    private _index = GVAR(PFHhandles) select _handle;
 
-    GVAR(perFrameHandlerArray) deleteAt (GVAR(PFHhandles) select _handle);
+    if (isNil "_index") exitWith {};
+    GVAR(deletedPFHIndices) pushback _index;
+
+    private _oldData = GVAR(perFrameHandlerArray) select _index;
+    _oldData set [5, true];
+
+    GVAR(perFrameHandlerArray) set [_index, _oldData];
+
     GVAR(PFHhandles) set [_handle, nil];
-
-    {
-        _x params ["", "", "", "", "", "_handle"];
-        GVAR(PFHhandles) set [_handle, _forEachIndex];
-    } forEach GVAR(perFrameHandlerArray);
 }, _handle] call CBA_fnc_directCall;
 
 nil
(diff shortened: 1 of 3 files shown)


dedmen:
>This PR fixes an issue that a Frame handler is skipped when the PFH that got executed before is deleted.

I don't understand that.
PFH 2 is skipped when PFH1 is deleted?

dedmen:
Ahh I see now looking at the changes. Had same issue in Intercept_CBA.
That's caused by the elements after the PFH moving forward when it is deleted from the Array. And as count/forEach just works by index. It doesn't see that the index should be decremented by one to actually get the next element.

Quick fix https://github.com/CBATeam/CBA_A3/pull/915/files#diff-4175958b72ec7a414e894e1389d51c2fL33
->
`} count +GVAR(perFrameHandlerArray);`
I don't know if that's deep or shallow copy? Really only needs shallow.

addons/common/init_perFrameHandler.sqf
```diff
@@ -15,23 +15,36 @@ GVAR(nextFrameNo) = diag_frameno + 1;
...
 
-        if (diag_tickTime > _delta) then {
+        if (diag_tickTime > _delta && !_isDeleted) then {
             _x set [2, _delta + _delay];
             [_args, _handle] call _function;
             false
         };
     } count GVAR(perFrameHandlerArray);
 
+    if !(GVAR(deletedIndices) isEqualTo []) then {
```
PabstMirror: probably should be `deletedPFHIndices`

addons/common/init_perFrameHandler.sqf
```diff
@@ -15,23 +15,36 @@ GVAR(nextFrameNo) = diag_frameno + 1;
...
     private _tickTime = diag_tickTime;
     call FUNC(missionTimePFH);
 
     // Execute per frame handlers
     {
-        _x params ["_function", "_delay", "_delta", "", "_args", "_handle"];
+        _x params ["_function", "_delay", "_delta", "", "_args", "_handle", "_isDeleted"];
 
-        if (diag_tickTime > _delta) then {
+        if (diag_tickTime > _delta && !_isDeleted) then {
```
PabstMirror: this will be called multiple times each frame, lets invert the logic on the var and skip the not
`_isActive`?

jokoho48: good point

addons/common/init_perFrameHandler.sqf
jokoho48: wups yeah oversaw that

addons/common/fnc_removePerFrameHandler.sqf
```diff
@@ -28,14 +28,17 @@ if (_handle < 0 || {_handle >= count GVAR(PFHhandles)}) exitWith {};
...
     params ["_handle"];
+    private _index = GVAR(PFHhandles) select _handle;
 
-    GVAR(perFrameHandlerArray) deleteAt (GVAR(PFHhandles) select _handle);
-    GVAR(PFHhandles) set [_handle, nil];
+    if (isNil "_index") exitWith {};
+    GVAR(deletedPFHIndices) pushback _index;
+
+    private _oldData = GVAR(perFrameHandlerArray) select _index;
+    _oldData set [5, false];
```
Killswitch00: Should 5 be 6 here to set the last array element of the `perFrameHandlerArray`entry? (The one set to "true" in `fnc_addPerFrameHandler`, `_isActiv` in `init_perFrameHandler.sqf)?

addons/common/init_perFrameHandler.sqf
```diff
@@ -16,22 +16,36 @@ GVAR(nextFrameBufferA) = [[[], {GVAR(nextFrameNo) = diag_frameno;}]];
...
+GVAR(deletedPFHIndices) = [];
 // per frame handler system
 [QFUNC(onFrame), {
     private _tickTime = diag_tickTime;
     call FUNC(missionTimePFH);
 
     // Execute per frame handlers
     {
-        _x params ["_function", "_delay", "_delta", "", "_args", "_handle"];
+        _x params ["_function", "_delay", "_delta", "", "_args", "_handle", "_isActiv"];
```
Killswitch00: `_isActive`? (english spelling)

addons/common/init_perFrameHandler.sqf
```diff
@@ -16,22 +16,36 @@ GVAR(nextFrameBufferA) = [[[], {GVAR(nextFrameNo) = diag_frameno;}]];
...
+            GVAR(perFrameHandlerArray) set [_x, objNull];
+        } count GVAR(deletedPFHIndices);
+
+        GVAR(perFrameHandlerArray) = GVAR(perFrameHandlerArray) - [objNull];
+
+        {
+            _x params ["", "", "", "", "", "_handle"];
+            GVAR(PFHhandles) set [_handle, _forEachIndex];
+        } forEach GVAR(perFrameHandlerArray);
+        GVAR(deletedPFHIndices) = [];
```
Killswitch00: Move up to line 40 to be in close proximity to where operations on `deletedPFHIndices` are finished.

addons/common/fnc_removePerFrameHandler.sqf
dedmen: Yes. 5 is _handle

addons/common/fnc_removePerFrameHandler.sqf
```diff
@@ -28,14 +28,17 @@ if (_handle < 0 || {_handle >= count GVAR(PFHhandles)}) exitWith {};
...
 
-    {
-        _x params ["", "", "", "", "", "_handle"];
-        GVAR(PFHhandles) set [_handle, _forEachIndex];
-    } forEach GVAR(perFrameHandlerArray);
+    GVAR(perFrameHandlerArray) set [_index, _oldData];
```
dedmen: This is not needed. You already modify by reference.

addons/common/init_perFrameHandler.sqf
```diff
@@ -16,22 +16,36 @@ GVAR(nextFrameBufferA) = [[[], {GVAR(nextFrameNo) = diag_frameno;}]];
...
+    if !(GVAR(deletedPFHIndices) isEqualTo []) then {
+        {
+            GVAR(perFrameHandlerArray) set [_x, objNull];
+        } count GVAR(deletedPFHIndices);
+
+        GVAR(perFrameHandlerArray) = GVAR(perFrameHandlerArray) - [objNull];
+
+        {
+            _x params ["", "", "", "", "", "_handle"];
+            GVAR(PFHhandles) set [_handle, _forEachIndex];
```
dedmen: Wait.. You are resetting the handles here? meaning everyone that saved his handles will delete some random PFH but not the one they intended to delete.

commy2:
Is this really worth it? One local variable, one more `&&` and one more `if isEqualTo []` every frame.

Could've just used:
`+ GVAR(perFrameHandlerArray)`
tbh.

PabstMirror:
We could get rid of the `_isActive` check and just do `_oldData set [0, {}];`

dedmen:
>Could've just used:
>\+ GVAR(perFrameHandlerArray)
>tbh.

https://github.com/CBATeam/CBA_A3/pull/915#issuecomment-384341180 ;)

commy2:
`+ ARR` is deep copy. `[] + ARR` is flat copy. Idk if flat actually is better for performance.

dedmen:
`myArr = [["string",123,{code}],["string",123,{code}],["string",123,{code}],["string",123,{code}],["string",123,{code}]]`
`+myArr` 0.0055ms 0.0056ms 0.0057ms
`[] + myArr` 0.0034ms 0.0034ms 0.0034ms

PabstMirror:
on delete: `GVAR(cleanupPFEH) = true;` and replace `!(x isEqualTo [])` with that

commy2:
Still don't like this.

PR #916, opened 2018-04-28
FIX: cba_fnc_deleteEntity where group must be local to the machine executing deleteGroup command

Vdauphin:
Hello, I don't know if this function is supposed to be executed only locally but here is the fix in case group is not local to the machine trying to delete the group, or at least a solution.

- Fix [`cba_fnc_deleteEntity`](https://github.com/CBATeam/CBA_A3/blob/master/addons/common/fnc_deleteEntity.sqf)  where the group must be local to the machine executing the [`deleteGroup`](https://community.bistudio.com/wiki/deleteGroup). 
- check if group is local
- if not delete it where the group is local if we know it
- if not send the `deletegroup` to the server where `groupOwner` can be
know

### DIFF
--- a/addons/common/fnc_deleteEntity.sqf
+++ b/addons/common/fnc_deleteEntity.sqf
@@ -43,7 +43,15 @@ switch (typeName _entity) do {
     case "GROUP" : {
         (units _entity) call CBA_fnc_deleteEntity;
         {deleteWaypoint _x} forEach (wayPoints _entity);
-        deleteGroup _entity;
+        if (local _entity) then {
+            deleteGroup _entity;
+        } else {
+            if (isServer) then {
+                _entity remoteExecCall ["CBA_fnc_deleteEntity", groupOwner _entity];
+            } else {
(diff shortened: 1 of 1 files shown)


commy2:
Just curious. Did you actually use that function and encountered this particular problem?

Vdauphin:
First, I wrote a similar function ([btc_fnc_delete](https://github.com/Vdauphin/HeartsAndMinds/blob/master_daily/%3DBTC%3Dco%4030_Hearts_and_Minds.Altis/core/fnc/common/delete.sqf)) and notice this issue so I wrote this [btc_fnc_deletegroup](https://github.com/Vdauphin/HeartsAndMinds/blob/master_daily/=BTC=co@30_Hearts_and_Minds.Altis/core/fnc/common/deletegroup.sqf).

Now, I want to use more CBA function and saw that, so I send the fix.

commy2:
I see. Script looks good to me.

Cuel:
I know this is merged already but wouldn't it be best to just ` _entity remoteExecCall ["CBA_fnc_deleteEntity", _entity];`

PR #918, opened 2018-05-10
French Translation

NiiRoZz:
This PR adds some french translation.

### DIFF
--- a/addons/main/stringtable.xml
+++ b/addons/main/stringtable.xml
@@ -9,6 +9,7 @@
             <Chinesesimp>CBA制作组</Chinesesimp>
             <Portuguese>Time CBA</Portuguese>
             <Russian>Команда CBA</Russian>
+            <French>CBA-Team</French>
         </Key>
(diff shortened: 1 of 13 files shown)


addons/main/stringtable.xml
```diff
@@ -17,6 +18,7 @@
             <Chinesesimp>https://www.github.com/CBATeam/CBA_A3</Chinesesimp>
             <Portuguese>https://www.github.com/CBATeam/CBA_A3</Portuguese>
             <Russian>https://www.github.com/CBATeam/CBA_A3</Russian>
+            <French>https://www.github.com/CBATeam/CBA_A3</French>
```
dedmen: Not needed. If it's not there Arma defaults to the topmost one. And as all others are the same anyway..
Saves a tiny bit of disk space ;)

jonpas: But will be marked as missing in stringtable checker. Keep it.

commy2:
gracias!!! (jk)

PR #922, opened 2018-05-12
Fix uniqueUnitItems typo and add assignedItems switch

dedmen:
- Fixes typo https://github.com/CBATeam/CBA_A3/pull/902/files#r187768475

Not a critical bug. Just causes the function to return more items that advertised. Which given it's inteded use case doesn't matter much.

- Add's switch to disable assignedItems because these half a dozen items matter for some users

Do you prefer to keep the useless semicolons that I removed?

### DIFF
--- a/addons/common/fnc_uniqueUnitItems.sqf
+++ b/addons/common/fnc_uniqueUnitItems.sqf
@@ -10,6 +10,7 @@ Parameters:
     _backpack       - Include items in backpack (Default: true)
     _vest           - Include items in vest     (Default: true)
     _uniform        - Include items in uniform  (Default: true)
+    _assignedItems        - Include assigned items  (Default: true)
 
 Example:
     (begin example)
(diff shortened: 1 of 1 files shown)


addons/common/fnc_uniqueUnitItems.sqf
```diff
@@ -10,6 +10,7 @@ Parameters:
     _backpack       - Include items in backpack (Default: true)
     _vest           - Include items in vest     (Default: true)
     _uniform        - Include items in uniform  (Default: true)
+    _assignedItems        - Include assigned items  (Default: true)
```
jonpas: What happened to alignment here?

dedmen: Dedmen just copy pasted a entry from above and changed the names. Dunno why I didn't see that 
(╯°□°）╯︵ ┻━┻

PR #925, opened 2018-05-17
Cache CreditsLine entries

dedmen:
cba_help_fnc_setCreditsLine is the major lag cause when opening the Pause menu.
arma3_x64_2018-05-17_20-33-03
This is a very simple fix to get rid of that.
Scanning all CfgPatches entries everytime a player presses Escape wasn't a good idea in the first place.

### DIFF
--- a/addons/help/fnc_setCreditsLine.sqf
+++ b/addons/help/fnc_setCreditsLine.sqf
@@ -29,11 +29,7 @@ if (CBA_DisableCredits) exitWith {};
 
 // find addon with author
 private _config = configFile >> "CfgPatches";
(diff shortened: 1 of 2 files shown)


commy2:
I love you.

addons/help/fnc_setCreditsLine.sqf
```diff
@@ -29,11 +29,7 @@ if (CBA_DisableCredits) exitWith {};
 
 // find addon with author
 private _config = configFile >> "CfgPatches";
```
shukari: This line is also now obsolete... _config never get used in that script.

PR #932, opened 2018-05-25
Replace BIS function call with engine command

Wakbub:
- Replace BIS function call `BIS_fnc_getCfgData` with with engine command `getText`.

**Links**
- [BIS_fnc_getCfgData](https://community.bistudio.com/wiki/BIS_fnc_getCfgData)
- [getText](https://community.bistudio.com/wiki/getText)

### DIFF
--- a/addons/accessory/fnc_switchAttachment.sqf
+++ b/addons/accessory/fnc_switchAttachment.sqf
@@ -18,7 +18,7 @@ Examples:
     (end)
 
 Author:
-    Robalo
+    Robalo, optimized by Anton
 ---------------------------------------------------------------------------- */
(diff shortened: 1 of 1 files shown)


dedmen:
Old one supports Number, Text and Array.
New one only supports text?

PabstMirror:
It just does ` hintSilent format ["%1", _switchItemHintText]`
So number/array wouldn't make much sense

Wakbub:
MRT_SwitchItemHint**Text** is always Text, so there is no point in "supporting" other types.

commy2:
I'd rather have someone make a nicer looking ui element, possibly one that can be reused for other purposes and other mods.

PR #937, opened 2018-06-11
[WIP] Fix script errors reporting wrong line numbers

dedmen:


ViperMaul:
Nice @dedmen 
Also what we need is a regex script to apply your same fix to all of the other SQF files that have this issue. There are well over 300+ additional files where `#include "script_component.hpp"` is located below comments.

SilentSpike:
This has bugged me for some long in ace too, good solve 👍

SilentSpike:
For the regex this should do it:

```regex
^\s*(/\*.+?\*/)\s*#include\s*"script_component\.hpp"
```
In sublime add the flags `(?s)(?-m)` to the front to disable multiline and enable dotall.

Replace with:
```regex
#include "script_component.hpp"\n\1
```

dedmen:
>Looks like you are already on that path. Thank you! 
correct I wanted to... Oh.. @SilentSpike Don't steal my work!!!!!111!11"
I'll also do the ACE PR when I find the time. Expect atleast a couple weeks

addons/common/fnc_getUISize.sqf
```diff
@@ -24,8 +25,6 @@ Examples:
 Author:
     Written by Deadfast and made CBA compliant by Vigilante
 ---------------------------------------------------------------------------- */
-
-#include "script_component.hpp"
```
dedmen: This doesn't match the default style. So I'd say it's okey that, that newline is gone?

dedmen:
The last commit has stuff that was not just done automatically. Take special care about that when reviewing the PR.

dedmen:
I'll happily fix conflicts. If you merge #939 I'll have to add them anyway.
I'd rather wait for atleast 2 reviews to make sure this is right. Also there are open questions from me above and the documenation generator needs to be tested.

If something is wrong in here it could break CBA at any point, sensitive or not. Don't wanna take that risk

addons/common/fnc_getUISize.sqf
SilentSpike: Yeah, not sure if you used my regex, but I specifically made it account for cases where there might be space and strip it out

SilentSpike:
The last commit looks all good 👍

addons/common/fnc_getUISize.sqf
dedmen: I used yours. But slightly modified `[\s\S]` instead of `.`

dedmen:
Just waiting for https://github.com/CBATeam/CBA_A3/pull/937#issuecomment-399756189
and https://github.com/CBATeam/CBA_A3/pull/937#issuecomment-399714544 
🕐 ⏳ ⏲ 🌳 🍪

dedmen:
You wanted it merged asap, not me :D 
merge conflicts shouldn't be an issue as this only changes essentially the first line of the scripts. Which is usually never touched.

PR #941, opened 2018-06-24
Cleanup private ARRAY

dedmen:
... and modernize vectCross


addons/ui/flexiMenu/fnc_getMenuDef.sqf
```diff
@@ -3,23 +3,20 @@
...
-_menuSources = _this select 1; // [_target, _menuSources];
-
-_menuDefs = [];
+private _target = _this select 0;
+private _menuSources = _this select 1; // [_target, _menuSources];
+private _menuDefs = [];
 
 { // forEach
-    private ["_params", "_menuSource", "_menuParams", "_array", "_menuDef"]; // declare locally to safe guard variables after _menuSource call, which is beyond our control of correctness.
+    private ["_menuDef"]; // declare locally to safe guard variables after _menuSource call, which is beyond our control of correctness.
```
commy2: Function is broken before and after. _menuDef at the eof will always report [].

PR #950, opened 2018-07-03
don't skip next PFH if current one is removed while iterating through

commy2:
- deactivate PFH by replacing the script with `{}`, then delete NextFrame (guaranteed to happen outside of iterating through the array that is modified)
- more elegant than https://github.com/CBATeam/CBA_A3/pull/915 because doesn't change the EachFrame function
- close #915 

Needs testing, I won't have access to Arma this and the next week.

### DIFF
--- a/addons/common/fnc_removePerFrameHandler.sqf
+++ b/addons/common/fnc_removePerFrameHandler.sqf
@@ -9,7 +9,7 @@ Parameters:
     _handle - The function handle you wish to remove. <NUMBER>
 
 Returns:
-    None
+    true if removed successful, false otherwise <BOOLEAN>
 
 Examples:
     (begin example)
@@ -24,24 +24,31 @@ Author:
 
 params [["_handle", -1, [0]]];
 
-if (_handle < 0 || {_handle >= count GVAR(PFHhandles)}) exitWith {};
-
 [{
     params ["_handle"];
 
-    GVAR(perFrameHandlerArray) set [GVAR(PFHhandles) select _handle select 0, {}];
+    private _index = GVAR(PFHhandles) param [_handle];
+    if (isNil "_index") exitWith {false};
 
-    [{
-        params ["_handle"];
+    GVAR(PFHhandles) set [_handle, nil];
+    (GVAR(perFrameHandlerArray) select _index) set [0, {}];
 
-        GVAR(perFrameHandlerArray) deleteAt (GVAR(PFHhandles) select _handle);
-        GVAR(PFHhandles) set [_handle, nil];
+    if (GVAR(perFrameHandlersToRemove) isEqualTo []) then {
+        [{
+            {
+                GVAR(perFrameHandlerArray) set [_x, objNull];
+            } forEach GVAR(perFrameHandlersToRemove);
 
-        {
-            _x params ["", "", "", "", "", "_handle"];
-            GVAR(PFHhandles) set [_handle, _forEachIndex];
-        } forEach GVAR(perFrameHandlerArray);
-    }, _handle] call CBA_fnc_execNextFrame;
+            GVAR(perFrameHandlerArray) = GVAR(perFrameHandlerArray) - [objNull];
+            GVAR(perFrameHandlersToRemove) = [];
+
+            {
+                _x params ["", "", "", "", "", "_index"];
+                GVAR(PFHhandles) set [_index, _forEachIndex];
+            } forEach GVAR(perFrameHandlerArray);
+        }] call CBA_fnc_execNextFrame;
+    };
+
+    GVAR(perFrameHandlersToRemove) pushBackUnique _index;
+    true
 }, _handle] call CBA_fnc_directCall;
-
-nil
--- a/addons/common/fnc_removePerFrameHandler.sqf
+++ b/addons/common/fnc_removePerFrameHandler.sqf
@@ -34,7 +34,7 @@ params [["_handle", -1, [0]]];
     (GVAR(perFrameHandlerArray) select _index) set [0, {}];
 
     if (GVAR(perFrameHandlersToRemove) isEqualTo []) then {
-        [{
+        {
             {
                 GVAR(perFrameHandlerArray) set [_x, objNull];
             } forEach GVAR(perFrameHandlersToRemove);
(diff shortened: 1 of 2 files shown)


commy2:
Q: What happens if the removePFH function is called twice with the same handle in the same frame (deleting already removed PFH waiting to be cleaned up NextFrame)?

addons/common/fnc_removePerFrameHandler.sqf
```diff
@@ -29,13 +29,19 @@ if (_handle < 0 || {_handle >= count GVAR(PFHhandles)}) exitWith {};
 [{
     params ["_handle"];
 
-    GVAR(perFrameHandlerArray) deleteAt (GVAR(PFHhandles) select _handle);
-    GVAR(PFHhandles) set [_handle, nil];
+    GVAR(perFrameHandlerArray) set [GVAR(PFHhandles) select _handle select 0, {}];
```
dedmen: Fun fact. `call` doesn't exit early if it has no code. It will actually really build a callstack and everything required. And then execute the empty code and exit right after.

addons/common/fnc_removePerFrameHandler.sqf
```diff
@@ -29,13 +29,19 @@ if (_handle < 0 || {_handle >= count GVAR(PFHhandles)}) exitWith {};
...
-        GVAR(PFHhandles) set [_handle, _forEachIndex];
-    } forEach GVAR(perFrameHandlerArray);
+    [{
+        params ["_handle"];
+
+        GVAR(perFrameHandlerArray) deleteAt (GVAR(PFHhandles) select _handle);
+        GVAR(PFHhandles) set [_handle, nil];
+
+        {
+            _x params ["", "", "", "", "", "_handle"];
```
dedmen: reusing _handle here looks confusing to some people. But these people don't look at CBA code anyway.

(I pressed send review and wanted to submit the comment after that.. But who could've guessed that the comment is discarded when you submit the review.)

commy2: I just indented the previous code. The function is simple enough I'd say.

addons/common/fnc_removePerFrameHandler.sqf
commy2: I wouldn't have implemented it any other way tbh. For simplicity.

SilentSpike:
I like the elegance, though running through it in my head I'm also not sure this isn't still broken as per commy's comment above.

The way it is currently, I believe that if the function is called **_n_** times in the same frame for the same handler, the next **_n - 1_** handlers would also be deleted. 

My first thought:
You'd need to add a sanity check for whether the `GVAR(PFHhandles)` value is already `nil` (meaning the handle has already been deleted that frame). However, this would only work assuming the handle index hasn't been refilled by another call to `addPerFrameHandler` between the duplicated `execNextFrame` calls (unsure if this can happen, I'll assume it probably can).

My next thought:
It'd be a hell of a lot simpler and actually foolproof to just check if the code is already `{}` in the scope above before the `execNextFrame` is even added.

jokoho48:
I like the base idea over what I have done before. but what I dislike is that you run over every PFH Multiple times when more than 1 PFH got remove per frame. so you would have both systems combined in a way.
so what I would do is still save the deletedIndices and only add 1 execNextFrame and use the removing Method I implemented just in the execNextFrame and checking if a remove already happend

commy2:
@SilentSpike Can you think of a way to potential bring this issue about? Reading the code again after sleep, everything looks fine, even if the same ID is removed twice.
@jokoho48 
>but what I dislike is that you run over every PFH Multiple times when more than 1 PFH got remove per frame.

This almost never happens though. Removing multiple PFH in the same frame.

dedmen:
>but what I dislike is that you run over every PFH Multiple times when more than 1 PFH got remove per frame.
I still prefer that over the alternative of executing some code every Frame even if it's not needed 99% of the time.

jokoho48:
>This almost never happens though. Removing multiple PFH in the same frame.

it does that is why I bringing this up. and also the check happens I frame After the PFH is removed

>I still prefer that over the alternative of executing some code every Frame even if it's not needed 99% of the time.

then you don't read my post correctly. I sad this is a better solution but I think that it could be more optimized. in the way that we could combine both ideas and use the execNextFrame for the base removal but still save the indices that are deleted

addons/common/fnc_removePerFrameHandler.sqf
```diff
@@ -29,13 +29,19 @@ if (_handle < 0 || {_handle >= count GVAR(PFHhandles)}) exitWith {};
...
+    GVAR(perFrameHandlerArray) set [GVAR(PFHhandles) select _handle select 0, {}];
 
-    {
-        _x params ["", "", "", "", "", "_handle"];
-        GVAR(PFHhandles) set [_handle, _forEachIndex];
-    } forEach GVAR(perFrameHandlerArray);
+    [{
+        params ["_handle"];
+
+        GVAR(perFrameHandlerArray) deleteAt (GVAR(PFHhandles) select _handle);
```
SilentSpike: @commy2 Yeah reading again after a sleep I don't think it would break anymore (since this line becomes `deleteAt nil` on any duplicated calls - which I assume won't error?).

SilentSpike: The only exception would be, if you were to remove the same handle twice in the same frame and also addEventHandler in the next frame, would it be possible that the same index (`_handle`) could be refilled in between the two `execNextFrame` calls? Resulting in the newly added EH being deleted immediately.

commy2:
What about this? Tested it a bit with laptop, and it seems to work as expected.

addons/common/fnc_removePerFrameHandler.sqf
```diff
@@ -24,24 +24,31 @@ Author:
...
+    if (isNil "_index") exitWith {false};
 
-    [{
-        params ["_handle"];
+    GVAR(PFHhandles) set [_handle, nil];
+    (GVAR(perFrameHandlerArray) select _index) set [0, {}];
 
-        GVAR(perFrameHandlerArray) deleteAt (GVAR(PFHhandles) select _handle);
-        GVAR(PFHhandles) set [_handle, nil];
+    if (GVAR(perFrameHandlersToRemove) isEqualTo []) then {
```
SilentSpike: Missing `!` ?

SilentSpike: Ah never mind I see how it works now, because it's next frame and only runs once due to this

addons/common/fnc_removePerFrameHandler.sqf
commy2: This is all changed now.

PabstMirror:
we've rewritten this fix many times but this one is the best; no run-time performance hit

PR #951, opened 2018-07-03
handle input gracefully in keybinding editKey ui function

commy2:
- Ignore vanilla input, i.e. don't close the menu if you assign a keyaction to `Num Enter`, `Enter`; don't change control focus when assigning a keyaction to `Tab` etc.
- the `call` is needed, because due to a bug, return values in eventhandlers are ignored if they come from `exitWith` in the main scope of the eventhandler script.

~~Needs testing, don't have access to Arma the next two weeks.~~


commy2:
Tested, works.

PR #952, opened 2018-07-03
change color of name of temporarily changed settings

commy2:
- This will make the names of settings that were changed by ui appear yellow, which helps when editing settings in bulk. Serves as a reminder which settings will be changed once OK is pressed, which changes will be discarded once Cancel is pressed.

~~Needs testing, don't have Arma for 2 weeks.~~


commy2:
Tested. Now appears to work.

PR #956, opened 2018-07-28
UI - Keep getFov params consistant with previous versions

PabstMirror:
https://github.com/acemod/ACE3/issues/6475

### DIFF
--- a/addons/ui/fnc_getFov.sqf
+++ b/addons/ui/fnc_getFov.sqf
@@ -25,7 +25,7 @@ Authors:
     streamlined by commy2
 ---------------------------------------------------------------------------- */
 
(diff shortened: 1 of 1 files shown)


SilentSpike:
I prefer the usage of `params` personally, it slightly changes the behaviour here, but I think it's preferable to flag incorrect/unintended input parameters.

commy2:
The behaviour is the same, except for the on screen messgae if -showScriptErrors is enabled / in 3den/3den preview.

PabstMirror:
If `_this` is `[0]` the behaviour will be different

I know plenty of people do play with script errors on
Perhaps, we should merge this and @alganthe's PRs and then switch to `params` for next CBA

PR #958, opened 2018-08-07
Fix Layout Editor running out of repositionable controls

mharis001:
- Layout editor only has 22 repositionable controls


commy2:
I need this. Thank you so much.

commy2:
Does Vanilla really start with Element000 ?

mharis001:
Done, and yeah `Element000` -> `Element021`.

addons/ui/RscDisplayOptionsLayout.hpp
```diff
@@ -0,0 +1,93 @@
+// Fixes Layout Editor only having 22 repositionable controls
+class RscDisplayOptionsLayout {
+    class controlsBackground {
+        class Element021;
+        #define ADD_ELEMENT(var1) class Element##var1: Element021 {\
+            idc = __EVAL(12000 + var1);\
+            onMouseEnter = "with uinamespace do {['mouseEnter',_this,''] call RscDisplayOptionsLayout_script;};";\
```
Dystopian: why are the same config values set in each element? Aren't they inherited?

dedmen: The Eventhandlers are added by script. So no Element000->021 doesn't have them so we need to add them. Which is what's happening here.

But how about we just make Element022 with the Eventhandlers and just have all the others inherit from that? Less code in the preprocessed config in the end.

commy2: I will take a look at the master config.

commy2: This is how it looks in the displays init script:
```sqf
// RscDisplayOptionsLayout.sqf
...
for "_i" from 0 to 21 do {
	_control = _display displayctrl (IDCBASE + _i);
	_control ctrladdeventhandler ["mouseenter","with uinamespace do {['mouseEnter',_this,''] call RscDisplayOptionsLayout_script;};"];
	_control ctrladdeventhandler ["mouseexit","with uinamespace do {['mouseExit',_this,''] call RscDisplayOptionsLayout_script;};"];
	_control ctrladdeventhandler ["mouseholding","with uinamespace do {['mouseMoving',_this,''] call RscDisplayOptionsLayout_script;};"];
	_control ctrladdeventhandler ["mousemoving","with uinamespace do {['mouseMoving',_this,''] call RscDisplayOptionsLayout_script;};"];
	_control ctrladdeventhandler ["mousebuttondown","with uinamespace do {['mouseButtonDown',_this,''] call RscDisplayOptionsLayout_script;};"];
	_control ctrladdeventhandler ["mousebuttonup","with uinamespace do {['mouseButtonUp',_this,''] call RscDisplayOptionsLayout_script;};"];
};
...
```
Strangely, in a few other places it just iterates through the child classes instead of assuming 0...21. /shrug

PR #959, opened 2018-08-08
Workaround for settings size limitation

jokoho48:
this PR is a workaround that fixes an issue that settings could only be 8kb long due to the use of format.

example: https://gist.github.com/neilzar/cdf4fa7a06d43560df64eea36c2cffb5
parse throws the error
>Error Generic error in expression
Error position: <parseSimpleArray format ["[%1]", _string>
File x\cba\addons\settings\fnc_parse.sqf, line 30

### DIFF
--- a/addons/settings/fnc_parse.sqf
+++ b/addons/settings/fnc_parse.sqf
@@ -27,7 +27,7 @@ private _fnc_parseAny = {
         _string = _string splitString WHITESPACE joinString "";
     };
 
-    parseSimpleArray format ["[%1]", _string] select 0
+    parseSimpleArray str (formatText ["[%1]", _string]) select 0
 };
 
 params [["_info", "", [""]], ["_validate", false, [false]], ["_source", "", [""]]];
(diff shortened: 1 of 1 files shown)


dedmen:
how about 
`["[",_string,"]"] joinString ""`
?
Should be faster.
Also 8kb? :D Where did you get that many settings?

jonpas:
> Also 8kb?
Exactly what he linked I assume.

dedmen:
>Exactly what he linked I assume.
:man_facepalming: Too hot outside.

jokoho48:
> Also 8kb? :D Where did you get that many settings?

allow sound mods and other small client side mods and you have so much in ace_common_checkPBOsWhitelist.

and good idea with joinString. will do

addons/settings/fnc_parse.sqf
```diff
@@ -27,7 +27,7 @@ private _fnc_parseAny = {
         _string = _string splitString WHITESPACE joinString "";
     };
 
-    parseSimpleArray format ["[%1]", _string] select 0
+    parseSimpleArray (["[", _string, "]"] joinString "") select 0;
```
commy2: I left out the semi colon to indicate the scope's return value.

PR #960, opened 2018-08-10
add tooltip about required mission start

commy2:
- Adds this revolutionary new concept called "tooltip" explaining why you cannot edit the addon settings or addon keybinds in main menu or mission select screen

### DIFF
--- a/addons/common/stringtable.xml
+++ b/addons/common/stringtable.xml
@@ -10,5 +10,9 @@
             <French>Community Base Addons - Composants communs</French>
             <Polish>Community Base Addons - Ogólne Komponenty</Polish>
         </Key>
+        <Key ID="STR_cba_common_need_mission_start">
+            <English>You must first start a mission.</English>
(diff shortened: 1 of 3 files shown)


addons/common/stringtable.xml
```diff
@@ -10,5 +10,9 @@
             <French>Community Base Addons - Composants communs</French>
             <Polish>Community Base Addons - Ogólne Komponenty</Polish>
         </Key>
+        <Key ID="STR_cba_common_need_mission_start">
+            <English>You must first start a mission.</English>
```
Dystopian: Dosn't it sound too personal? Maybe something like `Mission must be started.`?

commy2: >An example of a verb used in the imperative mood is the English sentence "Please be quiet". Such imperatives imply a second-person subject (you)[.]

Using the imperative is the best way to guide the user by tooltip in my experience.

PR #963, opened 2018-08-20
Settings - More appropriate slider speeds

mharis001:
- Change color setting sliders speed to an appropriate value
- Currently, pressing one of the arrows moves the slider all the way left/right


commy2:
The same for the other slider?

mharis001:
That would have to depend on the range, currently it always changes by 1 using the arrows. Could make that 5%(?) of the range.

commy2:
Yeah, same as this one basically. 5 and 10 %.

PR #965, opened 2018-08-21
Add CBA_fnc_compatibleMagazines that supports mag wells

PabstMirror:
Because of mag wells it's no longer safe to do simple code like
`private _weaponMagazines = getArray (configFile >> "CfgWeapons" >> _weapon >> "magazines");`

The filtering gets expensive, and might not always be needed, so I added a param for it
`rhs_weap_m27iar_grip` with 57 filtered entries is 0.26 ms
fast way is 0.028 ms but produces 69 entries because of duplicates and no guarantees that `in` is safe to use


commy2:
For consistency (https://github.com/CBATeam/CBA_A3/blob/master/addons/jr/fnc_compatibleItems.sqf) I would change the name to compatibleMagazines (remove "get").

I think it would be worth it to implement caching: https://github.com/CBATeam/CBA_A3/blob/master/addons/jr/fnc_compatibleItems.sqf#L30-L36
https://github.com/CBATeam/CBA_A3/blob/master/addons/jr/fnc_compatibleItems.sqf#L62

commy2:
The function should also report all entries lower case.
And with caching implemented, it should always filter duplicates. Not sure about non existing. Usually that is something that should error, although since that is a feature of magwells, it probably should just always filter them out if they're not present.

PabstMirror:
I thought it would be harder to cache because I wanted to support sub-muzzles like 
`configFile >> "CfgWeapons" >> "m4" >> "M203"`, but it was pretty easy

I like config capitalization because you can use `in` but `findIf {_x == z}` is fine as well; I can change if needed.

commy2:
I guess the JAM component, but that doesn't exist yet. Common is fine imo.
Config case is good too and would be consistent with compatibleItems, but it seems that most of the time BI uses all lower case. E.g. how they changed `getAllHitPointsDamage` to all lower case.

commy2:
All lower case means `toLower _mag in _mags` is possible.

commy2:
All lower case means `toLower _mag in _mags` is possible.
Config case this would look:
`configName (configFile >> "CfgMagazines" >> _mag) in _mags` and I don't think many people would use that, despite it being not all that bad.

PabstMirror:
Config case would simplify a func like this
https://github.com/acemod/ACE3/blob/master/addons/reloadlaunchers/functions/fnc_getLoadableMissiles.sqf

commy2:
Whatever we go with, it should be clarified in the function header.

dedmen:
ACE Arsenal would prefer configCase I think.
People can always make it lowercase quite easily and quickly if they need it.
However turning it back to config case isn't that easy nor quick.

addons/common/fnc_compatibleMagazines.sqf
```diff
@@ -0,0 +1,54 @@
+#include "script_component.hpp"
+/* ----------------------------------------------------------------------------
+Function: CBA_fnc_compatibleMagazines
+
+Description:
+    Retrievs a list of magazines that are compatible with a weapon.
```
dedmen: Retrieves

addons/common/fnc_compatibleMagazines.sqf
```diff
@@ -0,0 +1,54 @@
...
+};
+
+private _cacheKey = str _weapon;
+if (_cacheKey == "") exitWith { ERROR_1("Weapon Does Not Exist %1",_this); [] };
+
+if (isNil QGVAR(magNamespace)) then { GVAR(magNamespace) = call CBA_fnc_createNamespace; };
+
+private _returnMags = GVAR(magNamespace) getVariable _cacheKey;
+
+if (isNil "_returnMags") then {
```
Dystopian: this could be replaced with early return
`if (!isNil "_returnMags") exitWith {_returnMags};`

commy2: Don't forget to copy the returned array, otherwise someone will mess up the cache by editing it by reference.

PR #967, opened 2018-08-26
Settings - Add time setting type

mharis001:
- title, preview:
cba_time_setting


addons/settings/fnc_check.sqf
```diff
@@ -40,5 +40,9 @@ switch (toUpper _settingType) do {
     case "COLOR": {
         _value isEqualType [] && {count _value == count _defaultValue} && {_value isEqualTypeAll 0} && {{_x < 0 || _x > 1} count _value == 0}
     };
+    case "TIME": {
+        _settingData params ["_min", "_max"];
+        _value isEqualType 0 && {_value >= _min} && {_value <= _max} && {round _value == _value}
```
commy2: `round _value == _value`
Why?

addons/settings/fnc_init.sqf
```diff
@@ -35,7 +35,10 @@ Examples:
         ["Test_Setting_4", "COLOR",    ["-test color-",    "-tooltip-"], "My Category", [1, 1, 0]] call cba_settings_fnc_init;
 
         // EDITBOX --- extra argument: default value
-        ["Test_Setting_5", "EDITBOX", ["-test editbox-", "-tooltip-"], "My Category", "defaultValue"] call cba_settings_fnc_init;
+        ["Test_Setting_5", "EDITBOX",  ["-test editbox-", "-tooltip-"], "My Category", "defaultValue"] call cba_settings_fnc_init;
+
+        // TIME PICKER --- extra arguments: [_min, _max, _default]
+        ["Test_Setting_6", "TIME",     ["-test time-",    "-tooltip-"], "My Category", [0, 3600, 60]] call cba_settings_fnc_init;
```
commy2: Should mention "time in seconds" somewhere.

addons/settings/script_component.hpp
```diff
@@ -5,9 +5,9 @@
 #include "\a3\ui_f\hpp\defineCommonGrids.inc"
 #include "\a3\ui_f\hpp\defineResincl.inc"
 
-//#define DEBUG_MODE_FULL
-//#define DISABLE_COMPILE_CACHE
-//#define DEBUG_ENABLED_SETTINGS
+#define DEBUG_MODE_FULL
+#define DISABLE_COMPILE_CACHE
+#define DEBUG_ENABLED_SETTINGS
```
commy2: reminder to disable

addons/settings/fnc_check.sqf
mharis001: Since the visual doesn't show decimals, checking if rounded like the default slider

commy2: If `SLIDER` does the same, then I suppose this is correct.

addons/settings/gui.hpp
```diff
@@ -449,6 +449,78 @@ class GVAR(Row_ColorAlpha): GVAR(Row_Color) {
...
+            y = POS_H(1.1) + TABLE_LINE_SPACING / 2;
+            w = POS_W(6);
+            h = POS_H(0.9);
+            sizeEx = POS_H(1);
+            colorBackground[] = {0, 0, 0, 0.2};
+        };
+        class Hours: RscEdit {
+            idc = IDC_SETTING_TIME_HOURS;
+            style = ST_CENTER + ST_NO_RECT;
+            text = "00";
```
commy2: Omg I need this for the SLIDER.

commy2:
The control does not correctly tick "overwrite client" when altered in the MISSION tab, and does not highlight as yellow when changed, both if the edit boxes were used.
This is done by the SLIDER by using:
```sqf
    // automatically check "overwrite client" for mission makers qol
    [_controlsGroup, _source] call (_controlsGroup getVariable QFUNC(auto_check_overwrite));
```
in the keyUp event. The other purpose of the keyup event is to trim non number chars, which is done beautifully with the `text = "00";` config entry here instead. I wonder if the entry works with commas.

mharis001:
Will add the line to indicate a changed value to the KillFocus event. Like the 3DEN attribute, the value does not update with a KeyUp event. A KeyUp event could be added but I think the UX might be better without it.

Also, unfortunately, the `text = "00";` config entry does not trim non-number chars. That is just the default value of the edit box, I added it to test visuals when creating (not really needed anymore, so will get rid of). Add filtering of only number chars through KeyUp event?

addons/settings/fnc_gui_settingTime.sqf
```diff
@@ -0,0 +1,114 @@
...
+    (_ctrlSlider getVariable QGVAR(params)) params ["_setting", "_source"];
+    _value = round _value;
+
+    private _controlsGroup = ctrlParentControlsGroup _ctrlSlider;
+
+    {
+        _x params ["_idc", "_value"];
+
+        private _ctrlEdit = _controlsGroup controlsGroupCtrl _idc;
+        private _editText = if (_value < 10) then {format ["0%1", _value]} else {str _value};
```
mharis001: Replace with `CBA_fnc_formatNumber`?

commy2: Could do, but this works too. Only difference I believe would be rounding instead of truncating.

SilentSpike:
I would personally remove the slider all together for these settings because it clutters the UI.

I can't decide if I like including upper and lower limits for these 🤔 My gut says that time settings shouldn't be bound by limits (with the exception that negative time shouldn't be allowed) because otherwise why not just use a slider?

commy2:
SLIDER doesn't give feedback how long something takes in HH:MM:SS format.

SilentSpike:
On further consideration I think ideally the min and max should be optional for time settings. There are many use cases where you might want the user to select a time, but it doesn't matter how long/short it is.

I'd still remove the slider in the menu either way though.

commy2:
TIME vs DATE?

SilentSpike:
Nah, in general I'd say if you want a time from the user you wouldn't want to limit it.

Example: Cardiac arrest timer in ACE medical rewrite, prime use case for a time setting, but there's no purpose in limiting how long people should be able to set it.

commy2:
That's true for all sliders I suppose, but you could always make the maximum something unreasonable like 1 or maybe even 10 hours.

SilentSpike:
You've swayed me on the sliders 😄

Which in turn means limits make sense. No more considerations here from me 👍

mharis001:
I think this is done. Only thing maybe would adding a KeyUp event for trimming non-number chars and updating value on key press but I don't think it is necessary.

PR #972, opened 2018-08-31
Keybinding - Highlight individual duplicate keys

mharis001:
- title
- Match tooltip box of `_edit` and `_assigned`
- Before: 
cba_keybind_highlight_1
- After: 
cba_keybind_highlight_2

### DIFF
--- a/addons/keybinding/fnc_gui_update.sqf
+++ b/addons/keybinding/fnc_gui_update.sqf
@@ -64,8 +64,6 @@ private _tablePosY = 0;
     _keybinds = _tempNamespace getVariable [_action, _keybinds];
 
     private _keyNames = [];
-    private _isDuplicated = false;
-
     {
         private _keybind = _x;
         private _keyName = _keybind call CBA_fnc_localizeKey;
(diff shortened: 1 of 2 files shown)


addons/keybinding/fnc_gui_update.sqf
```diff
@@ -77,17 +75,17 @@ private _tablePosY = 0;
...
                 _duplicateKeybinds = _tempNamespace getVariable [_duplicateAction, _duplicateKeybinds];
 
-                if (_keybind in _duplicateKeybinds && {_action != _duplicateAction}) exitWith {
-                    _isDuplicated = true;
-                };
-            } forEach _addonActions;
-
-            _keyNames pushBack _keyName;
+                _keybind in _duplicateKeybinds && {_action != _duplicateAction}
+            } > -1) then {
```
commy2: Very cool. Apparently Travis doesn't understand  `if findIf`.

addons/keybinding/fnc_gui_update.sqf
```diff
@@ -123,12 +121,10 @@ private _tablePosY = 0;
...
     private _assigned = _subcontrol controlsGroupCtrl IDC_KEY_ASSIGNED;
-    _assigned ctrlSetText (_keyNames joinString ", ");
+    _assigned ctrlSetStructuredText parseText (_keyNames joinString ", ");
     _assigned ctrlSetTooltip _tooltip;
-
-    if (_isDuplicated) then {
-        _assigned ctrlSetTextColor [1,0,0,1];
-    };
+    _assigned ctrlSetTooltipColorBox [1,1,1,1];
+    _assigned ctrlSetTooltipColorShade [0,0,0,0.7];
```
commy2: These already were the values used I believe. Although setting them explicitly may be more robust if future changes happen (e.g. by BI).

mharis001: `RscStructuredText`, unlike `RscText`, has no entries for tooltip defined, and the defaults look different to these. Should probably just move these to config though.

commy2:
Fixed for the code validation. `_isDuplicated` isn't a boolean, but it does look easier to read like this. Idk.

PR #974, opened 2018-09-07
add searchbar to mission browser

commy2:
- same as SP scenario browser


addons/ui/fnc_initDisplayRemoteMissions.sqf
```diff
@@ -0,0 +1,93 @@
...
+
+params ["_display"];
+private _ctrlMaps = _display displayCtrl IDC_SERVER_ISLAND;
+private _ctrlMissions = _display displayCtrl IDC_SERVER_MISSION;
+
+ctrlPosition _ctrlMissions params ["_left", "_top", "_width", "_height"];
+
+private _ctrlSearch = _display ctrlCreate ["RscEdit", IDC_SEARCH];
+_ctrlSearch ctrlSetPosition [
+    _left,
```
mharis001: `_left + 0.1 * GUI_GRID_W`

addons/ui/fnc_initDisplayRemoteMissions.sqf
```diff
@@ -0,0 +1,93 @@
...
+    _top,
+    _width - GUI_GRID_W,
```
mharis001: `_width - 1.2 * GUI_GRID_W`

addons/ui/fnc_initDisplayRemoteMissions.sqf
```diff
@@ -0,0 +1,93 @@
...
+    GUI_GRID_H
```
mharis001: `0.9 * GUI_GRID_H`

addons/ui/fnc_initDisplayRemoteMissions.sqf
```diff
@@ -0,0 +1,93 @@
...
+_ctrlMissions ctrlSetPosition [
+    _left,
+    _top + GUI_GRID_H,
+    _width,
+    _height - GUI_GRID_H
+];
+_ctrlMissions ctrlCommit 0;
+
+// store all missions of the currently selected map
+private _fnc_storeMapMissions = {_this spawn {isNil { // delay a frame
```
dedmen: why no execNextFrame? spawn could delay multiple frames.

commy2: Doesn't work before postInit, and there is no postInit in i.e. 3den.

commy2:
Kept the size of the edit box, because it looked a bit squashed, and moved the listbox below down a bit instead.

PR #976, opened 2018-09-08
add Addon Options button to main menu

commy2:
- adds Addon Options to main menu directly, because why not


xetra11:
love the reason

PR #978, opened 2018-09-09
Let whitelisted UIDs save server settings

neilzar:
- Fix `fnc_set` to allow whitelisted to save server settings


addons/settings/fnc_set.sqf
```diff
@@ -132,7 +132,7 @@ switch (toLower _source) do {
 
             [QGVAR(refreshSetting), _setting] call CBA_fnc_globalEvent;
         } else {
-            if (IS_ADMIN_LOGGED) then {
+            if (FUNC(whitelisted)) then {
```
commy2: I think it should be:
```sqf
if (call FUNC(whitelisted)) then {
```
, because `if CODE` doesn't work.

PR #979, opened 2018-09-13
German translation, part 1

commy2:
- adds German translations
- removes one `original`, which we never make use of anywhere else
- fixd English spelling

### DIFF
--- a/addons/accessory/stringtable.xml
+++ b/addons/accessory/stringtable.xml
@@ -3,6 +3,7 @@
     <Package name="accessory">
         <Key ID="STR_CBA_accessory_Component">
             <English>Community Base Addons - Accessory Functions</English>
+            <German>Community Base Addons - Zubehörfunktionen</German>
             <Japanese>Community Base Addons - アクセサリー機能</Japanese>
             <Chinese>社群基礎模組 - 附件功能</Chinese>
             <Chinesesimp>社群基础模组 - 附件功能</Chinesesimp>
@@ -12,6 +13,7 @@
         </Key>
         <Key ID="STR_CBA_accessory_railNext">
             <English>Next rail item state</English>
+            <German>Nächster Zustand Schienenaufsatz</German>
             <Japanese>レイル アイテムを次モード</Japanese>
             <Chinese>下個導軌物品狀態</Chinese>
             <Chinesesimp>下个导轨物品状态</Chinesesimp>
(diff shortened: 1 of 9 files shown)


addons/common/stringtable.xml
```diff
@@ -3,6 +3,7 @@
     <Package name="Common">
         <Key ID="STR_CBA_Common_Component">
             <English>Community Base Addons - Common Component</English>
+            <German>Community Base Addons - Gemeinsame Komponente</German>
```
xetra11: auch: Allgemeine Komponente

commy2: I always understood this to be the "shared component" - stuff that is used in multiple places. That is at least how we treated it in ACE: stuff that is used in at least two places gets moved to ACE Common, although I realize that not everything in CBA Common or ACE Common is shared between two other components or even used by a single other one.

I would've used "Allgemeine Komponente" if it were named "CBA General" instead of "CBA Common".

addons/accessory/stringtable.xml
```diff
@@ -12,6 +13,7 @@
         </Key>
         <Key ID="STR_CBA_accessory_railNext">
             <English>Next rail item state</English>
+            <German>Nächster Zustand Schienenaufsatz</German>
```
dedmen: Could "Modus" be used here? Or atleast "Nächster Schienenaufsatz Zustant"?
This sounds very weird.

dedmen: STR_CBA_accessory_railNext_tooltip also uses "Modus"

addons/accessory/stringtable.xml
```diff
@@ -48,6 +53,7 @@
         </Key>
         <Key ID="STR_CBA_accessory_opticNext">
             <English>Next optics state</English>
+            <German>Nächster Zustand Optik</German>
```
dedmen: Again the "Modus" talk. `STR_CBA_accessory_opticNext_tooltip` uses modus.

addons/accessory/stringtable.xml
Blutze: +1 für mode/Modus.
Unabhängig davon: entweder "Zustand des Aufsatzes" oder "Aufsatzzustand". Nix mit Leerzeichen.

addons/common/stringtable.xml
xetra11: Yea from this perspective you convinced me. Good job. Go into politics! 📦

addons/accessory/stringtable.xml
Soldia1138: +1 for "Modus"

commy2: >STR_CBA_accessory_railNext_tooltip also uses "Modus"

ENG uses "state" in STR_CBA_accessory_railNext and "mode" in STR_CBA_accessory_railNext_tooltip.
state - Zustand, mode - Modus.

>entweder "Zustand des Aufsatzes" oder "Aufsatzzustand". Nix mit Leerzeichen.

They're not compound nouns how I used them. Just two successive nouns. Exactly like in the English original.
ENG is not in possessive form either, so why would I change it to genitive in German? It uses the nominative case, so I did the same.

PR #980, opened 2018-09-14
German translation, part 2

commy2:
- see #979

### DIFF
--- a/addons/main/stringtable.xml
+++ b/addons/main/stringtable.xml
@@ -14,14 +14,19 @@
         </Key>
         <Key ID="STR_CBA_URL">
             <English>https://www.github.com/CBATeam/CBA_A3</English>
-            <German>https://www.github.com/CBATeam/CBA_A3</German>
-            <Japanese>https://www.github.com/CBATeam/CBA_A3</Japanese>
-            <Chinese>https://www.github.com/CBATeam/CBA_A3</Chinese>
-            <Chinesesimp>https://www.github.com/CBATeam/CBA_A3</Chinesesimp>
-            <Portuguese>https://www.github.com/CBATeam/CBA_A3</Portuguese>
-            <Russian>https://www.github.com/CBATeam/CBA_A3</Russian>
             <French>https://www.github.com/CBATeam/CBA_A3</French>
+            <Spanish>https://www.github.com/CBATeam/CBA_A3</Spanish>
+            <Italian>https://www.github.com/CBATeam/CBA_A3</Italian>
             <Polish>https://www.github.com/CBATeam/CBA_A3</Polish>
+            <Russian>https://www.github.com/CBATeam/CBA_A3</Russian>
+            <German>https://www.github.com/CBATeam/CBA_A3</German>
+            <Czech>https://www.github.com/CBATeam/CBA_A3</Czech>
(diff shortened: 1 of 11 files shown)


addons/xeh/stringtable.xml
```diff
@@ -3,6 +3,7 @@
     <Package name="XEH">
         <Key ID="STR_CBA_XEH_Component">
             <English>Community Base Addons - Extended Event Handlers</English>
+            <German>Community Base Addons - Erweiterte Ereignishandler</German>
```
bux: Ereignisbehandlung

commy2: Apparently it's a toss up: https://www.dict.cc/english-german/event+handler.html

addons/main/stringtable.xml
```diff
@@ -14,6 +14,7 @@
         </Key>
         <Key ID="STR_CBA_URL">
             <English>https://www.github.com/CBATeam/CBA_A3</English>
+            <German>https://www.github.com/CBATeam/CBA_A3</German>
```
dedmen: Why not just simply use a single <Original> entry here as every entry is literally the same

addons/settings/stringtable.xml
```diff
@@ -3,6 +3,7 @@
     <Package name="Settings">
         <Key ID="STR_cba_settings_Component">
             <English>Community Base Addons - Settings Component</English>
+            <German>Community Base Addons - Einstellungskomponente</German>
```
dedmen: This says component. But music/network just don't.
`Community Base Addons - Einstellungen` ?

addons/main/stringtable.xml
commy2: This is a string so it can be changed if needed everywhere at once, while not having to use a macro which would have character escaping issues. It is translated I guess mainly so Jonpas' translation counter bot says 100% eventually.

addons/settings/stringtable.xml
commy2: I like this more. Should I change it in English as well?

addons/main/stringtable.xml
jonpas: Flummi's bot. ;)
But yes.

commy2:
Fixed some missing apostrophes in English while editing the file. I at least think those should be placed there.

addons/xeh/stringtable.xml
BaerMitUmlaut: I wouldn't translate this at all since it's a code thing. You didn't translate strings to "Zeichenketten" either.

bux: Just be consistent.

_https://help.sap.com/doc/abapdocu_752_index_htm/7.52/de-DE/abenevent_handler_glosry.htm_
_https://de.wikipedia.org/wiki/Ereignis_(Programmierung)_

commy2: >Auch die Ereignisbehandlungsroutinen (englisch event handler, deutsch Ereignishandler) und die Ereignisse selbst werden dann als Objekte modelliert.

jokoho48: when you want to keep it then please also translate String to "Zeichenketten"

commy2: ^ done

PR #981, opened 2018-09-14
German translation, part 3

commy2:
- see #979, #980
- use some base game translations instead of keeping copies in CBA

### DIFF
--- a/addons/modules/stringtable.xml
+++ b/addons/modules/stringtable.xml
@@ -3,6 +3,7 @@
     <Package name="Modules">
         <Key ID="STR_CBA_Modules_Component">
             <English>Community Base Addons - Modules Component</English>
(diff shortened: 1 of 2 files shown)


addons/modules/stringtable.xml
```diff
@@ -30,6 +33,7 @@
         </Key>
         <Key ID="STR_CBA_Modules_AttackLocType">
             <English>Attack Location Type</English>
+            <German>?????</German>
```
dedmen: That doesn't seem correct. Also very hard to pronounce.
"Fragenzeichen Fragezeichen Fragezeichen Fragezeichen Fragezeichen"
Seems very long.

commy2: Placeholder, because WIP, and it may already be in the game.

commy2:
done

PR #982, opened 2018-09-14
Add Feature Camera Player EH

jonpas:
- Port feature camera checker from ACE3
- Add `"featureCamera"` Player Event Handler
- Allow registering custom feature cameras

### DIFF
--- a/addons/common/fnc_registerFeatureCamera.sqf
+++ b/addons/common/fnc_registerFeatureCamera.sqf
@@ -10,7 +10,7 @@ Parameters:
     _callback - Activity check (should return true if active, false otherwise) <CODE>
 
 Returns:
-    None
+    Successfully registered <BOOL>
 
 Examples:
     (begin example)
@@ -25,4 +25,16 @@ Author:
 ---------------------------------------------------------------------------- */
 SCRIPT(registerFeatureCamera);
 
-GVAR(featureCameras) pushBackUnique _this;
+params [["_name", "", [""]], ["_callback", false, [{}]]];
+
+if (_name isEqualTo "") exitWith {
+    TRACE_1("Name empty",_name);
+    false
+};
+
+if (_callback isEqualTo false) exitWith {
+    TRACE_1("Callback not code",_name);
+    false
+};
+
+(GVAR(featureCameras) pushBackUnique _this) != -1 // return
--- a/addons/common/fnc_registerFeatureCamera.sqf
+++ b/addons/common/fnc_registerFeatureCamera.sqf
@@ -1,0 +1,28 @@
+#include "script_component.hpp"
+/* ----------------------------------------------------------------------------
+Function: CBA_fnc_registerFeatureCamera
+
+Description:
+    Registers a feature camera for use with CBA_fnc_getActiveFeatureCamera.
+
+Parameters:
+    _name     - Name (unique/tagged) <STRING>
+    _callback - Activity check (should return true if active, false otherwise) <CODE>
+
+Returns:
+    None
+
+Examples:
+    (begin example)
+        _result = [
+            "ace_spectator",
+            {!isNull (missionNamespace getVariable ["ace_spectator_isSet", objNull])}
+        ] call CBA_fnc_registerFeatureCamera
+    (end)
+
+Author:
+    Sniperwolf572, Jonpas
+---------------------------------------------------------------------------- */
+SCRIPT(registerFeatureCamera);
+
+GVAR(featureCameras) pushBackUnique _this;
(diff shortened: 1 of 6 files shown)


addons/common/fnc_registerFeatureCamera.sqf
```diff
@@ -0,0 +1,28 @@
...
+            {!isNull (missionNamespace getVariable ["ace_spectator_isSet", objNull])}
+        ] call CBA_fnc_registerFeatureCamera
+    (end)
+
+Author:
+    Sniperwolf572, Jonpas
+---------------------------------------------------------------------------- */
+SCRIPT(registerFeatureCamera);
+
+GVAR(featureCameras) pushBackUnique _this;
```
jonpas: Should probably add argument checking.

addons/events/fnc_addPlayerEventHandler.sqf
```diff
@@ -192,6 +200,12 @@ if (_id != -1) then {
                 GVAR(oldCameraView) = _data;
                 [QGVAR(cameraViewEvent), [_player, _data]] call CBA_fnc_localEvent;
             };
+
+            _data = call CBA_fnc_getActiveFeatureCamera;
```
dedmen: performance?
We might want to only add this when it's actually needed.

jonpas: It's just a bunch of null checks (7 in vanilla). Aren't all Player EHs only added when they are needed already?

jonpas: 0.0301 - 0.31 ms on 10k/10k cycles.

PabstMirror:
https://github.com/CBATeam/CBA_A3/compare/camera-playerEH...featureCamOpti?expand=1
Split arrays, simplify main loop:
0.0297 ms vs 0.0195 ms

addons/events/fnc_addPlayerEventHandler.sqf
dedmen: 0.03 ouhhhh.
How about a check whether either the array of eventhandlers is not empty (by using the CBA internal variable name of the array)
Or check that `GVAR(oldFeatureCamera` is not nil? and set it to non nil once a Eventhandler is added in line 92?

jonpas: That wouldn't work for cases where we might want to use internal variable (ref ACE PR).

commy2:
Comments on the commits, not the PR :x

jonpas:
Well yes, they were made before any PR was open. How are you supposed to comment on the PR if it doesn't exist?

addons/events/fnc_addPlayerEventHandler.sqf
```diff
@@ -214,6 +222,15 @@ if (_id != -1) then {
...
         };
+
+        GVAR(playerEHInfo) pushBack ([{call FUNC(playerEH_HalfSecond)}, 0.5] call CBA_fnc_addPerFrameHandler);
+        [QFUNC(playerEH_HalfSecond), {
+            private _data = call CBA_fnc_getActiveFeatureCamera;
+            if !(_data isEqualTo GVAR(oldFeatureCamera)) then {
+                GVAR(oldFeatureCamera) = _data;
+                [QGVAR(featureCameraEvent), [call CBA_fnc_currentUnit, _data]] call CBA_fnc_localEvent;
+            };
+        }] call CBA_fnc_compileFinal;
```
commy2: There is no reason to compileFinal this. It is only done for Map, because it is a mission event handler and therefore affected by the recompiled-every-execution-bug. That however does not apply to CBA PFH.

jonpas:
I have tested it, works perfectly fine with ACE3 (with https://github.com/acemod/ACE3/pull/6573) and ACRE2 (with https://github.com/IDI-Systems/acre2/pull/582).

addons/events/fnc_addPlayerEventHandler.sqf
```diff
@@ -214,6 +222,15 @@ if (_id != -1) then {
                 };
             } call CBA_fnc_directCall;
         };
+
+        GVAR(playerEHInfo) pushBack ([{
```
dedmen: The handler is still always added even if noone is listening for it.
Seems to be fairly easy to just add it in L92 instead.

jonpas: We could do this since https://github.com/acemod/ACE3/commit/922774d972f304f1c6edc6d2b3c49725030a16b4. Unsure how worth it is though, ACE3 is already set to use it.

PR #985, opened 2018-09-15
ProjectileTracking - Stop tracking stationary objects

PabstMirror:
Ref https://github.com/acemod/ACE3/issues/6565

### DIFF
--- a/addons/diagnostic/fnc_projectileTracking_trackProjectile.sqf
+++ b/addons/diagnostic/fnc_projectileTracking_trackProjectile.sqf
@@ -28,6 +28,10 @@ params ["_args", "_handle"];
 _args params ["_projectile", "_index", "_initialProjectileData"];
 
 if (!isNull _projectile) then {
+    private _speed = vectorMagnitude velocity _projectile;
+    if (_speed < 0.1) then { // If projectile is slowed down, stop tracking after this run
+        [_handle] call CBA_fnc_removePerFrameHandler;
+    };
 
     private _data = [];
     private _bulletData = [];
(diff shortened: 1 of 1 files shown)


addons/diagnostic/fnc_projectileTracking_trackProjectile.sqf
```diff
@@ -39,13 +43,9 @@ if (!isNull _projectile) then {
     } else {
         _bulletData = [_initialProjectileData];
         _data = [_handle, _bulletData];
+        GVAR(projectileData) set [_index, _data];
     };
-
-    _bulletData pushBack [(getPos _projectile), vectorMagnitude (velocity _projectile)];
-
-    _data set [1, _bulletData];
```
bux: did you delete this line intentional?

addons/diagnostic/fnc_projectileTracking_trackProjectile.sqf
```diff
@@ -39,13 +43,9 @@ if (!isNull _projectile) then {
...
-
-    GVAR(projectileData) set [_index, _data];
```
bux: same here? outside of the if block?

commy2: These are not needed, because only commands that modify the array are used. Previously the function set a sub array, and a sub sub array to exactly what they were before.

addons/diagnostic/fnc_projectileTracking_trackProjectile.sqf
PabstMirror:
```
{
    _bulletData = _data select 1;
} else {
    _data = [_handle, _bulletData];
};

_data set [1, _bulletData];
```
either way `_bulletData` is going to already be set in `_data`

Issue #108, opened 2015-08-07
Add Joint Ammo and Magazines (JAM)


ViperMaul:
GOAL: 
Provide common bullet and magazine classes for use by weapon mods in order to achieve better consistency and compatibility

TO DO:
- add common bullets (CfgAmmo classes)
- add magazines (CfgMagazines classes)
- add magazine groups
- provide macros as alternative to magazine groups so mod makers can use them
- configure vanilla weapons to use new magazines
- add support for AB, ACE3

TO NOT DO:
- create too many bullet and magazine variations

Source: https://github.com/robalo/mods/blob/master/asdg_jam/config.cpp
Relates to #43

aoighost:
VM, I want to look in to this, should I do it in a fork of this or JAM?

dedmen:
Problem is that magazineGroups are broken right? Because https://github.com/robalo/mods/blob/master/asdg_jam/config.cpp#L197 is uinsg magazineGroups.
Can someone confirm that they are broken?

robalo:
Last I checked they were broken, long ago but did not see any mention of them since. But this can be done without magazineGroups, carefully appending to magazines[] config array instead.  I'd still like to work on this, only need to find some time to kick-start it. Then I'm sure we'll get people contributing to it fast.

dedmen:
Just tested `magazineGroup` is working fine in the backend.
Just mousewheel action and drag-drop into magazine slot is not working perfectly. See https://feedback.bistudio.com/T82996#1650317

Mouse wheel menu is fixable with script.

dedmen:
I fixed it.
https://feedback.bistudio.com/T82996#1650367
Just took about 3 hours of work. Only possible with a non Battleye compatible hack. But maybe BI will just let me explain them how to fix it and implement that.

dedmen:
Official statement from reyhard that BI is working on magazineGroups (Now called magazineWells).
```
class CfgMagazineWells
{
    class Stanag556x45
    {
        BI_stanags[] = {30Rnd_556x45_Stanag,30Rnd_556x45_Stanag_Tracer_Red,30Rnd_556x45_Stanag_Tracer_Green,30Rnd_556x45_Stanag_Tracer_Yellow,30Rnd_556x45_Stanag_red,30Rnd_556x45_Stanag_green};
        CUP_stanags[] = {CUP_30Rnd_556x45_Stanag1, CUP_30Rnd_556x45_Stanag2};            
    }; 
};

[...]

PR #990, opened 2018-09-19
Split/extend the magwell classes and files

robalo:
Extend JAM based on discussion in #108


commy2:
lgtm
Anyone willing to write a wiki page about this?

PR #991, opened 2018-09-19
add WW2 era magwells

Drofseh:
- Add WW2 era magwells that should cover all current IFA3, FOW, LEN weapons
- Also add magwells for non-magazine fed shotguns.
All shotguns should use every magwell that is <= its tube capacity.

### DIFF
--- a/addons/jam/CfgMagazineWells.hpp
+++ b/addons/jam/CfgMagazineWells.hpp
@@ -6,26 +6,55 @@ class CfgMagazineWells {
     #include "magwells_762x39.hpp"
     #include "magwells_762x51.hpp"
     #include "magwells_762x54.hpp"
+    #include "magwells_77x58.hpp"
+    #include "magwells_792x57.hpp"
+    #include "magwells_145x114.hpp"
+    #include "magwells_303B.hpp"
+    #include "magwells_3006.hpp"
+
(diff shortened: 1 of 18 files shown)


commy2:
Lets all be optimistic about anyone ever using these.

dedmen:
Can we get more info about what these are
https://github.com/CBATeam/CBA_A3/blob/master/addons/jam/magwells_762x54.hpp#L15-L18
? SVD I know.. But the rest?

Drofseh:
@dedmen
The DPM covers the DP-27, DP-28, and DPM which was the Russian light machine gun in WW2
https://upload.wikimedia.org/wikipedia/commons/1/13/Machine_gun_DP_MON.jpg

The DT is another version of the DP that was meant for use as a vehicle weapon but was also used in the LMG role by the VDV during WW2, it has a larger magazine that probably fits in the DP but is to tall to use the DP sights.
https://vignette.wikia.nocookie.net/world-war-2/images/7/7f/DT_Machine_Gun_with_bipod.jpg

The Maxim gun is a machine gun used by 25+ nations from the 1880's onward.
Russia had the M1910 and Finland had the M09/21 and M/32-33 in 7.62x54r.
http://browningmgs.com/Maxim/Photos_Maxim/Maxim1910_01.jpg

PR #995, opened 2018-09-29
More JAM stuff

Drofseh:
- Sort #includes and their magwell classes in lexicographical order
- Put all calibers in their own included .hpp file (except manpad AT and AA launchers, not sure what to do with those yet.
- add a few new magwell calibers (.38 Special, .357 Mag, .408 Chey Tac, .45 Gap, 6.5 Creedmore, 6.5 Grendal)
- add magwells for Glock pistols.
- add 9x18mm Makarov calibre magwells as versions of the same weapons in other calibers.


Drofseh:
6.8 SPC and .300 Blackout have previously been using metric designations (`CBA_68x43_ACR` and `CBA_762x35_STANAG` respectively).

I only noticed this after I'd added imperial in `magwells_300BLK.hpp` and `magwells_68SPC.hpp`.

I've added the two pre-existing metric classnames to those files so nothing gets broken, but I think we should switch to imperial for those calibers as metric is not commonly used for them.

robalo:
762x39mm in STANAG mag ??

Drofseh:
Yup.
https://i.gyazo.com/e2a14060fcf546be454ed40d7f81a1b6.jpg

addons/jam/magwells_303B.hpp
```diff
@@ -1,4 +1,4 @@
-    class CBA_303B_LeeEn {};        // Lee Enfield
     class CBA_303B_BREN {};         // BREN gun
+    class CBA_303B_LeeEn {};        // Lee Enfield
```
dedmen: Just as a side note. We don't really need to shorten names to oblivion. Longer names don't hurt that much

dedmen:
>I've moved the two pre-existing metric classnames to those files so nothing gets broken, but I think we should switch to imperial for those calibers as metric is not commonly used for them.

I think it's enough to mark them with "deprecated, use X instead" for now.

robalo:
Wouldn't worry about deprecation much at this point. It's unreleased. Feel free to clean up extra crud.

dedmen:
People are already starting to implement it

Drofseh:
@commy2 I hadn't quite pushed the last changes to this PR

dedmen:
That's why you mark PR's as WIP :D
But you can take that for hacktoberfest ^^

Drofseh:
lol, I don't have the ability to add labels to CBA.

Drofseh:
More changes in #997

dedmen:
Not labels.
Just name your PR with `[WIP] PR name` and then later remove the wip tag from title

Drofseh:
Ahh, sure, I will do that in future.

PR #996, opened 2018-09-30
add lobby manager

commy2:
This is a tool for the editor to change the order and naming of groups and units as they will appear in the MP lobby.


addons/ui/fnc_openLobbyManager.sqf
```diff
@@ -0,0 +1,132 @@
...
+
+            _sort pushBack [_groupCounter, _unitCounter + _sortModifier, _index];
+        };
+    };
+
+    _sort sort true;
+
+    // apply sort to list via value, with value being the index in the sorted array
+    {
+        _ctrlSlots lbSetValue [_x#2, _forEachIndex];
```
commy2: Travis

dedmen: I think you got a typo here. Seems like you hit `#` by accident while you intended to write ` select `

jonpas: `#` is so much neater here.

commy2:
Done. Needs feedback.

bux:
Mark a few characters in the role description
image

Select another unit and observe how the same amount of characters are selected in the changed description
image

commy2:
-Characters remain marked when switching between entries

Dunno what to do about this. No commands for that. The obvious work around is to delete and recreate the edit box, but it looks worse in the script^^

xrufix:
Works fine for me.

**Feedback / Feature Requests**:

* When adding a Headless Client, empty side is displayed by default when entering the lobby. 
* Group-Callsign should be the same as in lobby, imo.
* Add abilty to automaticaly number slots either by group (as in vanilla), by side, or from top to bottom.
* Add ability to rename multiple slots at once or copy descriptions.

commy2:
>When adding a Headless Client, empty side is displayed by default when entering the lobby.

Is that a description of how it currently works, or how you would want it to work? It seems to me to be unrelated to the editor (this PR) and more of a thing of the lobby itself.

>Group-Callsign should be the same as in lobby, imo.

No, reason read: https://github.com/CBATeam/CBA_A3/pull/667#issuecomment-299422802 f

>Add ability to rename multiple slots at once or copy descriptions.

Can't you just use ctrl + c and ctrl + v? Idk how else this would be implemented.

xrufix:
@commy2 

* The lobby displays the side of the playable entity that was first added to the mission. When this is the HC, the empty page is shown. When cutting all player entities except HC and pasting them again, HC is the first entitiy and therefore the empty page is shown. I suspect this PR does something similar. Therefore it is related. It would be better to "paste" the HC as the last playable entity.

* I do not know much about Armas UI, is it impossible to select multiple list items?

commy2:
>HC

I see. Is the HC entity set to "playable"? If yes, then this is no problem, as it is also cutpasted with the rest in the desired order.

>I do not know much about Armas UI, is it impossible to select multiple list items?

Only one cursor possible.

commy2:
>Characters remain marked when switching between entries

fixed
>Give input field focus when switching entries

done

bux:
Groups lose their "Callsign" after pressing Ok.
image

bux:
Units of a squad will also get the "@GROUP" modifier to their "Role Description" not only the squad leader.

Drofseh:
> Units of a squad will also get the "@group" modifier to their "Role Description" not only the squad leader.

Having `@group` in the name of every unit is a good thing. 
https://github.com/CBATeam/CBA_A3/wiki/Name-Groups-in-Lobby
>Note that when respawn is disabled in the mission, every unit in the group should have the group name added to their Role Description, because killed units are removed from the lobby and therefore can no longer change the group name..

bux:
@commy2 I would love to either be able to edit the callsign here or have the "@" character set the callsign too.

commy2:
Adding the callsign attribute whenever a group is selected should be way to go I'd say.

commy2:
Has too many duplication bugs left due to wonky behavior of the Cut Paste functionality with groups with only one unit. I basically need to work around those using temporary team mates for the single unit group units, but that needs some frame delaying to work apparently, which isn't done or tried yet.

Krzmbrzl:
Couldn't we just implement a check if there are any groups with just one unit. If there are give an error message and don't do anything and if there aren't proceed as normal.

Or a very dirty solution-approach: For each group with only one unit one could add a unit to that group, perform the magic and delete the dummy units after having reordered the whole thing.

PR #998, opened 2018-10-02
Slot list manager Polish translation

veteran29:
- title
- Remove BOM character from beginning of the ui/stringtable.xml file


commy2:
Nice.

PR #999, opened 2018-10-03
Notification

synixebrett:
- Create a simple notify function that creates a small notification for mod and mission makers to use rather than using hint or write themselves

### DIFF
--- a/addons/ui/fnc_notify.sqf
+++ b/addons/ui/fnc_notify.sqf
@@ -1,0 +1,122 @@
+#include "script_component.hpp"
+/* ----------------------------------------------------------------------------
+Function: CBA_fnc_notify
+
+Description:
+    Displays a small notification to the player
+
+Parameters:
+    title of the notification, displayed in the top left
+    message displayed in the center
+    (optional) color of the notification title
+
+Returns:
+    Nothing
+
+Examples:
+    (begin example)
+        ["ARC", "Not Connected to Discord", [1,0,0,1]] call cba_fnc_notify;
+        ["CBA", "This text is white!"] call cba_fnc_notify;
+    (end)
+
+Author:
+    SynixeBrett
+
+---------------------------------------------------------------------------- */
+SCRIPT(notify);
+
+if !(hasInterface) exitWith {  };
+
+params [
+    ["_title", ""],
+    ["_message", ""],
+    ["_color", [1,1,1,1]]
+];
+
+if !(canSuspend) exitWith {[_title, _message, _color] spawn CBA_fnc_notify};
+
+disableSerialization;
+
+// Require a _title and _message
+
+if (_title isEqualTo "" && {_message isEqualTo ""}) exitWith {};
+
+private _display = findDisplay 46;
+
+private _index = uiNamespace getVariable [QGVAR(NotificationIndex), 0];
+uiNamespace setVariable [QGVAR(NotificationIndex), _index + 1];
+
+waitUntil {_index - GVAR(NotificationCurrentIndex) < GVAR(NotificationMax)};
+
+private _position = _index % GVAR(NotificationMax);
+private _safezoneX = safezoneW + safezoneX;
+
+// Create the 3 displays in the offscreen position
+
+private _ctrlBackground = _display ctrlCreate ["RscPicture", 20];
+_ctrlBackground ctrlSetPosition [
+    _safezoneX,
+    (0.24 + (_position * 0.08)) * safezoneH + (safezoneY * -0.5) ,
+    0.1 * safezoneW,
+    0.06 * safezoneH
+];
+_ctrlBackground ctrlSetText "#(argb,8,8,3)color(0,0,0,0.75)";
+_ctrlBackground ctrlCommit 0;
+
+private _ctrlTitle = _display ctrlCreate ["RscStructuredText", 10];
+_ctrlTitle ctrlSetPosition [
+    _safezoneX,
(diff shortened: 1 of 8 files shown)


jokoho48:
👎 don't like the use of spawn in there. should use waitAndExecute and waitUntilAndExecute maybe instead?

synixebrett:
I didn't use waitAndExecute since it has multiple blocks requiring sleep. This would require nesting would it not?

I thought the use of 3 times would be better overall then nesting 3 times

jokoho48:
I would build it more like a message queue so that you don't need to wait multiple times for every notification.

so that you have a function that just enqueues the notification, and one that pulls one and draws it and after x amount of time hides it and recalls itself to check if the queue is empty and when then just stops. with that, you only have X "tasks" running at the same time and not a "task" for every notification.

synixebrett:
Wouldn't that require a task running during the entire mission to check the queue? Where as this method has no background tasks running while there a no notifications.

Considering that these notifications would rarely be shown isn't this way better? Even though if 3 notifications are being shown, there are 3 tasks, there should rarely be 3 notifications.

jokoho48:
>Wouldn't that require a task running during the entire mission to check the queue? Where as this method has no background tasks running while there a no notifications.

no, because the function is called when you add a new notification. the draw function/tasks are only running when there is currently a notification on screen or in Queue. but than only the max amount of notifications that can get shown at the same time. and not more.

synixebrett:
> the function is called when you add a new notification

ah I see, I hadn't considered that way

addons/ui/fnc_notify.sqf
```diff
@@ -0,0 +1,122 @@
...
+uiNamespace setVariable [QGVAR(NotificationIndex), _index + 1];
+
+waitUntil {_index - GVAR(NotificationCurrentIndex) < GVAR(NotificationMax)};
+
+private _position = _index % GVAR(NotificationMax);
+private _safezoneX = safezoneW + safezoneX;
+
+// Create the 3 displays in the offscreen position
+
+private _ctrlBackground = _display ctrlCreate ["RscPicture", 20];
```
commy2: Magic variable should be a macro.

addons/ui/fnc_notify.sqf
```diff
@@ -0,0 +1,122 @@
...
+_ctrlBackground ctrlSetPosition [
+    _safezoneX,
+    (0.24 + (_position * 0.08)) * safezoneH + (safezoneY * -0.5) ,
+    0.1 * safezoneW,
+    0.06 * safezoneH
+];
+_ctrlBackground ctrlSetText "#(argb,8,8,3)color(0,0,0,0.75)";
```
commy2: I feel like this should use ctrlSetBackgroundColor instead. This as picture control seems awkward to me at least.

addons/ui/fnc_notify.sqf
```diff
@@ -0,0 +1,122 @@
...
+_ctrlBackground ctrlCommit 0;
+
+private _ctrlTitle = _display ctrlCreate ["RscStructuredText", 10];
```
commy2: Magic variable should be a macro.

addons/ui/fnc_notify.sqf
```diff
@@ -0,0 +1,122 @@
...
+    _safezoneX,
+    (0.24 + (_position * 0.08)) * safezoneH + (safezoneY * -0.5),
+    0.1 * safezoneW,
+    0.0255 * safezoneH
+];
+_ctrlTitle ctrlSetStructuredText parseText _title;
+_ctrlTitle ctrlSetTextColor _color;
+_ctrlTitle ctrlCommit 0;
+
+private _ctrlNotification = _display ctrlCreate ["RscStructuredText", 30];
```
commy2: Magic variable should be a macro.

addons/ui/fnc_notify.sqf
```diff
@@ -0,0 +1,122 @@
...
+_ctrlNotification ctrlSetPosition
+[
```
commy2: Inconsistent formatting. I prefer L57 & L67, and that is the style used in CBA in most places.

addons/ui/fnc_notify.sqf
```diff
@@ -0,0 +1,122 @@
...
+
+// move the displays onscreen
+
+private _ctrls = [_ctrlBackground, _ctrlTitle, _ctrlNotification];
+{
+    private _position = ctrlPosition _x;
+    private _posX = 0.9 * safezoneW + safezoneX;
+    _position set [0, _posX];
+    _x ctrlSetPosition _position;
+    _x ctrlCommit (uiNamespace getVariable [QGVAR(NotificationSpeed), 0.5]);
```
commy2: Why are these ui namespace? Doesn't settings put the variables in mission namespace? It seems weird to put this into ui namespace.

commy2:
Not happy about this being scheduled. Basically breaks if you save and load a savegame or have poor FPS in MP.

commy2:
This should be a controls group control template in config, so the script doesn't have to recreate the whole thing with slow sqf every time.

dedmen:
arma3_x64 2018-11-30 23-23-26-06
Screenshot from @commy2 
It's isn't of much use if it's in slack and not here ;)
The above obviously won't get merged like that

synixebrett:
That's what it looks like?

Wildly different from my PC.

I am going to be rewriting this PR after my exams anyway.

commy2:
I have a counter proposal here: #1036, which is kind of a port from the ACE one.

PR #1000, opened 2018-10-03
Remove BOM character from stringtables

veteran29:
- Remove BOM character from stringtable files

This repo `.editorconfig` is configured to encode files in UTF-8 **without** BOM, however currently most of the stringtables are encoded in UTF-8 **with** BOM character. From my tests it looks like it is due to usage of Tabler. 

Even when files are encoded without BOM the Tabler will save them with it.


bux:
https://github.com/bux/tabler/releases/tag/v0.8.0

PR #1001, opened 2018-10-04
Add french translations for CBA UI

alganthe:
- Add french translation for CBA UI.
- Add missing translation for joint Ammo Magazines.
- Add missing translation in strings, remove empty translations.

### DIFF
--- a/addons/strings/stringtable.xml
+++ b/addons/strings/stringtable.xml
@@ -28,8 +28,13 @@
         </Key>
         <Key ID="STR_CBA_FORMAT_NUMBER_THOUSANDS_SEPARATOR">
             <English>,</English>
+            <Czech> </Czech>
             <German>.</German>
-            <French>,</French>
+            <Polish> </Polish>
+            <Russian> </Russian>
+            <French> </French>
+            <Spanish> </Spanish>
+            <Italian> </Italian>
             <Japanese>,</Japanese>
             <Chinese>,</Chinese>
             <Chinesesimp>,</Chinesesimp>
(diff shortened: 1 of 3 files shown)


addons/ui/stringtable.xml
```diff
@@ -1,4 +1,4 @@
-﻿<?xml version="1.0" encoding="utf-8"?>
+<?xml version="1.0" encoding="utf-8"?>
```
commy2: Make sure this doesn't readd the BOM removed in https://github.com/CBATeam/CBA_A3/pull/1000

addons/strings/stringtable.xml
```diff
@@ -28,13 +28,8 @@
...
         <Key ID="STR_CBA_FORMAT_NUMBER_THOUSANDS_SEPARATOR">
             <English>,</English>
-            <Czech> </Czech>
             <German>.</German>
-            <Polish> </Polish>
-            <Russian> </Russian>
-            <French> </French>
-            <Spanish> </Spanish>
-            <Italian> </Italian>
+            <French>,</French>
```
commy2: Deleted entries which should not be deleted.

addons/ui/stringtable.xml
alganthe: i've used tabler 0.8.0, which shouldn't.

commy2: Why is it pink then?

alganthe: good question, no idea :/

alganthe: just checked, no BOM.

dedmen: Just checked #1000 it indeed converted UTF8-BOM to UTF8 in ui stringtable.
Also checked the merge commit. Both files have same encoding.. There was indeed no change here 🤔  @Github broken.

dedmen: However @alganthe changed the encoding from BOM to nonBOM in the strings stringtable which @veteran29 forgot

PR #1005, opened 2018-10-11
CBA_fnc_currentUnit report UAV entity when controlling UAV

### LINKED ISSUE #1004: Visionmode EH issue with drone, selecting night vision.
**Arma 3 Version:** `1.84` stable
**CBA Version:** `3.8.0` stable

**Mods:**
```
- CBA_A3
```

**Description:**
- Visionmode player event handler doesn't fire when switching to nightvision while controlling a drone. 

**Steps to reproduce:**
- Add Visionmode player event handler. Remote control a drone. Switch vision modes. It reports when changing from night vision to thermal, thermal to thermal, and thermal to normal vision, but night vision is never reported.
Example of the eventhandler code.
["visionMode", {systemChat str _this}, false] call CBA_fnc_addPlayerEventHandler;

- Editor (Singleplayer) / Editor (Multiplayer)

**RPT log file:**
- [pastebin](https://pastebin.com/watx2egd))

commy2:
- close #1004 

This is more complex than the original, and this is a function called every frame in ACE.

Thoughts?

### DIFF
--- a/addons/events/fnc_addPlayerEventHandler.sqf
+++ b/addons/events/fnc_addPlayerEventHandler.sqf
@@ -129,10 +129,12 @@ if (_id != -1) then {
         GVAR(oldCameraView) = "";
         GVAR(oldFeatureCamera) = "";
         GVAR(oldVisibleMap) = false;
+        GVAR(oldUAVControl) = [];
+        GVAR(controlledEntity) = objNull;
 
         GVAR(playerEHInfo) pushBack addMissionEventHandler ["EachFrame", {call FUNC(playerEH_EachFrame)}];
         [QFUNC(playerEH_EachFrame), {
-            private _player = call CBA_fnc_currentUnit;
+            private _player = missionNamespace getVariable ["bis_fnc_moduleRemoteControl_unit", player];
             if !(_player isEqualTo GVAR(oldUnit)) then {
                 [QGVAR(unitEvent), [_player, GVAR(oldUnit)]] call CBA_fnc_localEvent;
                 GVAR(oldUnit) = _player;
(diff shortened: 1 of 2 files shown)


PabstMirror:
This will have massive effects in ace because ace_player will be the uav robot
`typeOf ace_player = "B_UAV_AI"`
interaction menu, stamina, hearing... will all have problems with this 
(e.g. can self interact and check the robot's pulse)

PabstMirror:
Maybe use ace's uav eh https://github.com/acemod/ACE3/blob/master/addons/common/XEH_postInit.sqf#L285-L327

and then we could set `cba_controledUnit` on either player change or uav change events
there would be no polling overhead

commy2:
Hmm, one alternative would be to only use the UAV entity to read the visionMode, and keep currentUnit the way it is. Then one could additionally add an event whenever the controlledUnit changes.

commy2:
I'm a bit hesitant to use `PlayerViewChanged`, because that worked out so bad the last time.

dedmen:
Is there any practical reason why we need this? 
Ignoring that it breaks like... Most mods that use this?
TFAR/ACRE your radio will suddenly not work when you are controlling a UAV because the UAV doesn't have a Radio and your hearing/voice will suddenly be inside the UAV.
Don't know what ACE would do but I guess even worse.

commy2:
True. The only reason one would the need the UAV entity is to read the night vision mode from. I'll change the PR to do that instead.

commy2:
What about this?

addons/events/fnc_addPlayerEventHandler.sqf
```diff
@@ -129,13 +129,16 @@ if (_id != -1) then {
         GVAR(oldCameraView) = "";
         GVAR(oldFeatureCamera) = "";
         GVAR(oldVisibleMap) = false;
+        GVAR(oldUAVControl) = [];
+        GVAR(controlledEntity) = objNull;
 
         GVAR(playerEHInfo) pushBack addMissionEventHandler ["EachFrame", {call FUNC(playerEH_EachFrame)}];
         [QFUNC(playerEH_EachFrame), {
-            private _player = call CBA_fnc_currentUnit;
+            private _player = missionNamespace getVariable ["bis_fnc_moduleRemoteControl_unit", player];
```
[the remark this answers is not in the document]
commy2: I did it just to compensate for the loss due to this.

commy2:
Is it good?

addons/events/fnc_addPlayerEventHandler.sqf
```diff
@@ -189,7 +192,24 @@ if (_id != -1) then {
                 [QGVAR(turretEvent), [_player, _data]] call CBA_fnc_localEvent;
             };
 
-            _data = currentVisionMode _player;
+            // handle controlling UAV, UAV entity needed for visionMode
+            _data = UAVControl getConnectedUAV _player;
+            if !(_data isEqualTo GVAR(oldUAVControl)) then {
+                GVAR(oldUAVControl) = _data;
+
+                private _role = _data param [(_data find _player) + 1, ""];
```
dedmen: Is the wiki page wrong?
https://community.bistudio.com/wiki/UAVControl
_data should be 2 element array. find _player should always return 0.

commy2: >Is the wiki page wrong?

Always.

Drone with turret + driver/pilot looks like this:
[object1, "DRIVER", object2, "GUNNER"]
In MP, it can be controlled by multiple people. Therefore #0 is not necessarily the local player.

PabstMirror:
We could optimize for common case, which would be no UAV connected
```
_data = UAVControl getConnectedUAV player; 
if !(_data isEqualTo oldUAVControl) then {
0.0022 ms

if (!isNull getConnectedUAV player) then { // full code runs every time connected to UAV
0.0017 ms
```
https://github.com/CBATeam/CBA_A3/compare/currentUnit-UAV...uavPossibleOpti?expand=1
Not sure if worth it

PR #1008, opened 2018-10-16
Fix duplicate stringtable for JAM

Katalam:
Item STR_CBA_JR_Component listed twice

Just a rename of the stringtable item. No other changes needed.


PabstMirror:
Fixes 
```
Item STR_CBA_JR_Component listed twice
String STR_cba_jam_component not found
```

PR #1012, opened 2018-10-19
Invisible Targets

### LINKED ISSUE #1011: Feature request: Modified invisible targets
As I am not sure whether this is really within the scope of CBA nor can I really find a proper way to figure out who or how to contact, I'm opening this feature request.
**Description:**
- The vanilla game has these classes.
  - B_TargetSoldier
  - O_TargetSoldier
  - I_TargetSoldier
They have problems namely with being hidden and phyiscally blocking bullets. My suggestion would be to add modified classes which changes their models and doesn't hide them, so they're more versatile for mission and mod makers. I have made my own example, however classnames should be different and requiredAddons might also need to be different.
```
class CfgPatches
{
  class tint_targets
  {
    units[]={
      "B_TargetSoldier_tin",
      "O_TargetSoldier_tin",
      "I_TargetSoldier_tin"
    };
    weapons[]={};
    requiredVersion=1.0;
    requiredAddons[]=
    {
      "A3_Structures_F_Training"
    };
  };
};
class CfgVehicles 
{
  class B_TargetSoldier;
  class O_TargetSoldier;
  class I_TargetSoldier;
  class B_TargetSoldier_tin: B_TargetSoldier 
  {
    model = "\A3\Structures_F\System\ClutterCutter_small_F.p3d";
    scope = 2;
  };
  class O_TargetSoldier_tin: O_TargetSoldier 
  {
    model = "\A3\Structures_F\System\ClutterCutter_small_F.p3d";
    scope = 2;
  };
  class I_TargetSoldier_tin: I_TargetSoldier 
  {
    model = "\A3\Structures_F\System\ClutterCutter_small_F.p3d";
    scope = 2;
  };
};
```
I'm not sure what the stance would be add to add new classes. If this is appropriate for CBA I would be happy to try and rewrite it to conform to CBA standards.
And just for completeness sake, I have my whole rant that inspired me to discuss this. 
https://forums.bohemia.net/forums/topic/219853-the-invisible-targets-you-never-knew-existed-and-how-they-could-be-so-much-better/

Edited: Changed the requiredAddons to actually point to the addon containing the original definitions.

Tinter:
- Will add variants for `B_TargetSoldier`, `O_TargetSoldier` and `I_TargetSoldier` that have models that don't block projectiles, `scope = 2;` and fixing duplicate hitpoints error.
As discussed in  #1011 

If anything needs to be changed to comply with the CBA standards, just let me know and I'll be happy to change it. 

- close #1011


commy2:
Did a simple `class Turrets {};` not work?

commy2:
Team wants this in ACE instead, because out of scope for CBA.

Tinter:
Should we discuss it here or not?

commy2:
Sure.

Tinter:
Well without knowing the arguments, I'd say that this seems more like a tool than a feature or gameplay modification. I can't say the scope of CBA is particularly well-defined, as in "if your request has these three things, it fits CBA". I tried looking into it and all I could find mostly seemed to point to the scope including things that are useful for mod makers and mission makers, generally and not specifically. I would associate ACE more with gameplay overhaul and extra features that are meant to influence gameplay directly. 
I think this feature certainly fits the role of tool more than a gameplay feature. Someone installing ACE would see their game affected directly, while CBA doesn't do anything by itself, just like loading this feature as a standalone mod.

By putting this in ACE rather than CBA, anything using this feature would have a dependency on ACE rather than CBA, which doesn't make much sense since it doesn't have much to do with ACE in itself.
It'd probably also find more use in CBA since some people might not want to use ACE. Having a mission relying on ACE for the invisible targets would also force the medical system and different hud onto the mission.

I will concede that I haven't seen any features of CBA using new models, so I guess it fits under some interpretation of 'content'.

Is there anywhere I can see why the team would like it in ACE instead of CBA?

PR #1015, opened 2018-10-21
Changed/Updated isTurnedOut/isAttenuated function

### LINKED ISSUE #1014: variable _turn get never used in function isTurnedOut
**Arma 3 Version:** stable
**CBA Version:** stable

**Mods:**
```
- CBA_A3
```

**Description:**
https://github.com/CBATeam/CBA_A3/blob/fbd59d2ac1f85da79ea18a53b2d14e3ea285f60a/addons/common/fnc_isTurnedOut.sqf#L53

shukari:
This is a PR for radio addons like acre and tfar that there isolation is not on outside gunners and the vanilla isTurnedOut command cant handle that.

You are free to use the function or parts of it in your mod, have fun with it!


commy2:
This function is legacy due to isTurnedOut command, no?

shukari:
No, this functions has better results. For example the gunner of the B_LSV_01_armed_F (https://community.bistudio.com/wikidata/images/b/b5/B_LSV_01_armed_F.jpg) is with isTurnedOut FALSE... also cargo of a open cargo truck.

addons/common/fnc_isTurnedOut.sqf
```diff
@@ -27,54 +27,48 @@ Examples:
...
-            private _canHideGunner = getNumber(_turret >> "canHideGunner");
-            private _forceHideGunner = getNumber(_turret >> "forceHideGunner");
-            _turn = (_canHideGunner == 1 && _forceHideGunner == 0);
-
-            private _hatchAnimation = getText(_turret >> "animationSourceHatch");
-            _out = (_vehicle animationPhase _hatchAnimation) > 0;
+if (driver _vehicle == _unit) then {
+    _return = (_vehicle animationPhase "hatchDriver") > 0;
+} else {
+    private _assignedRole = assignedVehicleRole _unit;
```
commy2: This command reports `nil` while remote controlling a unit. Try this function while remote controlling an AI with zeus, and it should fail.

addons/common/fnc_isTurnedOut.sqf
```diff
@@ -27,54 +27,48 @@ Examples:
...
+        private _turretPath = _assignedRole select 1;
+        private _turret = [_vehicle, _turretPath] call CBA_fnc_getTurret;
 
-        } else {
-            if ((_assignedRole select 0) == "Cargo") then {
-                private _attenuateCargo = getArray(_cfg >> "soundAttenuationCargo");
-                if ((count _attenuateCargo) > 0) then {
-                    private _index = (count _attenuateCargo)-1; // wait for command to get cargo index
+        private _gunnerAction = getText (_turret >> "gunnerAction");
+        private _hatchAnimation = getText(_turret >> "animationSourceHatch");
```
commy2: space

addons/common/fnc_isTurnedOut.sqf
```diff
@@ -27,54 +27,48 @@ Examples:
...
-                        if(_index > (count _attenuateCargo)-1) then {
-                            _index = (count _attenuateCargo)-1;
-                        };
-                        if((_attenuateCargo select _index) == 0) then {
-                            _out = true;
-                        };
+                if (_index > -1) then {
+                    if (_index > (count _attenuateCargo) - 1) then {
+                        _index = (count _attenuateCargo) - 1;
+                    };
```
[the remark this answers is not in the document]
shukari: that is old code untouched by me but will change

shukari:
@commy2 because FFVs are ever outside is it possible to get the functions ace_common_fnc_getTurretIndex and ace_common_fnc_getTurretsFFV to cba?

dedmen:
This will return true, even if not actually turned out in "open" vehicles? Like cabrio's and stuff?
That'll be awesome.

shukari:
@dedmen yes... also gunners like the rhs humvee m2 gunner... also tfar can now handle open cargo seats with `soundAttenuationCargo` it is an array which define with cargoIndex is open to the outside... if its [1] no cargo can hear the outside... [0] means every... and [1,0] means cargo index 0 cant hear but all other.

TheMagnetar:
Top. One of the issues with intercom was detection of turned out units

addons/common/fnc_isTurnedOut.sqf
```diff
@@ -27,54 +27,46 @@ Examples:
...
+            private _cargoIndex = _vehicle getCargoIndex _unit;
+            private _window = ['ani_window_1', 'ani_window_2', 'ani_window_4', 'ani_window_3'] select (_cargoIndex + 1);
+            if ((_vehicle doorPhase _window) > 0) exitWith {_return = true};
+            
+            if ([_unit] call ace_common_fnc_getTurretIndex in ([_vehicle] call ace_common_fnc_getTurretsFFV)) exitWith {_return = true};
+            
+            private _attenuateCargo = getArray (_cfg >> "soundAttenuationCargo");
+            if !(_attenuateCargo isEqualTo []) then {
+                private _count = count _attenuateCargo;
+                _return = 0 == _attenuateCargo select (if (_cargoIndex >= _count) then {_count - 1} else {_cargoIndex});
```
commy2: This line... is art?

addons/common/fnc_isTurnedOut.sqf
```diff
@@ -27,54 +27,46 @@ Examples:
...
+        private _hatchAnimation = getText (_turret >> "animationSourceHatch");
+        
+        _return = (_vehicle animationPhase _hatchAnimation) > 0 || animationState player == _gunnerAction;
+    } else {
+        if (_position == "Cargo") then {
+            private _cargoIndex = _vehicle getCargoIndex _unit;
+            private _window = ['ani_window_1', 'ani_window_2', 'ani_window_4', 'ani_window_3'] select (_cargoIndex + 1);
+            if ((_vehicle doorPhase _window) > 0) exitWith {_return = true};
+            
+            if ([_unit] call ace_common_fnc_getTurretIndex in ([_vehicle] call ace_common_fnc_getTurretsFFV)) exitWith {_return = true};
```
commy2: These are unavailable when ACE is not used. Bad idea.

addons/common/fnc_isTurnedOut.sqf
```diff
@@ -27,54 +27,46 @@ Examples:
...
-
-        } else {
-            if ((_assignedRole select 0) == "Cargo") then {
-                private _attenuateCargo = getArray(_cfg >> "soundAttenuationCargo");
-                if ((count _attenuateCargo) > 0) then {
-                    private _index = (count _attenuateCargo)-1; // wait for command to get cargo index
+if (driver _vehicle == _unit) then {
+    _return = (_vehicle animationPhase "hatchDriver") > 0;
+} else {
+    (assignedVehicleRole _unit) params [["_position", ""], ["_turretPath", []]];
```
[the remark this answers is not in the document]
shukari: didnt know about them... will change

addons/common/fnc_isTurnedOut.sqf
shukari: Should i change it?

addons/common/fnc_isTurnedOut.sqf
shukari: I wrote it as attention in the first message and as @ mention comment to you https://github.com/CBATeam/CBA_A3/pull/1015#issuecomment-431695148

Can i add them to cba... or is there an other alternative?

addons/common/fnc_isTurnedOut.sqf
commy2: I've been looking at it for the last 6 minutes and I still have no idea what it does.

addons/common/fnc_isTurnedOut.sqf
commy2: Will thonk about to tomorrow morning.

addons/common/fnc_isTurnedOut.sqf
shukari: https://github.com/CBATeam/CBA_A3/pull/1015#issuecomment-431695545 the last part about `soundAttenuationCargo`

shukari:
I think i should make a own cba function or a function for tfar itself.
Because... isTurnedOut should be if someone with a hatch goes in or out... but i want all units that are not isolated to the outside aka are on the fresh air in vehicles.
I dont know what i should do... any ideas?

dedmen:
Was isTurnedOut deprecated so far? 
I personally don't have a problem to essentially change functionality. 
If you move it to a seperate function instead. I'd prefer the return value to be the isolation between 0-1 instead of just a bool.

I would vote for moving the ACE func into CBA. As it seems to be universally useful.
And the code itself (https://github.com/acemod/ACE3/blob/master/addons/common/functions/fnc_getTurretIndex.sqf) doesn't contain anything ACE specific.

dedmen:
Was isTurnedOut deprecated so far? 
I personally don't have a problem to essentially change functionality. 
If you move it to a seperate function instead. I'd prefer the return value to be the isolation (from soundAttenuation config) between 0-1 instead of just a bool.

I would vote for moving the ACE func into CBA. As it seems to be universally useful.
And the code itself (https://github.com/acemod/ACE3/blob/master/addons/common/functions/fnc_getTurretIndex.sqf) doesn't contain anything ACE specific.

dedmen:
Was isTurnedOut deprecated so far? 
I personally don't have a problem to essentially change functionality. 
If you move it to a seperate function instead. I'd prefer the return value to be the isolation (from soundAttenuation config) between 0-1 instead of just a bool.

I would vote for moving the ACE func into CBA. As it seems to be universally useful.
And the code itself (https://github.com/acemod/ACE3/blob/master/addons/common/functions/fnc_getTurretIndex.sqf) doesn't contain anything ACE specific.

shukari:
@dedmen `soundAttenuation` is only 1 if the index cant hear or 0 if he can... i saw no value between.

addons/common/fnc_isTurnedOut.sqf
```diff
@@ -27,54 +27,53 @@ Examples:
...
-                            _out = true;
-                        };
-                    };
-                };
-            };
-        };
+private _role = _unit call CBA_fnc_vehicleRole;
+
+// FFVs are allways outside
+if ([_unit] call ace_common_fnc_getTurretIndex in ([_vehicle] call ace_common_fnc_getTurretsFFV)) then {_return = true};
```
Dystopian: ACE functions in CBA?

addons/common/fnc_isTurnedOut.sqf
```diff
@@ -27,54 +27,53 @@ Examples:
...
+
+// Spezial condition for RHS car windows
+if (_vehicle isKindOf "rhsusf_m1025_w" && {_role in ["driver", "Turret"]}) then {
```
Dystopian: hack for 3rd party addons?

addons/common/fnc_isTurnedOut.sqf
shukari: AHAHAHAHAHAHAHAHAHAHAHAHAHA 😱 
There is the attention -> https://github.com/CBATeam/CBA_A3/pull/1015#issue-224558602
There is the first comment about the question -> https://github.com/CBATeam/CBA_A3/pull/1015#issuecomment-431695148
There is dedman talking about that -> https://github.com/CBATeam/CBA_A3/pull/1015#issuecomment-431797475

addons/common/fnc_isTurnedOut.sqf
shukari: Yes, extra with a comment... is nothing for CBA but for other mods i think... i would like to make it possible to add extra stuff for addons to make it more variable...
This file is atm heavy at WIP and i try all to make it working with all i can test with my available mods.

addons/common/fnc_isTurnedOut.sqf
Dystopian: Either github still have issues or I just overlooked those comments. And attention was overlooked for sure :)

addons/common/fnc_isTurnedOut.sqf
Dystopian: Maybe replace it with some config values check and then ask RHS to add them to their source.

shukari: Config is not possible for that... this are the windows that you can open via a useraction.
So its highly variable... i dont now how... i suppose a eventhandler thing or something is needed here.
It would be nice if this function can be changed by outside so mission/mod makers can make there own things and this function only make the static things.

addons/common/fnc_isTurnedOut.sqf
shukari: you are right... github is partially down atm https://status.github.com/messages

addons/common/fnc_isTurnedOut.sqf
```diff
@@ -27,54 +27,53 @@ Examples:
...
-                        if((_attenuateCargo select _index) == 0) then {
-                            _out = true;
-                        };
-                    };
-                };
-            };
-        };
+private _role = _unit call CBA_fnc_vehicleRole;
+
+// FFVs are allways outside
```
dedmen: always*

addons/common/fnc_isTurnedOut.sqf
dedmen: >There is dedman talking about that

That's not my name

addons/common/fnc_isTurnedOut.sqf
```diff
@@ -27,54 +27,53 @@ Examples:
...
+if ([_unit] call ace_common_fnc_getTurretIndex in ([_vehicle] call ace_common_fnc_getTurretsFFV)) then {_return = true};
+
+// Spezial condition for RHS car windows
```
dedmen: Special*

addons/common/fnc_isTurnedOut.sqf
```diff
@@ -27,54 +27,53 @@ Examples:
...
+if (_vehicle isKindOf "rhsusf_m1025_w" && {_role in ["driver", "Turret"]}) then {
+    private _cargoIndex = (_vehicle getCargoIndex _unit) min 2;
+    private _window = ['ani_window_1', 'ani_window_2', 'ani_window_4', 'ani_window_3'] select (_cargoIndex + 1);
+    _return = (_vehicle doorPhase _window) > 0;
+};
+if (_return) exitWith {_return}; // return
```
dedmen: kinda useless comment

addons/common/fnc_isTurnedOut.sqf
```diff
@@ -27,54 +27,53 @@ Examples:
...
+        _return = (_attenuateCargo select ((_vehicle getCargoIndex _unit) min (count _attenuateCargo - 1))) == 0;
     };
-} else {
-    _out = true;
+    
+    _return;
 };
-_out;
 
+_return;
```
dedmen: leave away `;` at return values

addons/common/fnc_isTurnedOut.sqf
```diff
@@ -27,54 +27,53 @@ Examples:
...
+    private _turret = [_vehicle, _unit call CBA_fnc_turretPath] call CBA_fnc_getTurret;
+
+    private _gunnerAction = getText (_turret >> "gunnerAction");
+    private _hatchAnimation = getText (_turret >> "animationSourceHatch");
+    
+    (_vehicle animationPhase _hatchAnimation) > 0 || {animationState player == _gunnerAction}; // return
+};
+
+if (_role == "cargo") exitWith {
+    private _attenuateCargo = getArray (configFile >> "CfgVehicles" >> (typeOf _vehicle) >> "soundAttenuationCargo");
```
dedmen: You have `configFile >> "CfgVehicles" >> (typeOf _vehicle)` all over the place.
Resolve it once before it'sf irst use. Then store it in a variable and reuse that.

addons/common/fnc_isTurnedOut.sqf
jonpas: Neither is dedmen, it's just your nickname. ;)

commy2:
I will tidy up this a bit...

addons/common/fnc_isTurnedOut.sqf
Dystopian: just checked and it looks like all `Turret` roles in these RHS Humvees are FFV so and no sense to check windows because FFV are checked above

commy2:
@shukari 
Please merge: https://github.com/shukari/CBA_A3/pull/1/files
The behavior should be unchanged, but easier to follow.

addons/common/fnc_isTurnedOut.sqf
shukari: because if the window is closed it returns false

addons/common/fnc_isTurnedOut.sqf
shukari: no, else the return true will be over riden by the exitwith conditions

shukari:
@commy2 ok... but i have the right to edit this branch... edit is enabled for contributors

commy2:
Just click this link: https://github.com/shukari/CBA_A3/pull/1
And then merge. You should have access to your own fork, otherwise Github is broken.

addons/common/fnc_isTurnedOut.sqf
shukari: driver and cargo need it... but turret, gunner and commander not... so init it every time?

dedmen: why not move gunner/turret above driver/cargo then?

shukari: like it

shukari: but the get never executed booth only one of them... so reuse is not there

shukari:
my test vehicles and before/after analysis is in first post

addons/common/fnc_isTurnedOut.sqf
```diff
@@ -6,75 +6,81 @@ Description:
...
 };
-_out;
 
+// FFVs are always outside
+if (_isFFV) exitWith {true};
+
+if (_role isEqualTo "commander") exitWith {
+    // nothing atm but so _cfgAttCargo will not executed
+    false // return
+};
```
commy2: This seems pointless. Should be deleted, because the last `false` will handle it.

shukari: the soundAtt... cfg will not executed... but yeah it seams pointless ^^

commy2: Instead, you do another check, which is in total more expensive, and makes everything harder to read.

shukari:
There is a `attenuationEffectType` (`getText (configfile >> "CfgVehicles" >> typeOf cursorObject >> "attenuationEffectType")`) 😄 

Types from `configfile >> "CfgSoundEffects" >> "AttenuationsEffects"`:
["","OpenCarAttenuation","PlaneAttenuation","TankAttenuation","CarAttenuation","TruckAttenuation","SemiOpenCarAttenuation2","MrapAttenuation","OpenHeliAttenuation","HeliAttenuation","SemiOpenHeliAttenuation","SemiOpenCarAttenuation","RHS_CarAttenuation"]

addons/common/fnc_isTurnedOut.sqf
```diff
@@ -29,6 +29,10 @@ params ["_unit"];
 if (isNull objectParent _unit) exitWith {true};
 if (isTurnedOut _unit) exitWith {true};
 
+private _cfg = configFile >> "CfgVehicles" >> typeOf _vehicle;
```
commy2: _config

addons/common/fnc_isTurnedOut.sqf
```diff
@@ -35,21 +34,20 @@ private _role = toLower (_unit call CBA_fnc_vehicleRole);
 private _isFFV = fullCrew [_vehicle, "turret"] findIf {_x select 0 == _unit && {_x select 4}} != -1;
 
 // special condition for RHS car windows
-private _windowsOpen = false;
 if (_vehicle isKindOf "rhsusf_m1025_w" && {_role in ["driver", "turret"]}) then {
-    // will override _isFFV if window is closed
+    //will override _isFFV if window is closed
     _isFFV = switch (_vehicle getCargoIndex _unit) do {
```
dedmen:
```
_vehicle doorPhase
(["ani_window_1","ani_window_2", "ani_window_4"] param [(_vehicle getCargoIndex _unit) + 1, "ani_window_3"])
> 0
```
magic

shukari: i did the same
```
private _cargoIndex = (_vehicle getCargoIndex _unit) min 2;
private _window = ['ani_window_1', 'ani_window_2', 'ani_window_4', 'ani_window_3'] select (_cargoIndex + 1);
```
@commy2 didnt liked it 😄

addons/common/fnc_isTurnedOut.sqf
```diff
@@ -6,75 +6,80 @@ Description:
...
-SCRIPT(isTurnedOut);
-
 params ["_unit"];
 
-private _turn = false;
-private _out = false;
+if (isNull objectParent _unit) exitWith {true};
+if (isTurnedOut _unit) exitWith {true};
+
+private _config = configFile >> "CfgVehicles" >> typeOf _vehicle;
```
Dystopian: `_vehicle` is used before defined. You should test it.

addons/common/fnc_isTurnedOut.sqf
```diff
@@ -6,75 +6,80 @@ Description:
...
-                        if(_index > (count _attenuateCargo)-1) then {
-                            _index = (count _attenuateCargo)-1;
-                        };
-                        if((_attenuateCargo select _index) == 0) then {
-                            _out = true;
-                        };
-                    };
-                };
-            };
+private _role = toLower (_unit call CBA_fnc_vehicleRole);
```
Dystopian: `_role` here, cargo index, `_isFFV` and turret path below can all be found from one `fullCrew` call. Something like this:
```
private _fullCrew = fullCrew [_vehicle, ""];
(
    _fullCrew select (_fullCrew findIf {_unit == _x select 0})
) params ["", "_role", "_cargoIndex", "_turretPath", "_isFFV"];
```
I'm not sure if `_cargoIndex` is equal to `getCargoIndex`, worth checking.
Also role should be `toLower`ed then.

addons/common/fnc_isTurnedOut.sqf
```diff
@@ -6,75 +6,80 @@ Description:
...
+private _isFFV = fullCrew [_vehicle, "turret"] findIf {_x select 0 == _unit && {_x select 4}} != -1;
+
+// special condition for RHS car windows
+if (_vehicle isKindOf "rhsusf_m1025_w" && {_role in ["driver", "turret"]}) then {
```
Dystopian: Once again I ask not to include 3rd party addon hacks to CBA. CBA is a framework. 3rd party addons should add hacks for CBA and not vice versa.

This case can be handled another way. RHS `Open/Close window` action script sets vehicle variable like `format ["CBA_isTurnedOut_%1_%2", _role, _turretPath]` to true/false. CBA function checks these variables. Role and turret path are enough to identify vehicle seat in all cases.

Yes this way will require some activity from RHS. I think if create ticket with their patched script they won't refuse to implement it. But this is universal enough way for any mod not RHS only.

TheMagnetar:
I think we are changing definitions here. One thing is to check if a unit is turned out (hatch etc) and the other is to also return true when a unit is open like the pilot/copilot of a little bird. Those positions, are not "turned out", nor is the driver of a quad. Changing this may break some stuff, so I propose to keep the "turned out" function to simply check that, and an additional function to check if the unit is in the open (quad driver, littlebird pilot)

addons/common/fnc_isTurnedOut.sqf
Dystopian: Also at least CUP armed SUV has scripted turn in/out solution. Such implementation could take it into account.

jonpas:
Fully agreed with Magnetar here.

Additional function will also serve me, throwing stuff out of the little bird's copilot seat anyone? ;)

addons/common/fnc_isTurnedOut.sqf
shukari: Forget to sync it with my testfile where i fixed it.
Thanks for spotting this.

shukari:
I am at this moment a little bit pissed off.
I ask questions get no answers and then days after some other come with the same question.

"I think i should make a own cba function or a function for tfar itself"
"isTurnedOut should be if someone with a hatch goes in or out"
 https://github.com/CBATeam/CBA_A3/pull/1015#issuecomment-431707026

_And once again I ask_
"i suppose a eventhandler thing or something is needed here.
It would be nice if this function can be changed by outside so mission/mod makers can make there own things and this function only make the static thing"
https://github.com/CBATeam/CBA_A3/pull/1015#discussion_r227014491

The rhs stuff is there that anyone can see what is needed for an api
https://github.com/CBATeam/CBA_A3/pull/1015#discussion_r226978447
Im new to cba dev stuff so i dont know the best way to achieve that.


# 1. Question: Should this be a own function like "CBA_fnc_isExposed"?
# 2. Question: How can 3rd party mods interact with this function? Like here mentioned https://github.com/CBATeam/CBA_A3/pull/1015#discussion_r227663777?

TheMagnetar:
But you do have all the answers!

- Some of us have already expressed that a separate would be better
- Interaction with 3rd party mods. @Dystopian suggested something already a variable with `format ["CBA_isTurnedOut_%1_%2", _role, _turretPath]`or similar.

shukari:
_Some of us have already expressed that a separate would be better_
you and @jonpas  said that... @commy2 and @dedmen said its ok... so what is now the way?

_suggested something already_
Yeah that is one way, but is there a better one? I didnt like the idea to have multiple public variables on a vehicle. So if no better idea comes up i will implement this, no problem.

commy2:
`CBA_fnc_isAttenuated`

commy2:
Just make it report `true` when attenuated, and if someone wants the opposite, they can invert it using `not`.

commy2:
@Dystopian 
>I'm not sure if _cargoIndex is equal to getCargoIndex, worth checking.

This is Arma. Of course they're *not* the same. One of them increments with FFV turrets, the other doesn't. They are desynched for vehicles with both cargo and FFV slots. I just can't remember which is which.

shukari:
@commy2 Thanks you for testing... i tested it fast with two vehicles and it worked there.... so good to know.

Dystopian:
> Of course they're _not_ the same. One of them increments with FFV turrets, the other doesn't.

@commy2 Just checked `I_Heli_light_03_dynamicLoadout_F`:
```
[[<NULL-object>,"driver",-1,[],false
<NULL-object>,"cargo",2,[],false
<NULL-object>,"cargo",3,[],false
<NULL-object>,"cargo",4,[],false
<NULL-object>,"cargo",5,[],false
<NULL-object>,"Turret",-1,[0],false
<NULL-object>,"Turret",0,[1],true
<NULL-object>,"Turret",1,[2],true]]
```
and `C_Van_01_transport_F`:
```
[[gD,"driver",-1,[],false
<NULL-object>,"cargo",0,[],false
<NULL-object>,"cargo",1,[],false
<NULL-object>,"Turret",2,[0],true
<NULL-object>,"Turret",4,[1],true
<NULL-object>,"Turret",6,[2],true
<NULL-object>,"Turret",8,[3],true
<NULL-object>,"Turret",10,[4],true
<NULL-object>,"Turret",3,[5],true
<NULL-object>,"Turret",5,[6],true
<NULL-object>,"Turret",7,[7],true
<NULL-object>,"Turret",9,[8],true
<NULL-object>,"Turret",11,[9],true]]
```
For both vehicles `getCargoIndex` equals `_cargoIndex`. Maybe you mix this up with `GetInCargo`action cargo number param, it was discussed in acemod/ACE3#5733.

Dystopian:
> Yeah that is one way, but is there a better one? I didnt like the idea to have multiple public variables on a vehicle. So if no better idea comes up i will implement this, no problem.

@shukari You can try to handle it without variables but with animation state checking.
1. There is `enabledByAnimationSource` config value for some FFV turrets (including that RHS Hammer and even some vanilla vehicles too), maybe it can be taken into account.
1. For driver (in RHS Hammer) there can be config value like `CBA_isTurnedOut_anim = "ani_window_1"`. Those turrets which can't be handled with `enabledByAnimationSource` can also have such config value.
1. I don't think there is sense to handle cargo seats but if yes create e.g. `CBA_isTurnedOut_cargoAnims[] = {"cargo_anim1","cargo_anim2"}`.
1. I don't know which command is better: `doorPhase` `animationPhase` `animationSourcePhase`.
1. For inverted animations additional check is required. It can be handled e.g. with another config value like `CBA_isTurnedOut_anim_state = 1` (easier for CBA to handle but harder to implement 3rd party mod) or anim value can be modified some way e.g. `CBA_isTurnedOut_anim = "!ani_window_1"` (easier for 3rd party but a little harder for CBA).
1. Vanilla vehicles can be PITA for this function because BI won't add these values and CBA compat for vanilla is quite ridiculous.

So in function you just check if anim is configured for this seat and check this anim state. Something like I implemented in acemod/ACE3#6144.

This is just concept, it should be planned carefully with testing on many vehicles (vanilla, 3rd party).

Also want to say about helicopters with ramp like `I_Heli_Transport_02_F` - if this function should take ramp state into account. From what I can see now it isn't accounted.

dedmen:
> `CBA_fnc_isAttenuated`

As I already recommended. It would be awesome if you could return a 0-1 attenuation value. Based on the cargoAttenuation sound config entry thingy

shukari:
@dedmen https://github.com/CBATeam/CBA_A3/pull/1015#issuecomment-431806340

TheMagnetar:
> As I already recommended. It would be awesome if you could return a 0-1 attenuation value. Based on the cargoAttenuation sound config entry thingy

I wouldn't. It would break ACRE2 sound attenuation system based on compartments. True/False should be sufficient.

shukari:
Changes are not tested and not finished

addons/common/fnc_isAttenuated.sqf
```diff
@@ -0,0 +1,78 @@
...
+
+Parameters:
+    _unit - Unit to check <OBJECT>
+
+Returns:
+    "true" for inside unit that cant hear the outside <BOOL>
+
+Examples:
+    (begin example)
+        if ([player] call CBA_fnc_isAttenuated) then {
```
dedmen: Conflicts with L12.
Line 12 says returns STRING "true" (which is wrong datatype btw) for when unit cant hear the outside.
This example says it returns true when the unit can hear the ourside. What now?

addons/common/fnc_isAttenuated.sqf
```diff
@@ -0,0 +1,78 @@
...
+private _vehicle = vehicle _unit;
+private _config = configFile >> "CfgVehicles" >> typeOf _vehicle;
+private _attenuationType = getText (_config >> "attenuationEffectType");
+if (_attenuationType in ["OpenCarAttenuation", "OpenHeliAttenuation"]) exitWith {false};
+
+private _fullCrew = fullCrew _vehicle;
+(_fullCrew select (_fullCrew findIf {_unit == _x select 0})) params ["", "_role", "_cargoIndex", "_turretPath", "_isFFV"];
+_role = toLower _role;
+
+// API try
```
dedmen: What does this mean? API try?

addons/common/fnc_isAttenuated.sqf
shukari: my fail... need someone for the header infos... my english is trash

addons/common/fnc_isTurnedOut.sqf
```diff
@@ -30,44 +27,22 @@ if (isNull objectParent _unit) exitWith {true};
...
-    // if attenuate array has less elements than the current cargo index, use the last value of the array
-    private _cargoIndex = _cargoIndex min (count _attenuateCargo - 1);
-
-    _attenuateCargo param [_cargoIndex, 1] == 0 // return
+if (_role isEqualTo "commander") exitWith {
+    _vehicle animationPhase _hatchAnimation > 0 // return
 };
 
-false
+false
```
TheMagnetar: New line missing at the end of file.

commy2:
Status?

shukari:
much todo in rl... will try to make progress this week

shukari:
Progress was made. Need feedback. On all my test vehicles it work like it should.
Is the API way ok?

addons/common/fnc_isAttenuated.sqf
```diff
@@ -0,0 +1,96 @@
+/* ----------------------------------------------------------------------------
+Function: CBA_fnc_isAttenuated
+
+Description:
+    Checks whether a unit is attenuated.
```
jonpas: Give explanation of what that means, eg:
`Checks whether a unit is attenuated, meaning in another compartment outside of outside voice range.`
Someone give a better one, I don't know how to describe that.

addons/common/fnc_isAttenuated.sqf
```diff
@@ -0,0 +1,96 @@
...
+
+Parameters:
+    _unit - Unit to check <OBJECT>
+
+Returns:
+    true - for inside unit that cant hear the outside <BOOL>
```
jonpas: `true if attenuated, false otherwise`

addons/common/fnc_isAttenuated.sqf
jonpas: I'd just do:
`_isAttenuated = [player] call CBA_fnc_isAttenuated;`
Seems more in form with other headers.

addons/common/fnc_isAttenuated.sqf
```diff
@@ -0,0 +1,96 @@
...
+
+Examples:
+    (begin example)
+        if ([player] call CBA_fnc_isAttenuated) then {
+            player sideChat "I cant hear you outside guy!";
+        };
+    (end)
+
+Config & Variable API:
```
jonpas: Will this work with auto-generated API pages?

addons/common/fnc_isAttenuated.sqf
```diff
@@ -0,0 +1,96 @@
...
+    Example:
+        Config: CBA_attenuatedRoles[] = {{"driver", -1, "ani_window_1"}, {"turret", {0}, "ani_window_2"}, {"gunner", -1, "ani_window_4"}};
+        Variable:  _vehicle setVariable ["CBA_attenuatedRoles", ["driver", -1, "ani_window_1"]];
+    Attributes
+        role
+        cargoindex or turretPath
+        animation for attenuated
+    Info
+        variable is override config
```
jonpas: `variable overrides config`

addons/common/fnc_isTurnedOut.sqf
```diff
@@ -1,80 +1,48 @@
...
 
 Returns:
-    "true" for turned out or "false" for not turned out [Boolean]
-    Cargo in exposed vehicles are "turned out" if they lack a
-    cargoSoundAttenuation matching index. Do NOT assume that because
-    this function returns false that the unit is generally exposed.
-
-    For example, ATVs will return false, even though they are "turned
-    out".
+    true - for turned out <BOOL>
```
jonpas: `true if turned out, false otherwise`

addons/common/fnc_isTurnedOut.sqf
```diff
@@ -1,80 +1,48 @@
...
+    Turned out means out of a hatch or something similar.
```
jonpas: That should go into Description.

addons/common/fnc_isTurnedOut.sqf
```diff
@@ -1,80 +1,48 @@
...
 
 Examples:
     (begin example)
-    if ( [player] call CBA_fnc_isTurnedOut ) then
-    {
-        player sideChat "I am turned out!";
-    };
+        if ([player] call CBA_fnc_isTurnedOut) then {
```
jonpas:
```suggestion
        _isTurnedOut = [player] call CBA_fnc_isTurnedOut;
```

addons/common/fnc_isAttenuated.sqf
shukari: no i dont know auto-gen API pages do you have an example where it is used?

jonpas: Sorry, was a question to @commy2 and @Killswitch00! :)

shukari:
I need feedback and somebody need to test this.

commy2:
Status?

TheMagnetar:
Testing today

TheMagnetar:
I have tested this and I have come into an issue where it is inconsistent with isTurnedOut command, therefore this PR needs some work but it is undoubtedly going to be really useful once some conditions are fixed.

If the unit enters for  example a gunner position in a helicopter, the suggested turnedOut CBA function returns true, while the command `isTurnedOut` returns false (as it actually should). In that case, the isAttenuated function returns false, while it should return true.

Tested helicopter is `B_Heli_Transport_03_F`.

Additionally I have a suggestion for isAttenuated. Currently it only checks for a specific unit if it is attenuated, breaking for example when 2 units are inside the same vehicle. In case of the previous helicopter, when sitting on the rear left seat near the ramp and the ramp is closed, isAttenuated returns true, but if you open the ramp returns false (as it should) for a unit standing outside the vehicle. Now, if another player unit is inside the rear compartment of the chinook, will "see" the player sitting on the ramp seat to be attenuated (when it should not since they are in the same compartment).  The opposite happens if a unit  is in a tank, a turned out unit is not attenuated for units outside the vehicle, but it is for units inside the vehicle.

While the second case may be solved with the combination of checking if the unit is turned out/attenuated functions, the first case (chinook) does not since the unit is never "turned out"

shukari:
> If the unit enters for example a gunner position in a helicopter, the suggested turnedOut CBA function returns true, while the command isTurnedOut returns false (as it actually should). In that case, the isAttenuated function returns false, while it should return true.

Thats wrong. The CBA_fnc_isTurnedOut is not the same like the command isTurnedOut. It will return true if outside of a vehicle, vanilla not and also will additionally report turnout gunners and open hatches. It is the better command isTurnedOut and not the same. It behaves like it should.

> Additionally I have a suggestion for isAttenuated. Currently it only checks for a specific unit if it is attenuated, breaking for example when 2 units are inside the same vehicle. In case of the previous helicopter, when sitting on the rear left seat near the ramp and the ramp is closed, isAttenuated returns true, but if you open the ramp returns false (as it should) for a unit standing outside the vehicle. Now, if another player unit is inside the rear compartment of the chinook, will "see" the player sitting on the ramp seat to be attenuated (when it should not since they are in the same compartment). The opposite happens if a unit is in a tank, a turned out unit is not attenuated for units outside the vehicle, but it is for units inside the vehicle.

The the CBA_fnc_isAttenuated function is and should be a only one player to the outside function.
To be honestly i dont fully understand the point of this. Is it only that a unit to unit check function would be cool? If this is your point please at another PR im done with this.
But for me all works like it should with the `B_Heli_Transport_03_F`

jonpas:
Since goal of this PR includes:
> This is a change for radio addons like acre and tfar that there isolation is not on outside gunners and the vanilla isTurnedOut command cant handle that.

it makes sense in my eyes that it is done properly, including Cargo and FFV seats. If it does not work for all cases, then neither mod will use those new functions.

shukari:
@jonpas sry but you dont unterstand this PR... look at function isAttenuated

jonpas:
Is my understanding of the following correct?
- `isTurnedOut` is `true` when part of unit's body is physically outside of the vehicle internals (benches of little bird, body out of the hatch etc.), `false` otherwise
- `isAttenuated` is `true` when unit is attenuated by the vehicle (is inside a sound-isolated compartment such as in Humvee or BDR, has headphones on etc.), `false` otherwise (technical gunner, boat, open vehicle etc.)

I have worded it badly in my previous comment, but in case of Chinook gunner this would result in:
- `isTurnedOut` -> `true` as he is not physically outside of the vehicle
- `isAttenuated` -> `true` because he is attenuated by the helicopter (and headphones)

@TheMagnetar is that your understanding also?
@shukari how does your understanding differ? Please, do not take this as a bad thing or any kind of flaming on you, we simply have to be very clear on what is what and make sure we understand each other. :)

shukari:
My understanding:
- `isTurnedOut` a function that is for me not important and does only trouble because of the the understanding. For me it is a addition to isTurnOut command that report hatches and gunner that look out.
- `isAttenuated` is the function that this PR is all about. It should help or give the radio addons an idea of a solution with roles that can speak with the outside

Im a little bit frustrating and done with this PR.
If after this comment is a problem with `isTurnedOut` i will revert it to the old and integrate the changes to isAttenuated.

jonpas:
_FYI: I have corrected my above example, got confused and inverted a thing!_

dedmen:
@jonpas without the "headphones on" part.
Yes. Return true when inside a closed compartment, and can't (really) hear people outside. I wouldn't count things like headphones.. Because you can just take them off, or maybe you don't have them on right now, no way to detect that, that's something that the radio mods do on their own.

Door gunner on Chinook could return both really.. But should be false (can hear people from outside)
He should be able to hear people that are infront of his window, but not people that are outside of the vehicle behind him or infront of the chinook.. But sound also travels around corners.
You can't really check if someones sound would travel through the window or not, so a true (can't hear people outside) would also be a sensible choice.

Ref to @magnetar
>In that case, the isAttenuated function returns false, while it should return true.
I disagree, he has a open window infront of him, he can hear sounds from outside.

addons/common/fnc_isAttenuated.sqf
```diff
@@ -41,11 +41,13 @@ if (_turnedOut) exitWith {false};
 // open vehicle
 private _vehicle = vehicle _unit;
 private _config = configFile >> "CfgVehicles" >> typeOf _vehicle;
+
+// class CfgSoundEffects >> class AttenuationsEffects
 private _attenuationType = getText (_config >> "attenuationEffectType");
-if (_attenuationType in ["OpenCarAttenuation", "OpenHeliAttenuation"]) exitWith {false};
+if (_attenuationType in ["OpenCarAttenuation", "OpenHeliAttenuation", "jsrs_OpenCar_Attenuation", "jsrs_SemiOpenCar_Attenuation"]) exitWith {false};
```
TheMagnetar: Can we avoid mod specific "compatibilities" in CBA?

commy2: If you check out those classes, you can see that they don't have anything to do with whether you should or shouldn't hear anyone outside the vehicle anyway.
Tbh, it seems to me like this is something the radio mods should worry about. There is no link  from the base game to this. The ingame sounds clearly follow different rules that wouldn't work for the radio mods.

shukari: @commy2 This line is THE key line of this function. Because of this value i get quads, boots, littlebirds and open jeeps as open vehicles. It is not important what these classes define in cfgSoundEffects... it is important that these classes give a hint that the vehicle is open to the outside. If i delete this check... the function is useless.

 @TheMagnetar yes you are right, BUT then the function is broken if someone use the clientside mod JSRS... what should i do?

shukari:
@dedmen headphones, distance and other blocking things are so special edge cases that they should done by the mod that needs them in my opinion.

addons/common/fnc_isAttenuated.sqf
commy2: >This line is THE key line of this function.

But it isn't. It is an arbitrary list that would have to be expanded every time someone makes a new class that would qualify in a random mod. I am not going to approve and merge references to i.e. JSRS in CBA.

commy2: >Ideally have a "registerAttenuationClass" function which adds those to a global array. Similar to how "registerFeatureCamera" is but without code stuff.

No, because vehicle attenuation as it appears in Arma is not linked to whether you should be able to hear someone outside the vehicle from inside or not.

shukari:
My time with this PR is now over... i was so happy to make something for CBA but this is the end. Im done. Have fun commenting.
Anybody can use the function or parts of it in there mods.

dedmen:
Just integrated into TFAR. Works beautifully. both `isTurnedOut` and `CBA_fnc_isTurnedOut` return false when turned out as the gunner in vehicles, even Vanilla vehicles. This works properly.
Performance is shit. But replacing 
`private _turnedOut = [_unit] call CBA_fnc_isTurnedOut;` 
with 
`private _turnedOut = isTurnedOut _unit;`
Works fine and improves runtime from 0.118ms to 0.015ms.
Don't see why calling the CBA func would be needed here, it checks for sound attenuation, but this function already does that anyway.

dedmen:
This function has duplicates for stuff in CBA_fnc_isTurnedOut. This already has _role and _fullcrew. The isTurnedOut code should just be properly merged into this, rather than stuff being done twice. Especially because CBA_fnc_isTurnedOut is the major slowness factor here.

dedmen:
I have a workaround for the "hardocded jsrs" issue. Can just read the attenuationEffectType's entries in config
`CfgSoundEffects >> AttenuationsEffects`
and just collect and average out the `gain` entries.
That would actually be THE correct way to solve this. soundAttenuationCargo is just a bool flag for "yes attenuation effect applies" or "ignore attenuation effect and pretend he's outside"

dedmen:
>I like your function way more than this PR, although it also should be called isAttenuated and not isTurnedOut.

It's a copy of the PR, just turned around to return isNotAttenuated.

commy2:
Refering to this:
>This fixes issues where isAttenuated would return true even though the AttenuationEffect's gain is 1, meaning not attenuated.
>And additionally I get correct isolation values for everything that IS attenuated.
>Also checking that as i said removes your hardcoded jsrs check. And IMO removes the need for CBA_attenuatedRoles because people can just configure their vehicle correctly instead.

PR #1016, opened 2018-10-22
ignore locked statics in taskDefend

### LINKED ISSUE #1013: "CBA_fnc_taskDefend" does not ignore locked statics.
**Arma 3 Version:** `1.84` (stable)
**CBA Version:** `3.9.0` (stable)

**Mods:**
```
- CBA_A3
- ACE3
- CUP Weapons/Vehicles/Units
- ACRE2
```

**Description:**
AI groups given the taskDefend function will mount statics that are fully locked (not just locked for players). 

**Steps to reproduce:**
- Place AI group.
- Give AI group CBA_fnc_taskDefend.
- Place static near them.
- Lock the static.

- Dedicated

Cuel:
- AI will not consider locked statics in taskDefend

fix #1013

### DIFF
--- a/addons/ai/fnc_taskDefend.sqf
+++ b/addons/ai/fnc_taskDefend.sqf
@@ -58,7 +58,7 @@ private _statics = _position nearObjects ["StaticWeapon", _radius];
 private _buildings = _position nearObjects ["Building", _radius];
 
 // Filter out occupied statics
-_statics = _statics select {(_x emptyPositions "Gunner") > 0};
+_statics = _statics select {locked _x != 2 && {(_x emptyPositions "Gunner") > 0}};
 
 // Filter out buildings below the size threshold (and store positions for later use)
 _buildings = _buildings select {
--- a/addons/ai/fnc_waypointGarrison.sqf
+++ b/addons/ai/fnc_waypointGarrison.sqf
@@ -30,7 +30,7 @@ params ["_group", "_position"];
 // leader should not issue attack orders
 _group enableAttack false;
 
-private _staticWeapons = _position nearObjects ["StaticWeapon", 50] select {_x emptyPositions "gunner" > 0};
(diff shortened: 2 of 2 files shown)


addons/ai/fnc_taskDefend.sqf
```diff
@@ -58,7 +58,7 @@ private _statics = _position nearObjects ["StaticWeapon", _radius];
 private _buildings = _position nearObjects ["Building", _radius];
 
 // Filter out occupied statics
-_statics = _statics select {(_x emptyPositions "Gunner") > 0};
+_statics = _statics select {locked _x != 2 && {(_x emptyPositions "Gunner") > 0}};
```
commy2: 0 - Unlocked; 1 - Default; 2 - Locked; 3 - Locked for player; -1 - vehicle is null.

Cuel: You want to also check for -1?

commy2:
Same for https://github.com/CBATeam/CBA_A3/blob/master/addons/ai/fnc_waypointGarrison.sqf and https://github.com/CBATeam/CBA_A3/blob/master/addons/modules/fnc_moduleDefend.sqf
?

addons/ai/fnc_taskDefend.sqf
commy2: No, but this is really messy.

Cuel:
Fixed the first one but moduleDefend uses CBA_fnc_taskDefend which is fixed in first commit

Cuel:
On that note it might also be worth check for `assignedGunner` in these aswell, if taskDefend runs for two groups they should try to assign to the same static even if it's assigned but not mounted?

commy2:
Don't think that amount of micromanaging is necessary. It is possible for another AI or player to enter the vehicle before the assigned AI can reach for example, and that would still be not covered by that check.
At that point, it is the responsibility of the mission maker.

commy2:
@Cuel ready to merge?

Cuel:
Yap

PR #1020, opened 2018-11-04
Add more magazine wells

xrufix:
- add QBZ-95 drum mags to vanilla weapon
- Katiba stick magazine wells
- 762x54 mags to "Zafir"
- Let Vanilla Mk1 EMR use 762x51_AR10 mag well and add Vanilla 7.62x51 to it.

### DIFF
--- a/addons/jam/CfgWeapons.hpp
+++ b/addons/jam/CfgWeapons.hpp
@@ -49,7 +49,7 @@ class CfgWeapons {
         magazineWell[] = {"CBA_580x42_TYPE95"};
     };
     class arifle_CTARS_base_F : Rifle_Base_F {
-        magazineWell[] = {"CBA_580x42_TYPE95"};
+        magazineWell[] = {"CBA_580x42_TYPE95", "CBA_580x42_TYPE95_XL"};
     };
 
     class arifle_AK12_base_F : Rifle_Base_F {
--- a/addons/jam/CfgWeapons.hpp
+++ b/addons/jam/CfgWeapons.hpp
@@ -42,6 +42,9 @@ class CfgWeapons {
     class EBR_base_F: Rifle_Long_Base_F {
         magazineWell[] = {"CBA_762x51_M14"};
     };
+    class DMR_03_base_F: Rifle_Long_Base_F {
+        magazineWell[] = {"CBA_762x51_M14"};
+    };
     class DMR_06_base_F: Rifle_Long_Base_F {
         magazineWell[] = {"CBA_762x51_M14"};
     };
--- a/addons/jam/CfgWeapons.hpp
+++ b/addons/jam/CfgWeapons.hpp
@@ -79,6 +79,10 @@ class CfgWeapons {
         magazineWell[] = {"CBA_65x39_Mk200"};
     };
 
+    class LMG_Zafir_F : Rifle_Long_Base_F {
+        magazineWell[] = {"CBA_762x51_LINKS", "CBA_762x54R_LINKS"}
+    };
+
     class launch_RPG7_F : Launcher_Base_F {
         magazineWell[] = {"CBA_RPG7"};
     };
--- a/addons/jam/CfgWeapons.hpp
+++ b/addons/jam/CfgWeapons.hpp
@@ -46,7 +46,7 @@ class CfgWeapons {
         magazineWell[] = {"CBA_762x51_M14"};
     };
     class DMR_06_base_F: Rifle_Long_Base_F {
-        magazineWell[] = {"CBA_762x51_M14"};
+        magazineWell[] = {"CBA_762x51_AR10"};
     };
     class arifle_SPAR_03_base_F : Rifle_Base_F {
         magazineWell[] = {"CBA_762x51_HK417"};
--- a/addons/jam/CfgWeapons.hpp
+++ b/addons/jam/CfgWeapons.hpp
@@ -57,7 +57,7 @@ class CfgWeapons {
     };
 
     class arifle_CTAR_base_F : Rifle_Base_F {
-        magazineWell[] = {"CBA_580x42_TYPE95"};
+        magazineWell[] = {"CBA_580x42_TYPE95", "CBA_580x42_TYPE95_XL"};
     };
     class arifle_CTARS_base_F : Rifle_Base_F {
         magazineWell[] = {"CBA_580x42_TYPE95", "CBA_580x42_TYPE95_XL"};
--- a/addons/jam/CfgWeapons.hpp
+++ b/addons/jam/CfgWeapons.hpp
@@ -35,6 +35,10 @@ class CfgWeapons {
         magazineWell[] = {"CBA_65x39_Katiba"};
     };
 
+    class arifle_ARX_base_F : Rifle_Base_F {
+        magazineWell[] = {"CBA_65x39_Katiba"};
+    };
+
     class EBR_base_F: Rifle_Long_Base_F {
         magazineWell[] = {"CBA_762x51_M14"};
     };
--- a/addons/jam/CfgWeapons.hpp
+++ b/addons/jam/CfgWeapons.hpp
@@ -42,7 +42,7 @@ class CfgWeapons {
     class EBR_base_F: Rifle_Long_Base_F {
         magazineWell[] = {"CBA_762x51_M14"};
     };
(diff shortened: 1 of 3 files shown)


addons/jam/CfgWeapons.hpp
```diff
@@ -49,7 +60,7 @@ class CfgWeapons {
         magazineWell[] = {"CBA_580x42_TYPE95"};
     };
     class arifle_CTARS_base_F : Rifle_Base_F {
-        magazineWell[] = {"CBA_580x42_TYPE95"};
+        magazineWell[] = {"CBA_580x42_TYPE95", "CBA_580x42_TYPE95_XL"};
```
robalo: I didn't notice we had the separate drum magwell (why do we need it btw ?).
You should add it to the arifle_CTAR_base_F as well.

addons/jam/CfgWeapons.hpp
```diff
@@ -71,6 +82,10 @@ class CfgWeapons {
         magazineWell[] = {"CBA_65x39_Mk200"};
     };
 
+    class LMG_Zafir_F : Rifle_Long_Base_F {
```
robalo: This is wrong. It's either 7.62x51 or 7.62x54R.

addons/jam/CfgWeapons.hpp
```diff
@@ -31,9 +31,20 @@ class CfgWeapons {
...
+    };
+
     class EBR_base_F: Rifle_Long_Base_F {
         magazineWell[] = {"CBA_762x51_M14"};
     };
+    class DMR_03_base_F: Rifle_Long_Base_F {
+        magazineWell[] = {"CBA_762x51_M14"};
+    };
     class DMR_06_base_F: Rifle_Long_Base_F {
         magazineWell[] = {"CBA_762x51_M14"};
```
robalo: Why ? What weapon is it and does it use M14 mags ?

addons/jam/CfgWeapons.hpp
```diff
@@ -71,6 +82,10 @@ class CfgWeapons {
         magazineWell[] = {"CBA_65x39_Mk200"};
     };
 
+    class LMG_Zafir_F : Rifle_Long_Base_F {
+        magazineWell[] = {"CBA_762x51_LINKS", "CBA_762x54R_LINKS"}
```
robalo: This is wrong. It's either 7.62x51NATO or 7.62x54R.

addons/jam/CfgWeapons.hpp
```diff
@@ -31,9 +31,20 @@ class CfgWeapons {
...
+
+    class arifle_ARX_base_F : Rifle_Base_F {
+        magazineWell[] = {"CBA_65x39_Katiba"};
+    };
+
     class EBR_base_F: Rifle_Long_Base_F {
         magazineWell[] = {"CBA_762x51_M14"};
     };
+    class DMR_03_base_F: Rifle_Long_Base_F {
+        magazineWell[] = {"CBA_762x51_M14"};
```
robalo: Why ? What weapon is it and does it use M14 mags ? Why not AR10 mags ? Or FAL ? Or G3 ?

addons/jam/CfgWeapons.hpp
xrufix: In reality, the Negev NG7 uses 7.62x51 NATO, in the game Zafir uses 7.62x54R.

I think it should support both, so that mission designers can decide if they want to give players NATO or Russian ammo, depending on the scenario.

addons/jam/CfgWeapons.hpp
xrufix: It's a SIG 556 ("Mk 1 EMR") chambered for 7.62x51. It supports the same magazines as vanilla's M14 and HK417.

To mimic Vanilla behaviour, I could add the same magazine wells that are used by other vanilla-DMRs.

addons/jam/CfgWeapons.hpp
robalo: So in game it's a modified Negev NG7 modified to use 7.62x54R. Why not just leave it at that ? More plausible than the weapon being able to use 2 very different ammo types.

xrufix: Because it isn't:

grafik

addons/jam/CfgWeapons.hpp
robalo: Again, an exercise of imagination, because SIG 556 doesn't exist in 7.62x51. Maybe a SIG 716 is a closer match ? In that case it should take AR10 mags, not M14. Yeah, vanilla has even the HK417 use that same mag, but let's keep JAM/magwell stuff realistic, not based off vanilla.

addons/jam/CfgWeapons.hpp
robalo: Yet the vanilla config would disagree:
`    class LMG_Zafir_F : Rifle_Long_Base_F {
        magazines[] = {"150Rnd_762x54_Box", "150Rnd_762x54_Box_Tracer"};
        descriptionShort = "Light Machine Gun<br />Caliber: 7.62x54 mm";
};
`

addons/jam/CfgWeapons.hpp
xrufix: I guess it's kind of a hybrid of both weapons.

Changing it to AR10 mags seems fair to me with the SIG 716 accepting them, but if so, vanilla mags fitting into it should also be categorized as AR10 mags.

I agree with you on keeping stuff realistic, but as long as BI doesn't introduce a lot of new mags, let's also make sure that vanilla weapons are compatible with modded magazines based on vanilla magazines that fit into it.

addons/jam/CfgWeapons.hpp
xrufix: Done, albeit I hate it when magazines are not represented visually.

addons/jam/CfgWeapons.hpp
robalo: Community mods add the AR10 mags and they will fit. Ignore the stray vanilla mod that fits all.
If you add the M14 magwell you will end up with a long array of mags that fit when they shouldn't.
If you put a better match magwell you only end up with the one weird vanilla mag case.
Just my 2c.

addons/jam/magwells_762x51.hpp
```diff
@@ -1,5 +1,9 @@
 
-    class CBA_762x51_AR10 {};       // AR-10 standard mag (20 rounds)
+    class CBA_762x51_AR10 {         // AR-10 standard mag (20 rounds)
+        BI_mags[] = {
+            "20Rnd_762x51_Mag"
```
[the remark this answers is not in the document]
xrufix: Oh yes.

addons/jam/CfgWeapons.hpp
commy2: In game ui text and usage trumps whatever is written on the texture. And being able to use both is super weird.

xrufix: I removed it, maybe I'll just create a copy of it with x54 in a mod of my own.

addons/jam/magwells_762x51.hpp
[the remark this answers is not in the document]
xrufix: I will remove this and add it to another PR, so this PR is clean from any debated commits.

addons/jam/CfgWeapons.hpp
```diff
@@ -31,10 +31,18 @@ class CfgWeapons {
...
+
+    class arifle_ARX_base_F : Rifle_Base_F {
+        magazineWell[] = {"CBA_65x39_Katiba"};
+    };
+
     class EBR_base_F: Rifle_Long_Base_F {
         magazineWell[] = {"CBA_762x51_M14"};
     };
-    class DMR_06_base_F: Rifle_Long_Base_F {
+    class DMR_03_base_F: Rifle_Long_Base_F {
```
robalo: This must be a mistake here. You replaced DMR_06_base_F with DMR_03_base_F

dedmen: https://github.com/CBATeam/CBA_A3/pull/1020/commits/93af9ef5fac23061f933f6ca15bd80b6fb2cbb12 This is the mistake. Commit message says it reverts the SIG 556. But it reverted the M14.

xrufix: Yep, has been fixed, thanks for spotting it!

PR #1022, opened 2018-11-09
add header for PREP() macro

### LINKED ISSUE #1021: CBA PREP() macro could be useful thing to have documented
**CBA Version:**  stable

**Description:**
For example ACE mentions this about the PREP macro:

> Functions shall be created in the functions\ subdirectory, named fnc_functionName.sqf They shall then be indexed via the PREP(functionName) macro in the XEH_preInit.sqf file.
> 
> The PREP macro allows for CBA function caching, which drastically speeds up load times. Beware though that function caching is enabled by default and as such to disable it you need to #define DISABLE_COMPILE_CACHE above your #include "script_components.hpp" include!

As it is useful outside CBA, maybe it should be documented also on the wiki and or here:
http://cbateam.github.io/CBA_A3/docs/files/main/script_macros_common-hpp.html

commy2:
- title
- close #1021

### DIFF
--- a/addons/main/script_macros_common.hpp
+++ b/addons/main/script_macros_common.hpp
@@ -905,7 +905,7 @@ Description:
     The PREP macro should be placed in a script run by a XEH preStart and XEH preInit event.
 
     The PREP macro allows for CBA function caching, which drastically speeds up load times.
-    Beware though that function caching is enabled by default and as such to disable it you, need to
+    Beware though that function caching is enabled by default and as such to disable it, you need to
     #define DISABLE_COMPILE_CACHE above your #include "script_components.hpp" include!
 
     The function will be defined in ui and mission namespace. It can not be overwritten without
--- a/addons/main/script_macros_common.hpp
+++ b/addons/main/script_macros_common.hpp
@@ -890,6 +890,39 @@ Author:
 #define SETVARMAIN SETVARMAINS(PREFIX)
 #define IFCOUNT(var1,var2,var3) if (count ##var1 > ##var2) then { ##var3 = ##var1 select ##var2 };
 
+/* -------------------------------------------
+Macro: PREP()
+
+Description:
+    Defines a function.
+
+    Full file path:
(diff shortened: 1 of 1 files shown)


addons/main/script_macros_common.hpp
```diff
@@ -890,6 +890,39 @@ Macro: GVARMAIN()
...
+    Full file path:
+        '\MAINPREFIX\PREFIX\SUBPREFIX\COMPONENT\fnc_<FNC>.sqf'
+
+    Resulting function name:
+        'PREFIX_COMPONENT_<FNC>'
+
+    The PREP macro should be placed in a script run by a XEH preStart and XEH preInit event.
+
+    The PREP macro allows for CBA function caching, which drastically speeds up load times.
+    Beware though that function caching is enabled by default and as such to disable it you, need to
```
jonpas: to disable it, you need

commy2: fixed

addons/main/script_macros_common.hpp
```diff
@@ -936,6 +962,13 @@ Macro: PATHTO_FNC()
 Author:
     dixon13, commy2
  ------------------------------------------- */
+#ifdef RECOMPILE
```
dedmen: Now the PATHTO_FNC header looks like it applies to recompile.

commy2: The alternative is having `RECOMPILE` to look like it belongs to `PREP` when it belongs to `PATHTO_FNC` only.

dedmen: maybe move it below PATHTO_FNC then? So that the header is still directly above what it belongs to

commy2: `PATHTO_FNC` uses `RECOMPILE`. I think it would break the way you described.

dedmen: Macro inside a different macro is only resolved when the second macro is used.
PATHTO_FNC is resolved first. Then the RECOMPILE inside it is resolved.
So it wouldn't cause any problems if it's defined after the PATHTO_FNC macro

PR #1023, opened 2018-11-14
Update JR for 1.86

robalo:
Convert ADR-97 rails to JR

### DIFF
--- a/addons/jr/cfgweapons.hpp
+++ b/addons/jr/cfgweapons.hpp
@@ -937,18 +937,7 @@ class CfgWeapons {
     class SMG_03_TR_BASE : Rifle_Base_F {
         class WeaponSlotsInfo;
     };
-    class SMG_03C_BASE : SMG_03_TR_BASE {
-        class WeaponSlotsInfo : WeaponSlotsInfo {
-            class CowsSlot : asdg_OpticRail1913_short {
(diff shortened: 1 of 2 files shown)


commy2:
They put it into vanilla?

jonpas:
With next update, yes. Already in RC.

jonpas:
Apparently they can even fit .22 rated suppressors. I am not sure if this is an off case (in which case we could support it) or a general one (in which case we would have to make it more detailed, which isn't worth it).

Drofseh:
.22, 5.56, and 5.7 suppressors would all have the same diameter hole, so the only problems would be how it attaches and if the suppressor is strong enough to container the pressure from the blast.

I wouldn't want to use a suppressor meant for .22LR on a 5.7 or 5.56 gun unless it was robust.

Kllrt:
I think that 7.62x39mm vz58 is missing :D

dedmen:
Well do CZ805 magazines also fit into G36? Then I'd say just reuse it.

Kllrt:
This is what I am not sure of, sadly...

robalo:
What do magazines have to do with this PR ? Wrong thread ?

dedmen:
>What do magazines have to do with this PR ?

Both are about vanilla :D

PR #1026, opened 2018-11-20
Exit early for invalid PFH handle

Cuel:
title, avoid the directCall

Might just be me but for repeating-temporary PFHs I use RETDEF() with -1 quite often

### DIFF
--- a/addons/common/fnc_removePerFrameHandler.sqf
+++ b/addons/common/fnc_removePerFrameHandler.sqf
@@ -23,6 +23,7 @@ Author:
 ---------------------------------------------------------------------------- */
 
 params [["_handle", -1, [0]]];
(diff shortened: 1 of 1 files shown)


commy2:
Two problems with this.

First of all, the exitWith is not inside the directCall, which means that the function is no longer scheduler safe after this PR. This could be fixed by moving the exitWith inside the directCall.

Secondly, the current L31 already exits the function pretty early. Not only for -1, but also -2 or positive out of bounds, or for already deleted ids.
You may make this a tiny bit faster for your case, but you also add a redundant check for everyone else. I feel like this PR makes the function on average slower.

Cuel:
Alright

commy2:
Ideas for improvements are always appreciated though!

PR #1030, opened 2018-11-21
WIP: Arma vanilla magwells

dedmen:
This is how it would look if we replace our magwells by the vanilla ones.
Need feedback. What do we do? 

Leave the BI magwells be and just make our own duplicates? 
Add inheritance? Problem with that is, mods that already add onto the CBA classes will break when we add inheritance. But right now it's soon enough that we can still do it.

Also the BI magwells are missing the reload tracers
30Rnd_556x45_Stanag_green, 30Rnd_556x45_Stanag_red and the mx variants.

### DIFF
--- a/addons/jam/CfgWeapons.hpp
+++ b/addons/jam/CfgWeapons.hpp
@@ -5,96 +5,96 @@ class CfgWeapons {
     class Launcher_Base_F;
 
     class mk20_base_F: Rifle_Base_F {
-        magazineWell[] = {"CBA_556x45_STANAG", "CBA_556x45_STANAG_L", "CBA_556x45_STANAG_XL", "CBA_556x45_STANAG_2D"};
+        magazineWell[] += {"CBA_556x45_STANAG", "CBA_556x45_STANAG_L", "CBA_556x45_STANAG_XL", "CBA_556x45_STANAG_2D"};
     };
     class SDAR_base_F: Rifle_Base_F {
-        magazineWell[] = {"CBA_556x45_STANAG"};
+        magazineWell[] += {"CBA_556x45_STANAG"};
     };
     class Tavor_base_F: Rifle_Base_F {
-        magazineWell[] = {"CBA_556x45_STANAG", "CBA_556x45_STANAG_L", "CBA_556x45_STANAG_XL", "CBA_556x45_STANAG_2D"};
+        magazineWell[] += {"CBA_556x45_STANAG", "CBA_556x45_STANAG_L", "CBA_556x45_STANAG_XL", "CBA_556x45_STANAG_2D"};
     };
     class arifle_SPAR_01_base_F : Rifle_Base_F {
-        magazineWell[] = {"CBA_556x45_STANAG", "CBA_556x45_STANAG_L", "CBA_556x45_STANAG_XL", "CBA_556x45_STANAG_2D", "CBA_556x45_STANAG_2D_XL"};
+        magazineWell[] += {"CBA_556x45_STANAG", "CBA_556x45_STANAG_L", "CBA_556x45_STANAG_XL", "CBA_556x45_STANAG_2D", "CBA_556x45_STANAG_2D_XL"};
     };
     class arifle_SPAR_02_base_F : Rifle_Base_F {
-        magazineWell[] = {"CBA_556x45_STANAG", "CBA_556x45_STANAG_L", "CBA_556x45_STANAG_XL", "CBA_556x45_STANAG_2D", "CBA_556x45_STANAG_2D_XL"};
+        magazineWell[] += {"CBA_556x45_STANAG", "CBA_556x45_STANAG_L", "CBA_556x45_STANAG_XL", "CBA_556x45_STANAG_2D", "CBA_556x45_STANAG_2D_XL"};
     };
 
     class UGL_F : GrenadeLauncher {
-        magazineWell[] = {"CBA_40mm_M203", "CBA_40mm_EGLM"};
+        magazineWell[] += {"CBA_40mm_M203", "CBA_40mm_EGLM"};
     };
 
     class arifle_MX_Base_F : Rifle_Base_F {
-        magazineWell[] = {"CBA_65x39_MX", "CBA_65x39_MX_XL"};
+        magazineWell[] += {"CBA_65x39_MX", "CBA_65x39_MX_XL"};
         class GL_3GL_F : UGL_F {
-            magazineWell[] = {"CBA_40mm_3GL", "CBA_40mm_M203", "CBA_40mm_EGLM"};
+            magazineWell[] += {"CBA_40mm_3GL", "CBA_40mm_M203", "CBA_40mm_EGLM"};
         };
     };
 
     class arifle_Katiba_Base_F : Rifle_Base_F {
-        magazineWell[] = {"CBA_65x39_Katiba"};
+        magazineWell[] += {"CBA_65x39_Katiba"};
     };
 
     class arifle_ARX_base_F : Rifle_Base_F {
-        magazineWell[] = {"CBA_65x39_Katiba"};
+        magazineWell[] += {"CBA_65x39_Katiba"};
     };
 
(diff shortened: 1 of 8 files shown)


jonpas:
> Need feedback. What do we do?

Inherit CBA ones from BI ones, remove 1 or 3 versions later.

jonpas:
Drop the "This is" and just write `// DEPRECATED! Don't use! Use vanilla one instead!`. The shorter it is, more people read it.

commy2:
You should name "the vanilla one".

jonpas:
`DEPRECATED! Use BI_wellName!`

addons/jam/magwells_65x39.hpp
```diff
@@ -1,28 +1,40 @@
...
             "30Rnd_65x39_caseless_green_mag_Tracer"
         };
     };
+    
+    class Mk200_65x39 {};
+    
+    class Katiba_65x39 {};
+    
+    
+
```
commy2: borked

addons/jam/magwells_762x39.hpp
```diff
@@ -1,11 +1,13 @@
...
+    class CBA_762x39_AK {               //This is DEPRECATED! Don't use it! Use the vanilla one instead
         BI_mags[] = {
             "30Rnd_762x39_Mag_F",
             "30Rnd_762x39_Mag_Green_F",
             "30Rnd_762x39_Mag_Tracer_F",
             "30Rnd_762x39_Mag_Tracer_Green_F"
         };
     };
+	
+	class AK_762x39 {};                 // Standard AK-47/AKM magazines
```
commy2: TAB!!!

dedmen: *panics and runs against the door*

robalo:
I would like to see how the vanilla classes look like.  Where are they ? Dev ? RC ?
The mix of non-tagged with CBA tagged classes looks a bit messy at first sight.

robalo:
My 0.02€

Unless vanilla implementation promises to get as developed as the current CBA one, deprecation should work the other way around. Deprecate vanilla classes and use a the complete and consistent CBA classes.

Kllrt:
@robalo They look like this https://pastebin.com/bfa806BH

Soldia1138:
What’s about e.g. the 100 Round MX magazines? The are commented out in the pasteboard but belong to the same config like the 30 round MX. 
CBA has two different entries there. 
The potential of conflicts is pretty high with a mixed approach. 

Would be easier for us to stay on CBA-only magwells

dedmen:
Weapon and magazine mod makers will always want to use the Vanilla variant, because they also want to be compatible with users who are not using CBA.
Meaning if we decide to keep a duplicate, they will have to add their magazines to vanilla and CBA. And they will also have to consume the magazines from vanilla and CBA, causing duplicates everywhere.
We can't really deprecate vanilla classes, as we cannot just force every weapon mod maker to depend on CBA.

robalo:
I'd rather have just a few duplicates - seeing how few the vanilla classes are (if those are real) - than confuse the hell out of everyone with a mixed approach.

dedmen:
>if those are real

@Kllrt is a BI dev btw.

robalo:
> > if those are real
> 
> @Kllrt is a BI dev btw.

I had no idea :D 🙇

PabstMirror:
If we are going to keep vanilla magwells I guess we might want something like
```
class arifle_MX_Base_F: Rifle_Base_F {
    magazineWell[]+= {"CBA_65x39_MX", "CBA_65x39_MX_XL"};
```

dedmen:
Btw why do we indent everything in the include files by one level?

commy2:
Waste of space.
The #include is already indented.

robalo:
Noticed the magwells were added in latest update. IMO to keep the implementation clean we should go with inheriting from the base classes. Even if it will cause some updating classes RPT errors, it won't break the game and they will be spotted and fixed easily.

What I don't like and expect will happen regardless of how we do it is modders throwing all their mags in the vanilla classes without care for the standard/large/drums etc distinction we have already. I still think that we should inverse the deprecation messages and say for example don't use AK_762x39, use CBA_762x39_AK and CBA_762x39_RPK instead.

dedmen:
>we should go with inheriting from the base classes

Inheritance was broken. And I didn't hear of a fix so far.

robalo:
Hmm, dunno about inheritance broken. Seems fine to me:
image

PabstMirror:
won't that lead to problems with mod load order?

addons/jam/CfgWeapons.hpp
```diff
@@ -5,96 +5,96 @@ class CfgWeapons {
...
-        magazineWell[] = {"CBA_40mm_M203", "CBA_40mm_EGLM"};
+        magazineWell[] += {"CBA_40mm_M203", "CBA_40mm_EGLM"};
     };
 
     class arifle_MX_Base_F : Rifle_Base_F {
-        magazineWell[] = {"CBA_65x39_MX", "CBA_65x39_MX_XL"};
+        magazineWell[] += {"CBA_65x39_MX", "CBA_65x39_MX_XL"};
         class GL_3GL_F : UGL_F {
-            magazineWell[] = {"CBA_40mm_3GL", "CBA_40mm_M203", "CBA_40mm_EGLM"};
+            magazineWell[] += {"CBA_40mm_3GL", "CBA_40mm_M203", "CBA_40mm_EGLM"};
```
robalo: There is no 40mm magwell in vanilla.

robalo:
How about this stupid idea: we go with all new untagged classes, vanilla style, extending current vanilla implementation, deprecating everything CBA_ in magwells.

Drofseh:
It might work, but it'll be kind of a pain.

PabstMirror:
same problem the other way,
someone makes mod for vanilla content and UBCs the vanilla well

maybe this PR will be ok, anyone who supported CBA mag wells in the last month is probably going to be able to update again
probably want to manually re-add the vanilla mags so even with a UBC they are still there

imho, the simplest solution is to just leave it how it is now and mod makers can just add both wells if they want to support both

dedmen:
>Hmm, dunno about inheritance broken. Seems fine to me:

Yes. But Arma doesn't see it that way. The magazine code doesn't look through parent entries.

addons/jam/CfgWeapons.hpp
dedmen: += won't break if the array exists right? So it doesn't really matter

jonpas:
Well that's what you get, it's in vanilla, that's a big enough pro for a small wait of 2-3 months.

addons/jam/CfgWeapons.hpp
robalo: Yep, but that's the thing, doesn't exist, same thing on the UGL_F parent.

robalo:
So which way from here:
1. Stick with what we got done, vanilla and CBA magwells can live happily side by side attached to weapons as modders see fit depending on whether they need/want to support CBA JAM.
2. Figure out new class naming convention that looks consistent with vanilla naming and still makes sense and propose BIS to include them or at least be aware of them.

For 1. I still feel that we should not deprecate existing CBA magwells in favor of vanilla, Keeping them may cause some duplication to pop up but don't see what harm that may cause really.

jonpas:
I vote 2, 1 if 2 won't go through.

addons/jam/CfgWeapons.hpp
robalo: Last I checked this kind of array addition combined with inheritance was really finicky.
The above bits seem fine though, did a quick check and magazineWell is populated fine.

commy2:
Voting for vanilla in parallel with CBA.

commy2:
Is this obsolete now? Close?

dedmen:
@commy2 yes.
But this
https://github.com/CBATeam/CBA_A3/pull/1030/files#diff-a59084d33a95ff61239e5b86ef372144L8
+= is still needed. See comment on the 1.92 magwell PR. We currently break mods and might break vanilla again with next update.

PR #1039, opened 2018-12-03
getConfigEntry tweaks

commy2:
- make function recognize SQF expressions that can be read using getNumber
- support keyword "STRING" and "TEXT"
- header fixes
- standard variable names

### DIFF
--- a/addons/common/fnc_getConfigEntry.sqf
+++ b/addons/common/fnc_getConfigEntry.sqf
@@ -9,16 +9,16 @@ Description:
     supplied default value.
 
 Parameters:
-    _configEntry - Entry to get value of [Config]
-    _entryType - "text", "number" or "array" [String]
-    _defaultValue - Value to return if config entry unavailable [Any]
+    _configEntry - Entry to get value of <CONFIG>
+    _entryType - "STRING", "NUMBER" or "ARRAY" <STRING>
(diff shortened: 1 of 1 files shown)


Dystopian:
Don't you want to remove `_return` variable and return from `switch` directly to slightly speedup the function?

addons/common/fnc_getConfigEntry.sqf
```diff
@@ -28,24 +28,28 @@ SCRIPT(getConfigEntry);
...
 params ["_configEntry", "_entryType", "_defaultValue"];
 
-private _r = _defaultValue;
-
-switch (toLower _entryType) do {
-    case "text": {
-        if (isText _configEntry) then {
-            _r = getText _configEntry;
-        };
+if (toUpper _entryType in ["STRING", "TEXT"]) exitWith {
```
PabstMirror:
```suggestion
if (_entryType == "STRING" || _entryType == "TEXT") exitWith {
```

bux: no `toUpper`?

PabstMirror: `==` isn't case sensitive

Dystopian: @PabstMirror's code is a little faster

Dystopian:
tested, works good

PR #1041, opened 2018-12-04
JAM update for 1.86

robalo:
Updates instead of overriding vanilla magwells.
Inherit from existing vanilla magwells.
Remove leading spaces.

### DIFF
--- a/addons/jam/CfgMagazineWells.hpp
+++ b/addons/jam/CfgMagazineWells.hpp
@@ -69,9 +69,12 @@ class CfgMagazineWells {
     class CBA_Panzerschreck {}; // Panzerschreck RPzB 54
     class CBA_PIAT {};          // PIAT
 
-    class CBA_RPG7 {
+    class RPG7;
+    class CBA_RPG7: RPG7 {
+    /*
         BI_rockets[] = {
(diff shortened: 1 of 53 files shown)


robalo:
Just to see what I was talking about in #1030

robalo:
I see what you mean. Existing implementation can override the inheritance. Could've left the vanilla mags un-commented in the CBA child classes to avoid losing the, that way.

dedmen:
Did you test if these actually work? As I said, inheritance was broken and reyhard never told me if he fixed it.

PabstMirror:
```
    class arifle_MX_test_F : arifle_MX_SW_F {
        displayName = "0 Test Gun";
        scope = 2;
        magazineWell[] = {"CBA_65x39_MX"};
        magazines[] = {};
    };
```
Config wise it should work, but I can't load the mag in
image


Good call @dedmen, biki has same warning

`// It's possible to add new arrays to class but for now inheritance (i.e. STANAG_556x45_New: STANAG_556x45 {};)  is not supported `

robalo:
This PR is a fail, scrap it, concentrate on #1030

PR #1042, opened 2018-12-06
ignore inherited arrays in magwells

commy2:
- title
- should accurately reflect vanilla behaviour now
- see https://github.com/CBATeam/CBA_A3/pull/1041#issuecomment-444558924


dedmen:
We need to remember to "unfix" this again once BI fixed it on their side.

commy2:
>We need to remember to "unfix" this again once BI fixed it on their side.

Sure.

PR #1043, opened 2019-01-05
Fix sqf_validator to work with # command

neilzar:
- Rename semiColumn to correct spelling semiColon
- Fix the # command to check if there is only whitespace before it, to differentiate between preprocessor command and select command.


dedmen:
>Rename semiColumn to correct spelling semiColon

You're a god.

PR #1045, opened 2019-01-18
sort maps alphabetically

commy2:
- This makes it so the maps are sorted alphabetically on the mission creation screen.
- The missions are already sorted this way, so the maps currently being sorted by config order is weird.


addons/ui/fnc_initDisplayRemoteMissions.sqf
```diff
@@ -5,6 +5,7 @@ params ["_display"];
 private _ctrlMaps = _display displayCtrl IDC_SERVER_ISLAND;
 private _ctrlMissions = _display displayCtrl IDC_SERVER_MISSION;
 
+lbSort _ctrlMaps;
```
dedmen: This is crap. Way too few lines. We need moar lines! Give us our lines!

PR #1049, opened 2019-01-26
add CBA_fnc_selectRandomArray

neilzar:
- Adds a function that will select a specified amount of elements from an array at random without picking the same element multiple times.

Unsure if the function name is the best option, but I think it might become too long otherwise.

### DIFF
--- a/addons/arrays/fnc_selectRandom.sqf
+++ b/addons/arrays/fnc_selectRandom.sqf
@@ -1,0 +1,37 @@
+#include "script_component.hpp"
+/* ----------------------------------------------------------------------------
+Function: CBA_fnc_selectRandom
+
+Description:
+    Select a specified amount of elements from an array without picking the same element multiple times.
+
+Parameters:
+    _array - Input Array <ARRAY>
+    _amount - Block <NUMBER>
+
+Returns:
+    New array with the specified amount of randomly selected elements <ARRAY>
+
+Example:
+    (begin example)
+        _result = [[1, 2, 3, 4, 5], 2] call CBA_fnc_selectRandom;
+        // _result => [2, 4] (random)
+    (end)
+
+Author:
+    NeilZar
+---------------------------------------------------------------------------- */
+SCRIPT(selectRandom);
+
+params [["_array", [], [[]]], ["_amount", 0, [0]]];
+
+if (_amount >= count _array) exitWith { _array };
+if (_amount <= 0) exitWith { [] };
+
+private _result = [];
+
+for "" from 0 to _amount - 1 do {
+    _result pushBack (_array deleteAt (floor random (count _array)))
+};
+
+_result
--- a/addons/arrays/fnc_selectRandom.sqf
+++ b/addons/arrays/fnc_selectRandom.sqf
@@ -30,7 +30,7 @@ if (_amount <= 0) exitWith { [] };
 
 private _result = [];
 
-for "_i" from 0 to _amount - 1 do {
+for "_i" from 1 to _amount do {
     _result pushBack (_array deleteAt floor random count _array)
 };
 
--- a/addons/arrays/fnc_selectRandom.sqf
+++ b/addons/arrays/fnc_selectRandom.sqf
@@ -7,7 +7,7 @@ Description:
 
 Parameters:
     _array - Input Array <ARRAY>
-    _amount - Block <NUMBER>
+    _amount - Amount to select <NUMBER>
 
 Returns:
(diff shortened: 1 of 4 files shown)


addons/arrays/fnc_selectRandom.sqf
```diff
@@ -0,0 +1,37 @@
...
+    (end)
+
+Author:
+    NeilZar
+---------------------------------------------------------------------------- */
+SCRIPT(selectRandom);
+
+params [["_array", [], [[]]], ["_amount", 0, [0]]];
+
+if (_amount >= count _array) exitWith { _array };
```
PabstMirror: should just just be `>`?
I feel like `[[1, 2, 3, 4], 4] call CBA_fnc_selectRandom;` should still do a shuffle

commy2: In my opion, even with `_amount` > `count _array`, this should do a shuffle, ich which case the functions return value should have less elements than `_amount`.

commy2:
Atm. this function has inconsistent behaviour about modifying the input array by reference.
```sqf
private _original = [1,2,3];
private _selected = [_original, 2] call CBA_fnc_selectRandom;

hint str _original; // []
```
vs.
```sqf
private _original = [1,2,3];
private _selected = [_original, 10] call CBA_fnc_selectRandom;

hint str _original; // [1,2,3]
```

I think a function like this should not modify its arguments on the principle of avoiding side effects.

The solution would be to deep copy after the `params` line:
```sqf
_array = + _array;
```

But if the array is itself made up of arrays, then it may be interesting to not copy these sub arrays as well, but to keep their references. This is an edge case, but I personally would expect that behaviour. This could be implemented easily by doing a shallow copy of the input array instead:
```sqf
_array = [] + _array;
```

addons/arrays/fnc_selectRandom.sqf
```diff
@@ -0,0 +1,37 @@
...
+if (_amount <= 0) exitWith { [] };
```
commy2: This line is superfluous. 

Suppose it were deleted and `_amount` would be less or equal than zero.
```sqf
//if (_amount <= 0) exitWith { [] };

private _result = [];

for "_i" from 0 to _amount - 1 do {
    _result pushBack (_array deleteAt floor random count _array)
};

_result
```
Then `_amount - 1` would always be less than zero.
A for loop from `A` to `B` where `B` < `A` would just be ignored. The code block would never be executed.
This means that `_result` would be set to `[]` and never be modified. The return value would always be `[]`, so the same as the return value of the `exitWith` block.

By deleting this line, you also remove one additional exit point from the function, and having as few of these as possible (preferably none) is generally a sign of good clean code.

addons/arrays/fnc_selectRandom.sqf
```diff
@@ -0,0 +1,37 @@
+#include "script_component.hpp"
+/* ----------------------------------------------------------------------------
+Function: CBA_fnc_selectRandom
+
+Description:
+    Select a specified amount of elements from an array without picking the same element multiple times.
+
+Parameters:
+    _array - Input Array <ARRAY>
+    _amount - Block <NUMBER>
```
Cuel: Block doesn't seem like a good description?

addons/arrays/fnc_selectRandom.sqf
```diff
@@ -0,0 +1,37 @@
...
+
+params [["_array", [], [[]]], ["_amount", 0, [0]]];
+
+_amount = _amount min (count _array);
+_array = + _array;
+
+private _result = [];
+
+for "_i" from 1 to _amount do {
+    _result pushBack (_array deleteAt floor random count _array)
```
commy2:
```suggestion
    _result pushBack (_array deleteAt floor random count _array);
```

commy2:
OK, now all we need is a better name. `selectRandom` is a command that picks one element at random, so what about `CBA_fnc_selectSomeRandom` or `CBA_fnc_selectRandomArray`?

neilzar:
I think `selectSomeRandom` is more descriptive.

commy2:
Anyone else prefers `selectRandomArray`? Makes me think of `inArea` vs. `inAreaArray`, `isEqualType` vs. `isEqualTypeArray`, `actionKeysNames` vs. `actionKeysNamesArray` etc.
The only command with "Some" is `someAmmo`.

dedmen:
`selectRandomArray` sounds like it would select a random array out of an array of arrays

commy2:
It selects an array of elements out of an array as opposed to a single element out of an array. The name fits really well.

PR #1051, opened 2019-02-08
Overhaul activateAddons

dedmen:
Cache the addons list at game start instead of always iterating through config at preInit.
Create a "sparse" list with only the addons that really need to be passed to `activateAddons` because activateAddons is VERY slow.

CBA common preInit is the 3rd slowest in my modset, and this code is the main cause.

Left of the red line you can see the iteration through config right after the preprocess and compile for the init functions. Red line is the call to activateAddons.


addons/common/XEH_preInit.sqf
```diff
@@ -47,10 +47,8 @@ call COMPILE_FILE(init_perFrameHandler);
 call COMPILE_FILE(init_delayLess);
 
 // Due to activateAddons being overwritten by eachother (only the last executed command will be active), we apply this bandaid
-private _addons = ("true" configClasses (configFile >> "CfgPatches")) apply {configName _x};
-
-activateAddons _addons;
-GVAR(addons) = _addons;
+activateAddons (uiNamespace getVariable [QGVAR(addonsSparse), []]);
+GVAR(addons) = (uiNamespace getVariable [QGVAR(addons), []]);
```
commy2: Why the parentheses?

addons/common/XEH_preStart.sqf
```diff
@@ -0,0 +1,31 @@
...
+LOG("addonsFull: " + str count _addonsFull);
+
+//Filter out addons that don't have any units defined as we don't need to activate these
+private _addonsSparse = _addonsFull select {
+    !(getArray(_cfgPatches >> _x >> "units") isEqualTo [])
+};
+LOG("addonsSparse1: " + str count _addonsSparse);
+
+//Filter out addons defined in CfgAddons as they are always activated
+private _cfgAddonsCategories = ("true" configClasses (configFile >> "CfgAddons"));
```
commy2: Why the parentheses?

commy2:
"sparse"?

addons/common/XEH_preStart.sqf
dedmen: Because I'm a fox and you're an angel u cocoon.

dedmen:
https://en.wikipedia.org/wiki/Sparse_file. Got a better name? 


Before:
tracy_2019-02-08_13-20-46
3,334 is the config lookup. And the empty space is the activate Addons.
314ms total for preinit
After:

tracy_2019-02-08_13-39-51
115ms for activate addons. 155ms total for preInit

Prestart:
tracy_2019-02-08_13-41-56

By filtering out addons with empty `units` we get from 3342 down to 1181 (with my modset)
CfgAddons stuff is apparently broken. whoops.

commy2:
Test with only CBA, CBA + ACE, and then with whatever you're using and compare them all.

dedmen:
tracy_2019-02-08_13-59-17
Fixed. Only 412 addons left if you filter out empty ones or already loaded ones.
That's 412 addons passed to activateAddons, instead of previously 3342.

common preinit down to 114ms for activateAddons and 137ms total time.


>and compare them all

I have things to do you know?

One run takes like 5 minutes.

PabstMirror:
ace uses `activatedAddons` command which dropped from count `1419` down to `1312`

https://github.com/acemod/ACE3/blob/master/addons/common/scripts/checkVersionNumber.sqf#L10

We should probably just start using `cba_common_addons`, which should contain the original values

commy2:
`activatedAddons` definitely should not be used, because even some modules break it.
On the other hand, I am not sure we should also break it, just because of some milliseconds.

jonpas:
> I have things to do you know?

So do most people...

addons/common/XEH_preStart.sqf
```diff
@@ -0,0 +1,24 @@
+#include "script_component.hpp"
+
+//See usage in XEH_preInit
+private _cfgPatches = configFile >> "CfgPatches";
+private _addonsFull = ("true" configClasses _cfgPatches) apply {configName _x};
+
+//Filter out addons that don't have any units defined as we don't need to activate these
+private _addonsSparse = _addonsFull select {
```
commy2:
```suggestion
 private _unitAddons = _addonsFull select {
```

addons/common/XEH_preStart.sqf
```diff
@@ -0,0 +1,24 @@
...
+    !(getArray(_cfgPatches >> _x >> "units") isEqualTo [])
```
commy2:
```suggestion
    !(getArray (_cfgPatches >> _x >> "units") isEqualTo [])
```

addons/common/XEH_preStart.sqf
```diff
@@ -0,0 +1,24 @@
...
+};
+
+//Filter out addons defined in CfgAddons as they are always activated
+private _cfgAddonsCategories = "true" configClasses (configFile >> "CfgAddons");
+
```
commy2:
```suggestion
```

addons/common/XEH_preStart.sqf
```diff
@@ -0,0 +1,24 @@
...
+private _cfgAddonsMods = [];
+
+{
+    _cfgAddonsMods append (getArray (_x >> "list"));
```
commy2:
```suggestion
    _cfgAddonsMods append getArray (_x >> "list");
```

addons/common/XEH_preStart.sqf
```diff
@@ -0,0 +1,24 @@
...
+} forEach _cfgAddonsCategories;
+
+_addonsSparse = _addonsSparse - _cfgAddonsMods;
```
commy2:
```suggestion
_unitAddons = _unitAddons - _cfgAddonsMods;
```

addons/common/XEH_preStart.sqf
```diff
@@ -0,0 +1,24 @@
...
+
+uiNamespace setVariable [QGVAR(addons), _addonsFull];
+uiNamespace setVariable [QGVAR(addonsSparse), _addonsSparse];
```
commy2:
```suggestion
uiNamespace setVariable [QGVAR(unitAddons), _unitAddons];
```

addons/common/XEH_preInit.sqf
```diff
@@ -47,10 +47,8 @@ call COMPILE_FILE(init_perFrameHandler);
 call COMPILE_FILE(init_delayLess);
 
 // Due to activateAddons being overwritten by eachother (only the last executed command will be active), we apply this bandaid
-private _addons = ("true" configClasses (configFile >> "CfgPatches")) apply {configName _x};
-
-activateAddons _addons;
-GVAR(addons) = _addons;
+activateAddons (uiNamespace getVariable [QGVAR(addonsSparse), []]);
```
commy2:
```suggestion
activateAddons (uiNamespace getVariable [QGVAR(unitAddons), []]);
```

addons/common/XEH_preStart.sqf
```diff
@@ -0,0 +1,24 @@
+#include "script_component.hpp"
+
+//See usage in XEH_preInit
+private _cfgPatches = configFile >> "CfgPatches";
+private _addonsFull = ("true" configClasses _cfgPatches) apply {configName _x};
```
commy2:
```suggestion
private _allComponents = "true" configClasses _cfgPatches apply {configName _x};
```

addons/common/XEH_preStart.sqf
```diff
@@ -0,0 +1,24 @@
...
+
+//Filter out addons that don't have any units defined as we don't need to activate these
+private _addonsSparse = _addonsFull select {
+    !(getArray(_cfgPatches >> _x >> "units") isEqualTo [])
+};
+
+//Filter out addons defined in CfgAddons as they are always activated
+private _cfgAddonsCategories = "true" configClasses (configFile >> "CfgAddons");
```
commy2:
```suggestion
private _allAddons = "true" configClasses (configFile >> "CfgAddons");
```

addons/common/XEH_preStart.sqf
```diff
@@ -0,0 +1,24 @@
...
+
+private _cfgAddonsMods = [];
+
+{
+    _cfgAddonsMods append (getArray (_x >> "list"));
+} forEach _cfgAddonsCategories;
```
commy2:
```suggestion
} forEach _allAddons;
```

addons/common/XEH_preStart.sqf
```diff
@@ -0,0 +1,24 @@
...
+
+_addonsSparse = _addonsSparse - _cfgAddonsMods;
+
+uiNamespace setVariable [QGVAR(addons), _addonsFull];
```
commy2:
```sugestion
uiNamespace setVariable [QGVAR(addons), _allComponents];
```

addons/common/XEH_preStart.sqf
```diff
@@ -0,0 +1,24 @@
...
+
+//Filter out addons that don't have any units defined as we don't need to activate these
+private _addonsSparse = _addonsFull select {
+    !(getArray(_cfgPatches >> _x >> "units") isEqualTo [])
+};
+
+//Filter out addons defined in CfgAddons as they are always activated
+private _cfgAddonsCategories = "true" configClasses (configFile >> "CfgAddons");
+
+private _cfgAddonsMods = [];
```
commy2:
```suggestion
private _preloadedAddons = [];
```

commy2:
We badly need multiline suggestions.

addons/common/XEH_preStart.sqf
dedmen: I voted to not care about case sensitive. People can write stuff in CfgAddons wrong yeah. But I don't want to waste the time tolowering thousands of strings.

addons/common/XEH_preStart.sqf
```diff
@@ -0,0 +1,23 @@
+#include "script_component.hpp"
+
+//See usage in XEH_preInit
+private _cfgPatches = configFile >> "CfgPatches";
+private _allComponents = "true" configClasses _cfgPatches apply {configName _x};
+
+//Filter out addons that don't have any units defined as we don't need to activate these
+private _unitAddons = _addonsFull select {
```
commy2:
```suggestion
private _unitAddons = _allComponents select {
```

addons/common/XEH_preStart.sqf
```diff
@@ -0,0 +1,23 @@
+#include "script_component.hpp"
+
+//See usage in XEH_preInit
+private _cfgPatches = configFile >> "CfgPatches";
+private _allComponents = "true" configClasses _cfgPatches apply {configName _x};
```
commy2:
```suggestion
private _allComponents = "true" configClasses _cfgPatches apply {toLower configName _x};
```

addons/common/XEH_preStart.sqf
```diff
@@ -0,0 +1,23 @@
...
+private _unitAddons = _allComponents select {
+    !(getArray (_cfgPatches >> _x >> "units") isEqualTo [])
+};
+
+//Filter out addons defined in CfgAddons as they are always activated
+private _allAddons = "true" configClasses (configFile >> "CfgAddons");
+private _preloadedAddons = [];
+
+{
+    _preloadedAddons append getArray (_x >> "list");
```
commy2:
```suggestion
    _preloadedAddons append (getArray (_x >> "list") apply {toLower _x});
```

dedmen: Good idea

PR #1052, opened 2019-02-08
Compile and cache configFile eventhandlers at preStart

dedmen:
They won't change after preStart as the config cannot be edited. So no reason to recompile them each preInit.

This is a mission with no eventhandlers in description or campaign config. Meaning 100% of that compileEventHandlers is just configFile. 670ms wasted each preInit for something that can be done once.

PR #1053, opened 2019-02-08
Cache incompatibleClasses at preStart

dedmen:
configFile cannot change, so we might aswell collect that stuff at preStart.


dedmen:
Measurements are here: https://github.com/CBATeam/CBA_A3/pull/1052#issuecomment-461762479
TLDR moved about 600ms of load from preInit to preStart.

PR #1054, opened 2019-02-08
Move flexiMenu functions to CfgFunctions

dedmen:
tracy_2019-02-08_14-30-35

This triggers me.

### DIFF
--- a/addons/ui/flexiMenu/init.sqf
+++ b/addons/ui/flexiMenu/init.sqf
@@ -8,22 +8,6 @@ GVAR(target) = objNull; // current object for interaction
 GVAR(holdKeyDown) = false; // default global behaviour of requiring holding key down. Can be overriden by menus.
 GVAR(hotKeyColor) = ""; // override colour of hotkey. It is set from rsc menu file, upon opening either the "menu" or "list". If not present, a default will be used.
 
-PREP_SUB(flexiMenu,keyDown);
-PREP_SUB(flexiMenu,keyUp);
-PREP_SUB(flexiMenu,menu);
-PREP_SUB(flexiMenu,list);
-PREP_SUB(flexiMenu,getMenuDef);
-PREP_SUB(flexiMenu,getMenuOption);
-PREP_SUB(flexiMenu,menuShortcut);
-PREP_SUB(flexiMenu,mouseButtonDown);
-PREP_SUB(flexiMenu,highlightCaretKey);
-PREP_SUB(flexiMenu,execute);
-
-// TODO: move these to public CfgFunctions soon.
-PREP_SUB(flexiMenu,add);
-PREP_SUB(flexiMenu,remove);
-PREP_SUB(flexiMenu,setObjectMenuSource);
-
 // prevent instant reactivation of menu after selection was made, while key still held down. Value is reset upon key release.
 GVAR(optionSelected) = false;
 // prevent multiple activations of menu due to key press via keyDown. onLoad can sometimes take a few milliseconds to init.
--- a/addons/ui/flexiMenu/init.sqf
+++ b/addons/ui/flexiMenu/init.sqf
@@ -14,7 +14,7 @@ FUNC(highlightCaretKey) = CBA_fnc_flexiMenu_highlightCaretKey;
 FUNC(execute) = CBA_fnc_flexiMenu_execute;
 FUNC(add) = CBA_fnc_flexiMenu_add;
 FUNC(remove) = CBA_fnc_flexiMenu_remove;
-FUNC(setObjectMenuSource) = CBA_fnc_setObjectMenuSource;
+FUNC(setObjectMenuSource) = CBA_fnc_flexiMenu_setObjectMenuSource;
 
 // list of all menu activation keys and associated types
 GVAR(typeMenuSources) = []; // types, keys and menu sources data
--- a/addons/ui/flexiMenu/init.sqf
+++ b/addons/ui/flexiMenu/init.sqf
@@ -1,5 +1,7 @@
 #include "\x\cba\addons\ui\script_component.hpp"
 
+if (isDedicated) exitWith {false};
+
 FUNC(keyDown) = CBA_fnc_flexiMenu_keyDown;
 FUNC(keyUp) = CBA_fnc_flexiMenu_keyUp;
 FUNC(menu) = CBA_fnc_flexiMenu_menu;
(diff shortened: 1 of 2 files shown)


commy2:
Could delete these: https://github.com/CBATeam/CBA_A3/blob/master/addons/ui/script_component.hpp#L83-L86

However, before the functions are named `cba_flexiMenu_fnc_execute`, but now they are named ` `cba_fnc_flexiMenu_execute` etc, or am I missing something?

dedmen:
Might be right yeah. But then we previously had both 
https://github.com/CBATeam/CBA_A3/pull/1054/files#diff-d0bf4bd8fc69fef6ce8641586ab5da29L23
cba_flexiMenu_fnc_add and 
https://github.com/CBATeam/CBA_A3/pull/1054/files#diff-1ed3566cd26e88a81c404dde0895265bR4
cba_fnc_flexuMenu_add

I guess I'll just move them to PREP then?

commy2:
master:
`cba_ui_fnc_execute`

PR:
`cba_fnc_flexiMenu_execute`

addons/ui/flexiMenu/init.sqf
```diff
@@ -8,22 +8,6 @@ GVAR(target) = objNull; // current object for interaction
...
-PREP_SUB(flexiMenu,menuShortcut);
-PREP_SUB(flexiMenu,mouseButtonDown);
-PREP_SUB(flexiMenu,highlightCaretKey);
-PREP_SUB(flexiMenu,execute);
-
-// TODO: move these to public CfgFunctions soon.
-PREP_SUB(flexiMenu,add);
-PREP_SUB(flexiMenu,remove);
-PREP_SUB(flexiMenu,setObjectMenuSource);
-
```
commy2:
```suggestion
FUNC(keyDown) = CBA_fnc_flexiMenu_keyDown;
FUNC(keyUp) = CBA_fnc_flexiMenu_keyUp;
FUNC(menu) = CBA_fnc_flexiMenu_menu;
FUNC(list) = CBA_fnc_flexiMenu_list;
FUNC(getMenuDef) = CBA_fnc_flexiMenu_getMenuDef;
FUNC(getMenuOption) = CBA_fnc_flexiMenu_getMenuOption;
FUNC(menuShortcut) = CBA_fnc_flexiMenu_menuShortcut;
FUNC(mouseButtonDown) = CBA_fnc_flexiMenu_mouseButtonDown;
FUNC(highlightCaretKey) = CBA_fnc_flexiMenu_highlightCaretKey;
FUNC(execute) = CBA_fnc_flexiMenu_execute;
FUNC(add) = CBA_fnc_flexiMenu_add;
FUNC(remove) = CBA_fnc_flexiMenu_remove;
FUNC(setObjectMenuSource) = CBA_fnc_setObjectMenuSource;
```

tbh, this whole subcomponent is cancer.

dedmen: Are you sure about this? could just move the old PREP_SUB that was there to PREP.hpp and get same effect as on master branch

commy2: PREP_SUB is there because the filepath is different. The files are in a sub folder.

dedmen: I know. But does anything speak against just using the old PREP_SUB code in PREP.hpp? Instead of just copying the functions to a new/old name?

commy2: Idk how PREP_SUB works. I just assume it recompiles every time and is nothing like PREP at all.

dedmen: Correct. Direct compile preproc..

addons/ui/flexiMenu/init.sqf
```diff
@@ -32,4 +16,18 @@ GVAR(lastAccessCheck) = [0, -1];
...
+FUNC(list) = CBA_fnc_flexiMenu_list;
+FUNC(getMenuDef) = CBA_fnc_flexiMenu_getMenuDef;
+FUNC(getMenuOption) = CBA_fnc_flexiMenu_getMenuOption;
+FUNC(menuShortcut) = CBA_fnc_flexiMenu_menuShortcut;
+FUNC(mouseButtonDown) = CBA_fnc_flexiMenu_mouseButtonDown;
+FUNC(highlightCaretKey) = CBA_fnc_flexiMenu_highlightCaretKey;
+FUNC(execute) = CBA_fnc_flexiMenu_execute;
+FUNC(add) = CBA_fnc_flexiMenu_add;
+FUNC(remove) = CBA_fnc_flexiMenu_remove;
+FUNC(setObjectMenuSource) = CBA_fnc_setObjectMenuSource;
```
commy2: Why did you move these down here? Now L32/L16f add a function to eventhandler that doesn't exist yet.

commy2:
Can we also delete these silly `description` entries from the other functions? They do nothing.

addons/ui/flexiMenu/init.sqf
dedmen: Because I wanted them all neatly together :D 
I'll move them to the top tomorrow if I remember.

addons/ui/flexiMenu/init.sqf
```diff
@@ -1,5 +1,19 @@
...
```
commy2: Now you moved them above the `isDedicated exitWith` thingy.

dedmen: Yeah. Because what happens if people call the funcs on server? Optimally they should do nothing I guess

dedmen: Leave me alone I don't have kids!

dedmen:
wat?
https://github.com/CBATeam/CBA_A3/pull/1054/files#diff-d0bf4bd8fc69fef6ce8641586ab5da29R17
It's right there.
And right there https://github.com/CBATeam/CBA_A3/pull/1054/files#diff-1ed3566cd26e88a81c404dde0895265bR12
so what?...


should
https://github.com/CBATeam/CBA_A3/pull/1054/files#diff-d0bf4bd8fc69fef6ce8641586ab5da29R17
be
`FUNC(setObjectMenuSource) = CBA_fnc_flexiMenu_setObjectMenuSource;` ?

commy2:
idk, dude, it's the flexi menu dude

commy2:
What about not doing the = stuff and instead to make use of CfgFunctions with a different tag?
https://community.bistudio.com/wiki/Arma_3_Functions_Library#Tag

dedmen:
We need some of them twice anyway as they were defined in CfgFunctions and as a PREP with different names.

commy2:
What I don't want is to introduce new (duplicate) globals with this PR if it can be avoided.

PR #1056, opened 2019-02-08
Resolve XEH redirect handlers

dedmen:
Resolve handlers that just point to a file.
Instead of executing `call compile preprocessFileLineNumbers...` every time before the EH is executed. Detect that pattern and resolve it at compile time and directly compile the target file.
Get's rid of (number of preInit scripts in ACE and CBA) calls to preprocessFile and compile at preInit
Get's rid of (number of postInit scripts in ACE and CBA) calls to preprocessFile and compile at postInit

After all the recent PR's to speed up pre/postInit this should give a nice additional boost.

Q: Should I implement that for Object handlers too? I guess yes because everything that uses the call preproc pattern will benefit, which I assume is done for objects too.

### DIFF
--- a/addons/xeh/fnc_compileEventHandlers.sqf
+++ b/addons/xeh/fnc_compileEventHandlers.sqf
@@ -69,6 +69,28 @@ private _resultNames = [];
         };
 
         if !(_eventFunc isEqualTo "") then {
+		
+			//Optimize "QUOTE(call COMPILE_FILE(XEH_preInit));" down to just the content of the EH script
+			if (!(["compile"] call CBA_fnc_isRecompileEnabled) && //Users might expect their preInit script to be reloaded everytime when debugging
+				{
+					(toLower (_eventFunc select [0,40])) isEqualTo "call compile preprocessfilelinenumbers '"
+					&& 
+					{
+							(_eventFunc select [count _eventFunc -2]) isEqualTo "';"
+					}
+				}
+			) then {
+				private _funcPath = _eventFunc select [40, count _eventFunc - 42];
+				//If there is a quote mark in the path, then something went wrong and we got multiple paths, just skip optimization
+				//Example cause: "call COMPILE_FILE(XEH_preInit);call COMPILE_FILE(XEH_preClientInit)"
+				if (_funcPath find "'" == -1) then {
+					_eventFunc = preprocessFileLineNumbers _funcPath;
+					
+					TRACE_2("eventfunction redirected",_customName,_funcPath);
+				};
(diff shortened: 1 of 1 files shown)


addons/xeh/fnc_compileEventHandlers.sqf
```diff
@@ -69,6 +69,27 @@ private _resultNames = [];
...
+            //Optimize "QUOTE(call COMPILE_FILE(XEH_preInit));" down to just the content of the EH script
+            if (!(["compile"] call CBA_fnc_isRecompileEnabled) && //Users might expect their preInit script to be reloaded everytime when debugging
+                {
+                    (toLower (_eventFunc select [0,40])) isEqualTo "call compile preprocessfilelinenumbers '"
+                    &&
+                    {
+                            (_eventFunc select [count _eventFunc -2]) isEqualTo "';"
+                    }
+                }
+            ) then {
```
commy2:
```sqf
if (
    !(["compile"] call CBA_fnc_isRecompileEnabled) && { // Users might expect their preInit script to be reloaded everytime when debugging
    (toLower (_eventFunc select [0,40])) isEqualTo "call compile preprocessfilelinenumbers '" && {
    (_eventFunc select [count _eventFunc -2]) isEqualTo "';"
}}) then {
```

dedmen:
End result of 8 hours of work


700ms CBA_fnc_compileEventHandlers :heavy_check_mark:
345ms CBA_fnc_supportMonitor :heavy_check_mark:
200ms activateAddons :heavy_check_mark:
530ms pylons postinit :heavy_check_mark:
850ms MCC_fnc_activateAddons :heavy_check_mark: 
475ms ASR_AI preInit compiles :heavy_check_mark:
340ms cba_ui_init :heavy_check_mark:
110ms BWA3 :question:
130ms TFAR static radio compiles :heavy_check_mark:
100ms ace_mindetector compiles :heavy_check_mark: 
XXXms getting rid of all the "compile preprocess" for all CBA/ACE pre/postInit functions.

3.67 seconds faster preInit.

Let's see how much difference the preproc roughly takes. Last run was 4.47s preInit.
After my XEH "fix" preInit is.....  3.69s about one second faster by just moving all this crap out of the way.

commy2:
Is this really worth it though?

dedmen:
Yes. Very much so. Got roughly a second faster preInit.
Without this you simply cannot cache the compiled code of the eventhandler. You are just caching a small snippet that makes no sense to cache and are constantly recompiling the actual eventhandling code.

PR #1057, opened 2019-02-08
Add return value to cba_fnc_preStart

PabstMirror:
Fix
```
 [0,9.635,0,"XEH: PreStart finished."]
 Error in expression <",_x] call bis_fnc_logFormat;
_function = [] call (uinamespace getvariable _x);
>
   Error position: <= [] call (uinamespace getvariable _x);
>
   Error Generic error in expression
 File A3\functions_f\initFunctions.sqf, line 499
```

a3 does
```
			_function = [] call (uinamespace getvariable _x);
			uinamespace setvariable [_x + "_preStart",_function];
```
and `c = call {a=b}` throws error


commy2:
Remember that one time when I said to never end your function with an assignment operator?

PR #1062, opened 2019-02-10
add cba_jam to requiredAddons in cba_main

commy2:
- makes sure main is loaded after jam
- like all other cba components

### DIFF
--- a/addons/main/config.cpp
+++ b/addons/main/config.cpp
@@ -29,6 +29,7 @@ class CfgPatches {
             "cba_diagnostic",
             "cba_help",
             "cba_jr",
+            "cba_jam",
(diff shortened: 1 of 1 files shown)


addons/main/config.cpp
```diff
@@ -29,6 +29,7 @@ class CfgPatches {
             "cba_diagnostic",
             "cba_help",
             "cba_jr",
+            "cba_jam",
```
dedmen: Are these sorted? ja < jr

commy2: It's not sorted alphabetically, but... thematically? Anyway, JR is the older one, so I put it first of these two.

PR #1063, opened 2019-02-12
Get rid of call compile's

dedmen:


dedmen:
Did I find a bug here?
https://github.com/CBATeam/CBA_A3/pull/1063/files#diff-41e9262ceb1f6e75d6d84600ca21259aL41
Old was
https://github.com/CBATeam/CBA_A3/commit/69594070dada9350e5d954fded28a8cdc73c8ca4#diff-a6e4b270ed5933e83a53eef568fd7175R15
name and value got swapped.

jokoho48:
> name and value got swapped.

but the format entries also so i would say no?
also why do you not use currentNamespace and not missionNamespace in getPosFromString to keep the same behavior

dedmen:
Well I assume currentNamespace to be missionNamespace in that place. It clearly tries to resolve variable names of groups an such, and they won't be in uiNamespace or whatever else.

addons/help/XEH_preClientInit.sqf
```diff
@@ -38,7 +38,7 @@ private _fnc_getKeyName = {
     private _result = "^";
 
     {
-        private _keyName = call compile format ["format ['%1', %2]", "%1", keyName _x];
+        private _keyName = format ["%1", keyName _x];
```
commy2: This doesn't looks like it does the same thing.

commy2:
`call compile format["format['%2',%1]",(keyName _x),"%1"];`
`call compile format ["format ['%1', %2]", "%1", keyName _x];`
These two are the same. One has %1 and %2 swapped, but also the args swapped. I did this because 1 comes before 2.

`format ["%1", keyName _x];`
This is completely different and breaks it.

dedmen:
>This is completely different and breaks it.

Explain. Besides the getVariable missing. I forgot that apparently.

dedmen:
`format ["format ['%1', %2]", "%1", keyName _x]` -> `"format ['%1', Enter]"`
`format ['%1', Enter]`  -> `"any"` 
`call compile "any"` wat?

commy2:
Close?

NouberNou:
I don't evene wanna know what was going on here...

commy2:
Just dedmen on his crusade against `compile`.

Some of the changes are actually nice.

PR #1065, opened 2019-02-15
Add macros to skip adding headers to cfgFunctions

PabstMirror:
Adds optional macros to remove both
- CfgFunction headers (which add code for `_fnc_scriptNameParent / _fnc_scriptName`)
- CBA Macro SCRIPT (which adds `scriptName 'PREFIX\COMPONENT\NAME'`)

e.g. `CBA_fnc_currentUnit` is a single line of code, but becomes
```
CBA_fnc_currentUnit = {
    private _fnc_scriptNameParent = if (isNil '_fnc_scriptName') then {'CBA_fnc_currentUnit'} else {_fnc_scriptName};
    private _fnc_scriptName = 'CBA_fnc_currentUnit';
    scriptName _fnc_scriptName;
    scriptName 'cba\common\currentUnit';
    missionNamespace getVariable ["bis_fnc_moduleRemoteControl_unit", player]
```
Changing execution time from 0.0022 ms to 0.0045 ms.

There should be no change to anyone else using CBA macros unless they add these defines.
For CBA both are set in main/script_mod.hpp

### DIFF
--- a/addons/common/script_component.hpp
+++ b/addons/common/script_component.hpp
@@ -1,9 +1,6 @@
 #define COMPONENT common
 #include "\x\cba\addons\main\script_mod.hpp"
 
-#define SKIP_FUNCTION_HEADER
-#define SKIP_SCRIPT_NAME
-
 #ifdef DEBUG_ENABLED_COMMON
     #define DEBUG_MODE_FULL
 #endif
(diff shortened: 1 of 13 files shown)


PabstMirror:
for addons not using the macro `PATHTO_FNC`1 (https://github.com/CBATeam/CBA_A3/blob/master/addons/events/CfgFunctions.hpp)
- convert to use macro and drop description
- just add `CFGFUNCTION_HEADER` to existing classes?

PR #1066, opened 2019-02-15
add custom mission filter

commy2:
- adds Show Custom Missions/Show All Missions filter to CREATE GAME menu
- state is remembered in the user profile
- default state is of course Show All Missions
- removes optional `no_stock_missions`, because obsolete


jonpas:
Lovely!

PR #1071, opened 2019-02-18
Add Magazine Well support for SMAW

Dahlgren:


PabstMirror:
add one for the spotting round?

Dahlgren:
Good catch @PabstMirror, added `CBA_SMAW_SPOTTING_RIFLE`

PR #1072, opened 2019-02-18
Add Magazine Well support for MAAWS

Dahlgren:


addons/jam/CfgMagazineWells.hpp
```diff
@@ -67,6 +67,12 @@ class CfgMagazineWells {
 
     // AT and AA Launchers, ordered lexicographically
     class CBA_Bazooka {};       // M1, M1A1 Bazooka
+    class CBA_Carl_Gustaf {     // MAAWS, RAWS
```
Dahlgren: The launcher's real name is "Carl Gustaf Recoilles Rifle", not sure what we should use as class?

https://en.wikipedia.org/wiki/Carl_Gustaf_recoilless_rifle

Drofseh: I like Carl Gustaf.

MAAWS is a US specific term that none of these other countries use.
https://en.wikipedia.org/wiki/Carl_Gustaf_recoilless_rifle#Users

PR #1073, opened 2019-02-21
Refactor Help Component

commy2:
- streamlining / modernizing some code
- removes random component line from main menu and pause menu
- insteads shows full list in diary


dedmen:
But I liked the random component line ._.
Who looks into diary, like ever?

commy2:
I restored some commented out diary entries (with updated links).
<details><summary>Wiki</summary>



</details>

<details><summary>Bugtracker</summary>



</details>

<details><summary>Credits</summary>



</details>

Credits is basically the same as the main menu random component line and I like the full list (that is less in your face) more.

I think Bugtracker and Wiki are superfluous. And you can't have clickable links in the diary. Should they be removed entirely?

Keybinding is the same, except that I deleted a few newlines at the end of the entry, so the spacing looks prettier.

jonpas:
I thought diary supports HTML including links?

Bug tracker and wiki could stay, but maybe under a single header, not two.

commy2:
It's fake HTML with different tags:
https://community.bistudio.com/wiki/createDiaryRecord

commy2:
<details><summary>ACE and CBA</summary>





</details>

Note that `ace - etc.` will be displayed as `@ace` for most people, since it is the folder name of the addon.

addons/help/XEH_postInit.sqf
```diff
@@ -0,0 +1,43 @@
...
+        } forEach _addonActions;
+
+        _keys = _keys + "<br/>";
+    } forEach _addons;
+
+    // delete last line breaks
+    _keys = _keys select [0, count _keys - 10];
+
+    _unit createDiaryRecord [QGVAR(docs), [localize "STR_CBA_Help_Keys", _keys]];
+    _unit createDiaryRecord [QGVAR(docs), [localize "STR_CBA_Credits", call (uiNamespace getVariable QGVAR(credits))]];
```
PabstMirror: I think this is failing with large modpack (`count call (uiNamespace getVariable "cba_help_credits")` = 33772)
rpt says ` Unknown entity: ' Accessories</font> by Red Hammer Studios</font><br/><font colo`
that part is exactly 3350 chars into the `call (uiName` string

commy2: `createDiaryRecord` has a hard limit for chars?

commy2: It was the & conflicting with the pseudo html of the diary / sqf TEXT.

PR #1074, opened 2019-02-21
fix admin macros

commy2:
- currently (3.10 RC 1), the IS_ADMIN(_LOGGED) macros are broken
- bug introduced in https://github.com/CBATeam/CBA_A3/pull/1048/files
- features a crafty way to use # in single quote strings

### DIFF
--- a/addons/main/script_macros_common.hpp
+++ b/addons/main/script_macros_common.hpp
@@ -1752,7 +1752,8 @@ Example:
 Author:
     commy2
 ------------------------------------------- */
-#define IS_ADMIN serverCommandAvailable '#kick'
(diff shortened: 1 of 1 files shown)


addons/main/script_macros_common.hpp
```diff
@@ -1752,7 +1752,8 @@ Macro: IS_ADMIN
 Author:
     commy2
 ------------------------------------------- */
-#define IS_ADMIN serverCommandAvailable '#kick'
+#define IS_ADMIN_SYS(x) x##kick
```
PabstMirror: this now evaluates to `serverCommandAvailable '"kick"'` which returns false

PR #1061, opened 2019-02-10
Add P90, UMP40/45/9 & Beretta M9 magwells

3Mydlo3:
**When merged this pull request will:**
- Add FN P90 magwell
- Add magwell support to vanilla P90
- Add HK UMP 40/45/9 magwells
- Add Beretta M9 magwell
- Add FN Five-Seven magwell
- Add Zastava M21 magwell
- Add Zastava M76 magwell

PR #1076, opened 2019-02-22
More magwells

Drofseh:
- Change 570x28 to 57x28, since the cartridge is always called the 5.7x28mm
This is more correct and keeps the naming convention of other magwells that do not add trailing zeros.
- Add magwels for the CZ 75, 85, 97, 2075, P-01, P-06, P-07, P-09 for Urban.
- Add .45 ACP fullsize Glock magwell to SMG_01 (Vermin SMG/KRISS Vector)
- Add 9x19mm Scorpion Evo 3 magwell for SMG_02 (Sting 9mm/CZ Skorpion Evo)
- Add 9x19mm MP5 magwell to SMG_05 (Protector 9 mm/MP5K)
* I'm not 100% happy with having these vanilla two vanilla SMGs as 9x19mm since in vanilla they use 9x21mm and because they should not be able to interchange magazines with each other IRL.
However I think it would be confusing for players if weapons that appear to be able to interchange magazines with mod weapons aren't able to (for example the vanilla MP5K and the NIA or CUP MP5's).

### DIFF
--- a/addons/jam/CfgMagazineWells.hpp
+++ b/addons/jam/CfgMagazineWells.hpp
@@ -32,7 +32,7 @@ class CfgMagazineWells {
 
     // Pistol calibre magwells, ordered lexicographically in metric and imperial groups
     #include "magwells_10mmAuto.hpp"    // 10mm Auto | 10mm Automatic | 10x25mm
-    #include "magwells_570x28.hpp"      // 5.70x28mm
+    #include "magwells_57x28.hpp"       // 5.7x28mm
     #include "magwells_762x25.hpp"      // 7.62x25mm Tokarev
     #include "magwells_762x38R.hpp"     // 7.62x38mmR | 7.62 mm Nagant
     #include "magwells_763x25.hpp"      // 7.63x25mm Mauser | .30 Mauser Automatic
--- a/addons/jam/CfgWeapons.hpp
+++ b/addons/jam/CfgWeapons.hpp
@@ -76,7 +76,6 @@ class CfgWeapons {
     class arifle_AKS_base_F: Rifle_Base_F {
         magazineWell[] = {"CBA_545x39_AK", "CBA_545x39_RPK"};
     };
-
     class LMG_03_base_F: Rifle_Long_Base_F {
         magazineWell[] = {"CBA_556x45_MINIMI"};
     };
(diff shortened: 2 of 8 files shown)


PabstMirror:
changing name from #1061 which hasn't been released, so no compat problems

addons/jam/CfgWeapons.hpp
```diff
@@ -101,7 +101,20 @@ class CfgWeapons {
         magazineWell[] = {"CBA_338NM_LINKS"};
     };
 
-    class SMG_03_TR_BASE: Rifle_Base_F {
-        magazineWell[] = {"CBA_570x28_P90"};
+    class SMG_01_Base : Rifle_Short_Base_F {
```
commy2:
```suggestion
    class SMG_01_Base: Rifle_Short_Base_F {
```

addons/jam/CfgWeapons.hpp
```diff
@@ -101,7 +101,20 @@ class CfgWeapons {
...
+        magazineWell[] = {"CBA_45ACP_Glock_Full"};
     };
+
+    class SMG_02_base_F : Rifle_Short_Base_F {
```
commy2:
```suggestion
    class SMG_02_base_F: Rifle_Short_Base_F {
```

addons/jam/CfgWeapons.hpp
```diff
@@ -101,7 +101,20 @@ class CfgWeapons {
...
+        magazineWell[] = {"CBA_9x19_ScorpionEvo3"};
+    };
+
+    class SMG_03_TR_BASE : Rifle_Base_F {
```
commy2:
```suggestion
    class SMG_03_TR_BASE: Rifle_Base_F {
```

addons/jam/CfgWeapons.hpp
```diff
@@ -101,7 +101,20 @@ class CfgWeapons {
...
+        magazineWell[] = {"CBA_57x28_FNP90"};
+    };
+
+    class SMG_05_base_F : Rifle_Short_Base_F {
```
commy2:
```suggestion
    class SMG_05_base_F: Rifle_Short_Base_F {
```

addons/jam/CfgMagazineWells.hpp
```diff
@@ -32,7 +32,7 @@ class CfgMagazineWells {
 
     // Pistol calibre magwells, ordered lexicographically in metric and imperial groups
     #include "magwells_10mmAuto.hpp"    // 10mm Auto | 10mm Automatic | 10x25mm
-    #include "magwells_570x28.hpp"      // 5.70x28mm
+    #include "magwells_57x28.hpp"       // 5.7x28mm
```
commy2: This one is from 3.9 or earlier, right?
```suggestion
    #include "magwells_570x28.hpp"      // 5.7x28mm
```
Have to keep it for bwc.

addons/jam/magwells_57x28.hpp
```diff
@@ -0,0 +1,6 @@
+    class CBA_57x28_FN57 {};        // FN Five-Seven
```
commy2: Renamed. See above.

addons/jam/CfgMagazineWells.hpp
jonpas: No, it's from #1061.

commy2: Then I support changing this now.

Drofseh:
@commy2 requested changes done.

PR #1079, opened 2019-02-25
add scriptNames to functions for debugging

PabstMirror:
Adds `scriptName` to some functions defined inside sqf files.
Because of #1065 this will have no effect on released versions

### DIFF
--- a/addons/common/init_perFrameHandler.sqf
+++ b/addons/common/init_perFrameHandler.sqf
@@ -19,6 +19,7 @@ GVAR(waitUntilAndExecArray) = [];
 
 // per frame handler system
 [QFUNC(onFrame), {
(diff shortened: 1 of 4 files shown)


addons/common/init_perFrameHandler.sqf
```diff
@@ -19,6 +19,7 @@ GVAR(waitUntilAndExecArray) = [];
 
 // per frame handler system
 [QFUNC(onFrame), {
+    SCRIPT(onFrame);
```
dedmen: I have wished for this stuff for SOOOO long, to be able to see it in the debugger. Which it can't because there is no identifying factor in there.
Maybe just move it into a SQF file ^^

PR #1081, opened 2019-02-26
Activate all addons

PabstMirror:
On current master an old mission fails to load on a dedicated server with error:
```
 Warning Message: You cannot play/edit this mission; it is dependent on downloadable content that has been deleted.
ace_javelin
```
On a different mission, it seems like ace_javelin is not in activated addons
image

This partially reverts changes in #1051

### DIFF
--- a/addons/common/XEH_preStart.sqf
+++ b/addons/common/XEH_preStart.sqf
@@ -3,21 +3,4 @@
 //See usage in XEH_preInit
 private _cfgPatches = configFile >> "CfgPatches";
 private _allComponents = "true" configClasses _cfgPatches apply {configName _x};
-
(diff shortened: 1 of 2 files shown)


addons/common/XEH_preStart.sqf
```diff
@@ -3,21 +3,4 @@
 //See usage in XEH_preInit
 private _cfgPatches = configFile >> "CfgPatches";
 private _allComponents = "true" configClasses _cfgPatches apply {configName _x};
-
-//Filter out addons that don't have any units defined as we don't need to activate these
```
dedmen: This was the actual problem. We could still keep filtering out CfgAddons.

PR #1084, opened 2019-02-28
Preload 3DEN and Zeus item lists

commy2:
**When merged this pull request will:**
- Adds `cba_ui_fnc_preload3den` to generate the item list used as mission attribute in the editor.
- Adds `cba_ui_fnc_preloadCurator` to generate the complete item list for the `RscAttributeInventory` display used on weapon crates in Zeus.
- Adds `test_preload.sqf";` unit test to check validity of the preloaded item lists (compares them to the lists you get with vanilla or w/e other modset)

Note that the test requires a previewed mission in 3den for the curator part.

The functions to generate these lists are taken from vanilla, but optimized / ace-ified.

PR #1087, opened 2019-03-02
fix undefined flexi menu functions

commy2:
- Restore the old variable names before the flexi menu was moved to  CfgFunctions
- some optimization
- some formatting

This is needed for the TFAR radio menus, because they, for whatever reason, are actually based on the flexi menu of all things.

### DIFF
--- a/addons/ui/CfgFunctions.hpp
+++ b/addons/ui/CfgFunctions.hpp
@@ -19,9 +19,6 @@ class CfgFunctions {
             class flexiMenu_keyUp {
                 file = QUOTE(PATHTOF(flexiMenu\fnc_keyUp.sqf));
             };
-            class menu {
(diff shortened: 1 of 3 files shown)


addons/ui/XEH_preStart.sqf
```diff
@@ -1,7 +1,9 @@
 #include "script_component.hpp"
 
-PREP(initDisplayInterrupt);
-PREP(initDisplayMultiplayerSetup);
-PREP(initDisplayOptionsLayout);
-PREP(initDisplayPassword);
-PREP(initDisplayRemoteMissions);
+if (hasInterface) then {
```
Dystopian: conflict with #1084

commy2: Rebased it away.

PR #1091, opened 2019-03-08
Jam - fix p90 magazineWell and add test script

PabstMirror:


commy2:
```sqf
2 == getNumber
```
Why? ^^

commy2:
The test never reports anything if everything goes well, and doesn't show anything on screen if it failed, but I guess that is okay. It passed.

PR #1092, opened 2019-03-08
settings, fix set for out of bounds priority

commy2:
- fix a freeze for invalid priority used in FUNC(set).

```sqf
["banana", "CHECKBOX", "Banana", "Banana", true] call CBA_settings_fnc_init;
["banana", true, 3] call CBA_settings_fnc_set;
```
Before PR: game freeze, after PR: undefined behavior.


commy2:
That's good, because it is used like 7 times in the scripts, and they can all stay as they are.

PR #1094, opened 2019-03-09
fix error popup in Arsenal Garage

commy2:
- fixes error popup `Picture logo_cba_ca.paa not found` when opening the Virtual Garage
- remove a version number from a place that will not get updated
- attempt to hide the invisible target from the Arsenal using `scopeArsenal = 0;`

Predictably `scopeArsenal = 0;` doesn't work. It never works. I just leave it here as a reminder for everyone else that `scopeArsenal` indeed does not work.

### DIFF
--- a/addons/main_a3/CfgMods.hpp
+++ b/addons/main_a3/CfgMods.hpp
@@ -2,7 +2,7 @@ class CfgMods {
     class PREFIX {
         author = "$STR_CBA_Author";
         dir = "@CBA_A3";
-        name = "Community Base Addons";
+        name = "Community Base Addons v0.0.0";
         picture = "x\cba\addons\main\logo_cba_ca.paa";
         hidePicture = 1;
         hideName = 1;
--- a/addons/main_a3/CfgMods.hpp
+++ b/addons/main_a3/CfgMods.hpp
@@ -2,15 +2,13 @@ class CfgMods {
     class PREFIX {
         author = "$STR_CBA_Author";
         dir = "@CBA_A3";
-        name = "Community Base Addons v3.9.1";
+        name = "Community Base Addons";
         picture = "x\cba\addons\main\logo_cba_ca.paa";
         hidePicture = 1;
         hideName = 1;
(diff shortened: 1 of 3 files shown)


addons/main_a3/CfgMods.hpp
```diff
@@ -2,15 +2,13 @@ class CfgMods {
     class PREFIX {
         author = "$STR_CBA_Author";
         dir = "@CBA_A3";
-        name = "Community Base Addons v3.9.1";
```
PabstMirror: change it to 0.0.0
there is code to change on build
https://github.com/CBATeam/CBA_A3/blob/master/tools/make.py#L77
https://github.com/CBATeam/CBA_A3/blob/master/tools/make.py#L597

commy2:
```suggestion
        name = "Community Base Addons v0.0.0";
```
Like this?

PabstMirror: yeah, also need to change
https://github.com/CBATeam/CBA_A3/blob/master/tools/make.py#L77
versionFiles = ["mod.cpp","addons\main_a3\CfgMods.hpp"]

commy2:
@PabstMirror Merge?

addons/main_a3/CfgMods.hpp
```diff
@@ -2,15 +2,13 @@ class CfgMods {
...
+        name = "Community Base Addons v0.0.0";
         picture = "x\cba\addons\main\logo_cba_ca.paa";
         hidePicture = 1;
         hideName = 1;
         actionName = "Website";
         action = "$STR_CBA_URL";
         description = "Bugtracker: https://github.com/CBATeam/CBA_A3/issues<br/>Documentation: https://github.com/CBATeam/CBA_A3/wiki";
-        logo = "logo_cba_ca.paa";
```
jonpas: This works if you have the logo inside a PBO as well, not just root of `@mod` folder.

PR #1095, opened 2019-03-09
fix spacing for mods without description

commy2:
- There are two empty lines betweem mods with no description. This looks better.


dedmen:
>betweem mods

>betweem

:thinking:

PR #1122, opened 2019-04-22
fix CBA_fnc_compatibleMagazines skipping muzzle magazines

commy2:
**When merged this pull request will:**
- `CBA_fnc_compatibleMagazines` ignores sub muzzle magazines and magwells
- Therefore arifle_mx_F and arifle_mx_GL_F are claimed to use the same mags even though GL is compatible with all the under barrel grenade launcher rounds

Before:
```sqf
"arifle_MX_GL_F" call CBA_fnc_compatibleMagazines
//["30Rnd_65x39_caseless_mag","30Rnd_65x39_caseless_mag_Tracer","30Rnd_65x39_caseless_green","30Rnd_65x39_caseless_green_mag_Tracer"]

"Throw" call CBA_fnc_compatibleMagazines
[]
```

After:
```sqf
"arifle_MX_GL_F" call CBA_fnc_compatibleMagazines
//["30Rnd_65x39_caseless_mag","30Rnd_65x39_caseless_mag_Tracer","30Rnd_65x39_caseless_green","30Rnd_65x39_caseless_green_mag_Tracer","1Rnd_HE_Grenade_shell","UGL_FlareWhite_F","UGL_FlareGreen_F","UGL_FlareRed_F","UGL_FlareYellow_F","UGL_FlareCIR_F","1Rnd_Smoke_Grenade_shell","1Rnd_SmokeRed_Grenade_shell","1Rnd_SmokeGreen_Grenade_shell","1Rnd_SmokeYellow_Grenade_shell","1Rnd_SmokePurple_Grenade_shell","1Rnd_SmokeBlue_Grenade_shell","1Rnd_SmokeOrange_Grenade_shell","3Rnd_HE_Grenade_shell","3Rnd_UGL_FlareWhite_F","3Rnd_UGL_FlareGreen_F","3Rnd_UGL_FlareRed_F","3Rnd_UGL_FlareYellow_F","3Rnd_UGL_FlareCIR_F","3Rnd_Smoke_Grenade_shell","3Rnd_SmokeRed_Grenade_shell","3Rnd_SmokeGreen_Grenade_shell","3Rnd_SmokeYellow_Grenade_shell","3Rnd_SmokePurple_Grenade_shell","3Rnd_SmokeBlue_Grenade_shell","3Rnd_SmokeOrange_Grenade_shell"]

"Throw" call CBA_fnc_compatibleMagazines
[...]

PR #1123, opened 2019-04-22
compatibleMagazines - Add optional bool to get all muzzles

PabstMirror:


addons/common/XEH_preInit.sqf
```diff
@@ -71,4 +71,6 @@ activateAddons GVAR(addons);
     }];
 }] call CBA_fnc_addClassEventHandler;
 
+GVAR(magNamespace) = call CBA_fnc_createNamespace; // namespace used for CBA_fnc_compatibleMagazines
```
commy2: Now the function no longer works in preInit from components loaded before cba_common.

commy2: The caching at least, so I guess it's fine in a way?

addons/common/fnc_compatibleMagazines.sqf
```diff
@@ -22,56 +23,42 @@ Author:
...
 
-params [["_weapon", "", ["", configNull]]];
+params [["_weapon", "", ["", configNull]], ["_allMuzzles", false, [false]]];
 
 if (_weapon isEqualType "") then {
     _weapon = configFile >> "CfgWeapons" >> _weapon;
 };
 
-private _cacheKey = str _weapon;
+private _cacheKey = format ["%1#%2",_weapon,_allMuzzles];
```
commy2:
```suggestion

if (_cacheKey == "") exitWith {
    ERROR_1("Weapon Does Not Exist %1",_this);
    []
};
```

addons/common/XEH_preInit.sqf
commy2:
```suggestion
```

addons/common/XEH_preInit.sqf
```diff
@@ -71,4 +71,6 @@ activateAddons GVAR(addons);
...
+
```
commy2:
```suggestion
```

addons/common/fnc_compatibleMagazines.sqf
PabstMirror: cacheKey will never be "" because `format ["%1#%2",_weapon,_allMuzzles];`
this was only there for the getVar "" thing, not really needed

commy2:
ref #1122

PR #1124, opened 2019-04-23
Add Italian translation

alef:
- Add Italian translation for most CBA components (9 left out)

### DIFF
--- a/addons/strings/stringtable.xml
+++ b/addons/strings/stringtable.xml
@@ -31,7 +31,12 @@
         </Key>
         <Key ID="STR_CBA_FORMAT_NUMBER_THOUSANDS_SEPARATOR">
             <English>,</English>
+            <Czech> </Czech>
             <German>.</German>
+            <Polish> </Polish>
+            <Russian> </Russian>
+            <French> </French>
+            <Spanish> </Spanish>
             <Italian>,</Italian>
             <Japanese>,</Japanese>
             <Chinese>,</Chinese>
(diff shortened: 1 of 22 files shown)


addons/strings/stringtable.xml
```diff
@@ -30,13 +31,8 @@
         </Key>
         <Key ID="STR_CBA_FORMAT_NUMBER_THOUSANDS_SEPARATOR">
             <English>,</English>
-            <Czech> </Czech>
```
PabstMirror: don't remove these

veteran29: Does arma support non breaking space character? I wonder if that could be used here, it's really easy to mess that when using Tabler.

addons/vectors/stringtable.xml
```diff
@@ -1,4 +1,4 @@
-﻿<?xml version="1.0" encoding="utf-8"?>
+<?xml version="1.0" encoding="utf-8"?>
```
commy2: Which of these is right?

alef: It seems Tabler saves _stringtable.xml_ files without [BOM](https://en.wikipedia.org/wiki/Byte_order_mark). On your master branch, only vectors/ , versioning/ and xeh/ have BOM. All the others do not. I would assume that the one without BOM (my branch) are right.

addons/strings/stringtable.xml
bux: How would I differentiate between a user error and intentional blank space :/

addons/vectors/stringtable.xml
dedmen: I already mentioned somewhere that these 3 had invalid bom in the past. We also had a couple script files with invalid I think.

dedmen: On master branch backpacks module stringtable still has bom.

PR #1125, opened 2019-04-28
add setting to disable disposable launchers 

commy2:
- add setting to disable disposable launchers
- only if enabled (default) will disposable launcher be replaced with used tube after firing
- only if enabled will disposable launcher magazines in cargo be replaced with loaded disposable launchers
- if disabled, disposable launchers can be reloaded
- tooltips for auto-drop-used-launcher setting

### DIFF
--- a/addons/disposable/fnc_replaceMagazineCargo.sqf
+++ b/addons/disposable/fnc_replaceMagazineCargo.sqf
@@ -20,6 +20,8 @@ Author:
     commy2
 ---------------------------------------------------------------------------- */
 
+if (!GVAR(replaceDisposableLauncher)) exitWith {};
+
 params ["_box"];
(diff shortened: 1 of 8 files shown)


addons/disposable/fnc_replaceMagazineCargo.sqf
```diff
@@ -20,6 +20,8 @@ Author:
     commy2
 ---------------------------------------------------------------------------- */
 
+if (!GVAR(replaceDisposableLauncher)) exitWith {};
```
PabstMirror: This runs before `Settings Initialized`
might need to use
```
["CBA_settingsInitialized", {
    ["All", "Init", FUNC(replaceMagazineCargo), nil, nil, true] call CBA_fnc_addClassEventHandler;
```

PR #1126, opened 2019-04-29
Update Japanese translation

classicarma:
- Supported to new entries


commy2:
There are some more strings in this PR: https://github.com/CBATeam/CBA_A3/pull/1125/files

PR #1129, opened 2019-04-30
fix headDir 1.92

### LINKED ISSUE #1128: CBA_fnc_headDir broken in 1.92
**Arma 3 Version:** `1.92` (stable)
**CBA Version:** `3.10.1` (stable)

**Mods:**
```
- CBA_A3
- ACE3
```

**Description:**
-  CBA_fnc_headDir returns four distinct values rather than a smooth range of values as the player turns in a circle.

**Steps to reproduce:**
- Monitor the value of `[player] call CBA_fnc_headDir` as the player turns in a circle.

- SP/Local MP

**RPT log file:**
- N/A

Here's a video: https://streamable.com/h6a0l

It seems like this could potentially be related to the changes that BI made to position commands (like setPos, for example, `player setPos [6969696, 0, 0]` results in the player ending up at world position `[500000, 0, 0]` - they've capped a lot of things it seems), look at this line of the CBA function: https://github.com/CBATeam/CBA_A3/blob/master/addons/common/fnc_headDir.sqf#L68

EDIT - `AGLToASL positionCameraToWorld [0, 0, 99999999];` gets capped out at `[-500000,-500000,500000]`, `[500000,-500000,500000]`, `[500000,500000,500000]`, or `[-500000,500000,500000]` based on the direction you're facing. This is definitely the issue!

That's definitely a 🤔

commy2:
- fix #1128

### DIFF
--- a/addons/common/fnc_headDir.sqf
+++ b/addons/common/fnc_headDir.sqf
@@ -65,7 +65,7 @@ if (_unit != call CBA_fnc_currentUnit) then {
 } else {
     private _camPos = AGLToASL positionCameraToWorld [0, 0, 0];
     _isExternalCam = (_camPos vectorDistance getPosASL _unit) > 2;
-    private _viewPos = AGLToASL positionCameraToWorld [0, 0, 99999999];
+    private _viewPos = AGLToASL positionCameraToWorld [0, 0, 10000];
     private _vector = _viewPos vectorDiff _camPos;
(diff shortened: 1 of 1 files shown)


PabstMirror:
Might be able to simplify with https://community.bistudio.com/wiki/getCameraViewDirection in the future

commy2:
Everything with `vectorNormalized`-`vectorMultiply`-`vectorMagnitude` can be improved with https://community.bistudio.com/wiki/vectorLinearConversion

PR #1130, opened 2019-04-30
Change getVolume to use visual clipping type

Wakbub:
- Change [getVolume](https://cbateam.github.io/CBA_A3/docs/files/common/fnc_getVolume-sqf.html) to use visual instead of general clipping type for [boundingBoxReal](https://community.bistudio.com/wiki/boundingBoxReal).

Arma 3's [1.92 update](https://dev.arma3.com/post/spotrep-00088) changed [boundingBoxReal](https://community.bistudio.com/wiki/boundingBoxReal) so that it can be passed a number for clipping type. Visual clipping type (0) gives more precise measurements when estimating the volume of an object compared to general clipping type (3).

Real bounding boxes with visual and general clipping types are shown below for certain objects (a vehicle, an ammo box, and a building) where the differences are profound. Some objects have almost identical real bounding boxes despite different clipping types (e.g., a Strider has less than 1% difference in volume). Real bounding boxes with visual clipping type are colored in white and those with general clipping type are colored in black.

Vehicle
_Figure 1_. Real bounding box of a vehicle (I_Truck_02_fuel_F). Visual clipping type: 105.406 m³. General clipping type: 280.191 m³. Difference: ≈38%.

Ammo box
_Figure 2_. Real bounding of an ammo box (Box_T_East_Ammo_F). Visual clipping type: 0.171219 m³. General clipping type: 1.37109 m³. Difference: ≈12%.

Building
_Figure 3_. Real bounding box of a building (Land_cargo_house_slum_F). Visual clipping type: 69.1653 m³. General clipping type: 174.096 m³. Difference: ≈40%.

There exists two other clipping types: shadow (1) and geometry (2). For the tested objects there were no differences in volume between shadow, geometry, and general clipping types. Regarding performance, there was no noticeable difference between any of the four clipping types. It thus seems motivated to use visual.

### DIFF
--- a/addons/common/fnc_getVolume.sqf
+++ b/addons/common/fnc_getVolume.sqf
@@ -4,7 +4,8 @@ Function: CBA_fnc_getVolume
 
 Description:
     Return the volume of the bounding box of an object's model.
-    The bounding box is retrieved using boundingBoxReal instead of boundingBox for more precise measurements.
+    The bounding box is retrieved using boundingBoxReal instead of boundingBox
+    with visual clipping type for more precise measurements.
 
 Parameters:
     _object - Object to calculate the volume of <OBJECT>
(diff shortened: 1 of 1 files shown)


commy2:
This is cool. Do I understand correctly that the old unary command and the new binary command with left hand set argument 3 report the same result?

dedmen:
>Do I understand correctly that the old unary command and the new binary command with left hand side argument 3 report the same result?

Yes. We talked about implementing this command a year ago when I wrote it and sent it to BI and they didn't take it ^^

commy2:
Thanks.

PR #1131, opened 2019-04-30
JR - Fix suppressor file path changes in 1.92

PabstMirror:
pictures changed in 1.92
it's defined fine in vanilla configs

### DIFF
--- a/addons/jr/cfgweapons.hpp
+++ b/addons/jr/cfgweapons.hpp
@@ -920,17 +920,14 @@ class CfgWeapons {
     class muzzle_snds_H_MG: muzzle_snds_H { // we were seriously lacking a MG suppressor since BI hid these (1.64)
         scope = 2;
         displayName = "$STR_CBA_JR_cfgweapons_muzzle_snds_h_mg"; // 7.62 MG, can also be used on 6.5 and 5.56 MGs
-        picture = "\a3\weapons_f_mark\Data\UI\gear_acca_snds_338_tan_CA.paa";
         model = "\a3\weapons_f\acc\acca_snds_338_tan_F";
     };
     class muzzle_snds_H_MG_blk_F: muzzle_snds_H_MG {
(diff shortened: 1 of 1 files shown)


dedmen:
Now our black and green silencers have the same picture? That can't be right.

PabstMirror:
vanilla configs are
```
class muzzle_snds_H_MG_blk_F: muzzle_snds_H_MG {
    picture = "\a3\Weapons_F_Exp\Acc\Data\UI\icon_muzzle_snds_H_MG_blk_F_ca.paa";
class muzzle_snds_H_MG_khk_F: muzzle_snds_H_MG {
    picture = "\a3\Weapons_F_Exp\Acc\Data\UI\icon_muzzle_snds_H_MG_khk_F_ca.paa";
```
and those are different images

dedmen:
Ah we edit, not create that class. 
Why do we change the `model`? atleast for `muzzle_snds_H_MG_blk_F` we have the same as vanilla.
Also the vanilla name seems to be "Sound suppressor MG" which is same as ours.
What do we even do here? 
We don't change their model, we don't change the name, we don't change the picture.

commy2:
Good questions. I think we may have once added these unused textures and names or something, but now BI does use them themselves. Should probably all be deleted.

PabstMirror:
vanilla models are different (I think they were either shorter or longer) but looked fine to me

robalo:
When I added the MG suppressor classes to CBA the vanilla ones were hidden and used either the 7.62 SCAR model or the 5.56 YHM one. IMO using the .338 model looks a lot more like an MG suppressor than those.

robalo:
> 
> 
> Good questions. I think we may have once added these unused textures and names or something, but now BI does use them themselves. Should probably all be deleted.

Nope, they're still hidden, the only issue was picture path has changed in 1.92 for the .338 suppressors which we use as MG suppressors.

commy2:
So everything is alright and works as expected here?

PR #1134, opened 2019-05-03
fix fail to read settings from settings file

commy2:
- title
- bug is old, but only circumstances reveal it in 3.11.0 with CBA only
- Thanks to cyruz, diwako, max255pl and whoever I forgot for helping with debugging and thanks for reporting ofc :P

### DIFF
--- a/addons/settings/fnc_addSetting.sqf
+++ b/addons/settings/fnc_addSetting.sqf
@@ -45,4 +45,8 @@ Author:
     commy2
 ---------------------------------------------------------------------------- */
 
-call (uiNamespace getVariable QFUNC(init)) == 0
+if (isNil QFUNC(init)) then {
+    #include "XEH_PREP.sqf"
+};
+
+call FUNC(init) == 0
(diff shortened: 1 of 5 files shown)


diwako:
Tested on dedicated linux server. Server enforced settings for ACE and ACRE2 during the test as expected from the cba_settings.sqf file in a PBO loaded on server.

addons/settings/fnc_addSetting.sqf
```diff
@@ -45,4 +45,8 @@ Author:
     commy2
 ---------------------------------------------------------------------------- */
 
-call (uiNamespace getVariable QFUNC(init)) == 0
+if (isNil QFUNC(init)) then {
```
jonpas: Comment why?

commy2: If this one function undefined (implies all others are too) then define all functions.
It's basically an include guard.

jonpas: Not comment for me (but probably also for me in a few months).

PR #1138, opened 2019-05-04
add buildingPos dummy object

### LINKED ISSUE #1136: Donation: Custom Building Position
So I made this little custom building position, which is basically a small invisible object that can be placed down in the editor with one building position configured at its center.

-What does it do? It adds a small invisible disk object (much like a tiny invisible helipad) that has 1 building position at its center.

-How is this helpful? Well, it works together with scripts that utilize building positions. Its most obvious use would be extending the functionality of garrison and loot scripts. In those cases it would allow for additional building positions to be "added" to buildings that already have some preconfigured, or "adding" building positions to structures that don't have any (such as the bunker tower, used in the sample screenshots).

Sample Screenshots: (note that the disk is still visible in these screenshots as I hadn't configured the invisible texture at that point yet)

https://i.imgur.com/poaiquU.jpg
https://i.imgur.com/Sv51wP0.jpg

Source files can be found here: 

https://github.com/G3rrus/Custom-Building-Position

commy2:
- close #1136


commy2:
- basically an invisible micro building with a single AI building position
- for use in the editor and with scripts (`buildingPos` command)

<details><summary>dummy object snaps to buildings</summary>



</details>

<details><summary>dummy object can be used together with the GARRISON waypoint to create a formation</summary>




</details>

- Class is grouped in editor VR Objects / Helpers because that made the most sense to me and I don't want to create a new category for a single item.



</details>

PabstMirror:
might want to double check that p3d
pboProject build time goes from 8 to 28 seconds (cleaned p:\temp each time)
seems to be animation keyframes on LOD 0?

commy2:
Can you see these in Oxygen? I am not huge into that program.

PabstMirror:
Untitled

commy2:
Animation should be deleted now. Model size went from 27kB to 6kB.

commy2:
- added the helper as well as the target dummies to Zeus.

PabstMirror:
Still taking 28 seconds
changed texture from `a3\weapons_f\data\clear_empty.paa` to `#(argb,8,8,3)color(0,0,0,0,co)`
and it's back to 8?
might just be pboProject being weird, but I also don't see any harm in just using that color as a texture

commy2:
If that works then fine. Should be identical. Prodedural texture seems slightly preferable even, because you don't rely on external files.

Push the change or let me do it on the weekend?

commy2:
I added `CBA_fnc_buildingPositions`, which is exactly the same as `BIS_fnc_buildingPositions` and `buildingPos -1`, except that it also reports all custom building positions (`CBA_buildingPos` objects) inside the ClipVisual bounding box real of the object and now my head hurts due to the math.

```sqf
bunker call BIS_fnc_buildingPositions
//[]

bunker call CBA_fnc_buildingPositions
//[[3706.63,349.067,2.68139]]
```

commy2:
Function approved?

PR #1141, opened 2019-05-10
add manual reticle night switch keybind

commy2:
- default bound to Numpad 0
- requires `CBA_ScriptedOptic/manualReticleNightSwitch` to be set to 1. Otherwise still only automatic switch on ambient darkness


commy2:
Pls don't merge before #1142 is merged into this one. Don't want to deal with merge conflicts even though I am the only person working on this atm.

PR #1146, opened 2019-05-15
Add more params to notify function

Dubjunk:
- title
- Added queue and lifetime params

I've noticed that the queue prevent to display "urgent notifications" which you can't avoid.
With the provided lifetime and `queue` parameter you're able to clear the queue and therefore display directly an urgent notification.

### DIFF
--- a/addons/ui/fnc_notify.sqf
+++ b/addons/ui/fnc_notify.sqf
@@ -35,7 +35,6 @@ params [
     ["_lifetime", 4, [0]]
 ];
 
-#define LIFE_TIME _lifetime
 #define FADE_IN_TIME 0.2
 #define FADE_OUT_TIME 1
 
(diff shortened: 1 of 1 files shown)


jonpas:
I am not sure I like this. Priority notifications yes, but this will just make everyone force show. We should clearly specify to use this rarely or something like that.

commy2:
I noticed the same problem.

Is the function still backwards compatible with the changed params?

My idea was to have a priority value (NUMBER) for each notification and that changes the position in queue including potentially changing the displayed notification (maybe pusing the previous one to be displayed again down the queue).

Dubjunk:
It's not backwards compatible.
A priority value would be nice too but in my case I don't want to show the "old" notify.

Maybe change the queue param to `Bool, Number`
On true -> normal queue
On false -> force show and empty the queue
Number -> add a priority to it

TheMagnetar:
If it's not backward compatible make it a new function.

commy2:
>Maybe change the queue param to Bool, Number

Sounds good, except that I would make those two different params (arguments).

jonpas:
I would prefer "don't stop queue" parameter instead of "force show", meaning if another notification comes in, it instantly overwrites it. Also solves the problem of mods forcing away important notifications with less important ones, but at the same time hides specifically marked less important ones right away.

I like lifetime.

jonpas:
Replaced by #1175.

PR #1150, opened 2019-05-24
Replace %20 with space when displaying mission name

commy2:
- Replaces %20 in the mission name with spaces on mission select screen, lobby, briefing and map.

### DIFF
--- a/addons/ui/fnc_initDisplayRemoteMissions.sqf
+++ b/addons/ui/fnc_initDisplayRemoteMissions.sqf
@@ -89,8 +89,10 @@ private _fnc_storeMapMissions = {_this spawn {isNil { // delay a frame
     private _ctrlMissions = _display displayCtrl IDC_SERVER_MISSION;
 
     private _missions = [];
+    private _fnc_replace = uiNamespace getVariable "CBA_fnc_replace";
+
     for "_i" from 0 to (lbSize _ctrlMissions - 1) do {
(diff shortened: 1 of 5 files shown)


addons/ui/fnc_initDisplayRemoteMissions.sqf
```diff
@@ -89,8 +89,10 @@ private _fnc_storeMapMissions = {_this spawn {isNil { // delay a frame
     private _ctrlMissions = _display displayCtrl IDC_SERVER_MISSION;
 
     private _missions = [];
+    private _fnc_replace = uiNamespace getVariable "CBA_fnc_replace";
+
     for "_i" from 0 to (lbSize _ctrlMissions - 1) do {
-        private _name = _ctrlMissions lbText _i;
+        private _name = [_ctrlMissions lbText _i, "%20", " "] call _fnc_replace; // replace %20 with space
```
Dystopian: looks like `_fnc_replace` is bad name

commy2: How so? It is `CBA_fnc_replace` fetched from ui namespace in case the host started the game with -world=empty.

Dystopian: oh sorry I just missed assigning line :confused: :man_facepalming:

commy2:
https://www.w3schools.com/tags/ref_urlencode.asp

PR #1151, opened 2019-05-24
add CBA_fnc_randomVector3D

commy2:
- adds a function to generate a random 3 dimensional unit vector with uniform spherical distribution
- I see a lot of people struggling with this, implementing non uniform distributions and it bothers me...
- used in ACE in multiple places now (as macro) i.e. https://github.com/acemod/ACE3/blob/master/addons/hot/script_component.hpp#L19-L24

### DIFF
--- a/addons/vectors/CfgFunctions.hpp
+++ b/addons/vectors/CfgFunctions.hpp
@@ -22,6 +22,7 @@ class CfgFunctions {
             PATHTO_FNC(vectRotate2D);
             PATHTO_FNC(vectRotate3D);
             PATHTO_FNC(vectSubtract);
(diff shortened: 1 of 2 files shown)


PabstMirror:
distribution doesn't look right
image

commy2:
`vectorNormalized [-1 + random 2,-1 + random 2,-1 + random 2]`
This is not uniformly distributed. It will bias towards where the 8 edges are of the cube.

You would have to discard all results that lie outside the sphere first (vecMag > 1) and reroll in that case before normalizing.

The problem with my math was that elevation went from 0° to 180° instead of -90° to 90°. I fixed it. Please test with that script of yours.

PR #1154, opened 2019-06-02
fix updating flashlight fx after attachment switch

tbeswick96:
- Move adding of next attachment to next frame
- Move switched localEvent to next frame (after add**Item)
- Allows different flashlight modes to be used and update properly when switching
- Currently switching from one flashlight to another in the same frame whilst the flashlight is on causes the light effect not to update to the new flashlight

### DIFF
--- a/addons/accessory/fnc_switchAttachment.sqf
+++ b/addons/accessory/fnc_switchAttachment.sqf
@@ -56,15 +56,27 @@ if (!isNil "_switchItem") then {
     switch (_currWeaponType) do {
         case 0: {
             _unit removePrimaryWeaponItem _currItem;
-            [{_this#0 addPrimaryWeaponItem _this#1}, [_unit, _switchItem]] call CBA_fnc_execNextFrame;
+            [{
+                params ["_unit", "", "_switchItem"];
+                _unit addPrimaryWeaponItem _switchItem;
+                ["CBA_attachmentSwitched", _this] call CBA_fnc_localEvent;
+            }, [_unit, _currItem, _switchItem, _currWeaponType]] call CBA_fnc_execNextFrame;
         };
         case 1: {
             _unit removeHandgunItem _currItem;
-            [{_this#0 addHandgunItem _this#1}, [_unit, _switchItem]] call CBA_fnc_execNextFrame;
+            [{
+                params ["_unit", "", "_switchItem"];
+                _unit addHandgunItem _switchItem;
+                ["CBA_attachmentSwitched", _this] call CBA_fnc_localEvent;
+            }, [_unit, _currItem, _switchItem, _currWeaponType]] call CBA_fnc_execNextFrame;
         };
         case 2: {
             _unit removeSecondaryWeaponItem _currItem;
-            [{_this#0 addSecondaryWeaponItem _this#1}, [_unit, _switchItem]] call CBA_fnc_execNextFrame;
+            [{
+                params ["_unit", "", "_switchItem"];
+                _unit addSecondaryWeaponItem _switchItem;
+                ["CBA_attachmentSwitched", _this] call CBA_fnc_localEvent;
+            }, [_unit, _currItem, _switchItem, _currWeaponType]] call CBA_fnc_execNextFrame;
         };
     };
     private _switchItemHintText = getText (__cfgWeapons >> _switchItem >> "MRT_SwitchItemHintText");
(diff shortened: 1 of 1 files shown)


tbeswick96:
Video to show:
https://www.youtube.com/watch?v=AgkcpN-At0Q

Shows normal flashlight on, switching to far causes no effect change.
Switching to far then off then on causes effect change.

ExecNextFrame forces the effect to refresh

dedmen:
>switching from one flashlight to another in the same frame whilst the flashlight is on causes the light effect not to update to the new flashlight

ugh.. Arma.

Near/Far flashlight is a nice idea, is your mod for that public or do you play to propose that to ACE?

tbeswick96:
Wasn't planning on making the far flashlights part of ace, I lumped them onto a 3CB flashlight/ir combo

Any thoughts on the localEvent?

dedmen:
>Any thoughts on the localEvent?

Oops missed that. I think it should. People may use it to re-replace the item or detach it or whatever. Would be weird if it's not there yet.

tbeswick96:
@dedmen Do you think this is ok? https://github.com/CBATeam/CBA_A3/pull/1154/files#diff-73655685f2cbb36065c6caf471ec6630R75
Should run after the adds as it's added after right?

dedmen:
Also the event says "switched" in past tense. So it has to fire after.
Yes, nextFrame EH's run in order they were added.

commy2:
>Currently switching from *one flashlight to another in the same frame whilst the flashlight is on* causes the light effect not to update to the new flashlight

I don't get it. The same frame as what else exactly? The flashlight being on? Is the flashlight being turned off anywhere here?

tbeswick96:
Switching in this context means removing one attachment and adding another. When both of the attachments are flashlights, the light effect of the new flashlight does not update unless you manually turn the light on and off after switching. This is because the remove and add are done in the same frame

addons/accessory/fnc_switchAttachment.sqf
```diff
@@ -56,23 +56,23 @@ if (!isNil "_switchItem") then {
     switch (_currWeaponType) do {
         case 0: {
             _unit removePrimaryWeaponItem _currItem;
-            _unit addPrimaryWeaponItem _switchItem;
+            [{_this#0 addPrimaryWeaponItem _this#1}, [_unit, _switchItem]] call CBA_fnc_execNextFrame;
```
commy2:
```suggestion
            [{(_this select 0) addPrimaryWeaponItem (_this select 1)}, [_unit, _switchItem]] call CBA_fnc_execNextFrame;
```

addons/accessory/fnc_switchAttachment.sqf
```diff
@@ -56,23 +56,23 @@ if (!isNil "_switchItem") then {
...
         };
         case 1: {
             _unit removeHandgunItem _currItem;
-            _unit addHandgunItem _switchItem;
+            [{_this#0 addHandgunItem _this#1}, [_unit, _switchItem]] call CBA_fnc_execNextFrame;
```
commy2:
```suggestion
            [{(_this select 0) addHandgunItem (_this select 1)}, [_unit, _switchItem]] call CBA_fnc_execNextFrame;
```

addons/accessory/fnc_switchAttachment.sqf
```diff
@@ -56,23 +56,23 @@ if (!isNil "_switchItem") then {
...
         };
         case 2: {
             _unit removeSecondaryWeaponItem _currItem;
-            _unit addSecondaryWeaponItem _switchItem;
+            [{_this#0 addSecondaryWeaponItem _this#1}, [_unit, _switchItem]] call CBA_fnc_execNextFrame;
```
commy2:
```suggestion
            [{(_this select 0) addSecondaryWeaponItem (_this select 1)}, [_unit, _switchItem]] call CBA_fnc_execNextFrame;
```

addons/accessory/fnc_switchAttachment.sqf
```diff
@@ -56,23 +56,23 @@ if (!isNil "_switchItem") then {
...
         };
     };
     private _switchItemHintText = getText (__cfgWeapons >> _switchItem >> "MRT_SwitchItemHintText");
     if !(_switchItemHintText isEqualTo "") then {
         hintSilent format ["%1", _switchItemHintText];
     };
     playSound "click";
-    ["CBA_attachmentSwitched", [_unit, _currItem, _switchItem, _currWeaponType]] call CBA_fnc_localEvent;
+    [{["CBA_attachmentSwitched", _this] call CBA_fnc_localEvent}, [_unit, _currItem, _switchItem, _currWeaponType]] call CBA_fnc_execNextFrame;
```
commy2:
```suggestion
```

commy2:
What happens if you hold down the key with this change? (switch attachment every frame. multiple times per frame)?

Should be possible I guess, because the current attachment is removed immediately, but would test anyway.

tbeswick96:
Holding the key down just switches once

PR #1155, opened 2019-06-03
Extend 3den selectable date range from 1900 to 2050

Drofseh:
- This will allow missions makers to set a date in 3den from 1900 to 2050.
By default 3den only allows a selectable date range from 1982 to 2050
- This will significantly improve the QOL for those making mission in the WW2, Vietnam, or other non-modern eras.


addons/common/config.cpp
```diff
@@ -14,6 +14,8 @@ class CfgPatches {
     };
 };
 
+class ctrlCombo;
```
commy2: Should be moved into `Cfg3DEN.hpp`. This class will not be used by any other header file.

addons/common/Cfg3DEN.hpp
```diff
@@ -6,6 +6,20 @@
...
     class Attributes {
+
+        class Default;
+        class Title: Default {
+            class Controls;
+        };
+
+        class Date: Title {
+            class Controls: Controls {
+                class ValueYear : ctrlCombo {
```
commy2:
```suggestion
                class ValueYear: ctrlCombo {
```

addons/common/Cfg3DEN.hpp
```diff
@@ -6,6 +6,20 @@
...
+                    onLoad="_ctrlYear = _this select 0;    for '_y' from 1900 to 2050 do { _lbadd = _ctrlYear lbadd str _y; _ctrlYear lbsetvalue [_lbadd,_y]; };    _ctrlYear lbsetcursel 53;";
```
commy2:
```suggestion
                    onLoad = "\
                        _ctrlYear = _this select 0;\
                        for '_y' from 1900 to 2050 do {\
                            _lbadd = _ctrlYear lbadd str _y;\
                            _ctrlYear lbsetvalue [_lbadd,_y];\
                        };\
                        _ctrlYear lbsetcursel 53;\
                    ";
```

You already sort of fixed the formatting by removing the left hand side white space from the expression. May as well format it like this to make it readable.

Drofseh:
Thanks for the suggestions @commy2, I'll implement them.

Would `"A3_Data_F_Exp_B_Loadorder"` be good to add to `requiredAddons[]`?
It in turn requires `"A3_3DEN","A3_3DEN_Language","A3_Data_F_Exp_B","A3_Language_F_Exp_B"`

commy2:
common is already requiring Loadorder Tanks, which requires Loarorder ExpB iirc.

Drofseh:
`A3_3DEN` is the one that's really required for this

commy2:
Yes, and LO ExpB requires A3_3DEN, which means that Tanks requires 3DEN implicitly as well already, and so does common.

addons/common/config.cpp
Drofseh: done

addons/common/Cfg3DEN.hpp
Drofseh: done

addons/common/Cfg3DEN.hpp
Drofseh: done

Drofseh:
Alright, I'm happy with it if you are.

commy2:
This change should go along with adding the years to ModuleDate_F.
However, adding values to arguments in modules out of order may break modules placed without CBA / before CBA, because the value is referenced by config index and if you set 1982 in vanilla (index 0) and then load this patch, it might become 1900, because that is the new first value with index 0.

... iirc

Drofseh:
Yah, updating ModuleDate_F looks like it will break things to me.

<details>
<summary>Details</summary>

```
	class ModuleDate_F: Module_F
	{
		author="$STR_A3_Bohemia_Interactive";
		_generalMacro="ModuleDate_F";
		scope=2;
		displayName="$STR_A3_CfgVehicles_ModuleDate_F_0";
		category="Events";
		function="BIS_fnc_moduleDate";
		isGlobal=1;
		isTriggerActivated=1;
		class Arguments
		{
			class Year
			{
				displayName="$STR_A3_CfgVehicles_ModuleDate_F_Arguments_Year_0";
				description="";
				class values
				{
					class NoChange
					{
						name="$STR_DISP_OPT_NOCHANGE";
						value=-1;
						default=1;
					};
					class Number1982
					{
						name="1982";
						value=1982;
					};
					class Number1983
					{
						name="1983";
						value=1983;
					};
					class Number1984
					{
						name="1984";
						value=1984;
					};
					//etc.
```

</details>

commy2:
But changing only one seems awkward to me :S

jonpas:
Eden should be used instead of modules anyways. If changing the module breaks something, leave it out.

addons/common/CfgVehicles.hpp
```diff
@@ -5,4 +5,97 @@ class CfgVehicles {
...
+                    YEAR(1972);
+                    YEAR(1973);
+                    YEAR(1974);
+                    YEAR(1975);
+                    YEAR(1976);
+                    YEAR(1977);
+                    YEAR(1978);
+                    YEAR(1979);
+                    YEAR(1980);
+                    YEAR(1981);
```
commy2:
```suggestion
                    YEAR(1981);
                    class Number1982 {};
```
This should fix ordering.

Drofseh: It does not appear to do so

addons/common/CfgVehicles.hpp
```diff
@@ -5,4 +5,97 @@ class CfgVehicles {
         scope = 1;
         displayName = "";
     };
+
+    class Module_F;
+
```
commy2:
```suggestion
```

addons/common/Cfg3DEN.hpp
```diff
@@ -4,8 +4,30 @@
 #define POS_H_TITLE(N) (N * 3.5 + 1 * 5) * (pixelH * pixelGrid * 0.50)
 #define POS_H_VALUE(N) N * 3.5 * (pixelH * pixelGrid * 0.50)
 
+class ctrlCombo;
+
 class Cfg3DEN {
     class Attributes {
+
```
commy2:
```suggestion
```

PR #1157, opened 2019-06-07
update CBA_fnc_uniqueUnitItems

### LINKED ISSUE #1156: magazines not attached to weapons in use are omitted from CBA_fnc_uniqueUnitItems
**Arma 3 Version:** `1.92` (stable)
**CBA Version:** `3.11.2` (stable)

**Mods:**
```
- CBA_A3
```

**Description:**
- `CBA_fnc_uniqueUnitItems` does not return any magazines not currently attached to a weapon in use by the unit. It seems that `getItemCargo` doesn't return them while `magazines` does.

**Steps to reproduce:**
- Place down ammo carrier, drop weapons and observe result of `CBA_fnc_uniqueUnitItems`

- Dedicated / Editor (Singleplayer) / Editor (Multiplayer)

**RPT log file:**
- https://gist.github.com/zharf/208e4ab33df63c9f0283a0e8264d1df2

commy2:
- close #1156
- added `_magazines` flag, default false. Unless set to true, no mags will be reported (even with `_weaponItems` set to true)
- `_magazines` flag will not report *loaded* mags unless `_weaponItems` flag is also set to true.
- added note why `binocular` command is not needed (covered by `assignedItems`)
- no longer reports empty string
- formatting

### DIFF
--- a/addons/common/fnc_uniqueUnitItems.sqf
+++ b/addons/common/fnc_uniqueUnitItems.sqf
@@ -54,8 +54,7 @@ if (_weaponItems) then {
     _allItems append [
         primaryWeapon _unit,
         secondaryWeapon _unit,
-        handgunWeapon _unit,
-        binocular _unit
+        handgunWeapon _unit // binocular covered by assignedItems
     ];
 
     _allItems append primaryWeaponItems _unit;
--- a/addons/common/fnc_uniqueUnitItems.sqf
+++ b/addons/common/fnc_uniqueUnitItems.sqf
@@ -8,10 +8,11 @@ Description:
 Parameters:
     _unit           - Unit to retrieve the items from
     _weaponItems    - Include weapons, attachments, loaded magazines (Default: false)
-    _backpack       - Include items in backpack (Default: true)
-    _vest           - Include items in vest     (Default: true)
-    _uniform        - Include items in uniform  (Default: true)
-    _assignedItems  - Include assigned items    (Default: true)
+    _backpack       - Include items in backpack    (Default: true)
+    _vest           - Include items in vest        (Default: true)
+    _uniform        - Include items in uniform     (Default: true)
+    _assignedItems  - Include assigned items       (Default: true)
+    _magazines      - Include not loaded magazines (Default: false)
 
 Example:
     (begin example)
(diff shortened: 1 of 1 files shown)


addons/common/fnc_uniqueUnitItems.sqf
```diff
@@ -22,28 +23,66 @@ Returns:
...
+    _allItems append handgunItems _unit;
+
+    if (_magazines) then {
+        _allItems append primaryWeaponMagazine _unit;
+        _allItems append secondaryWeaponMagazine _unit;
+        _allItems append handgunMagazine _unit;
+        _allItems pushBack (_unit call CBA_fnc_binocularMagazine);
+    };
+
+    _allItems = _allItems - [""];
```
dedmen: Might be cheaper to do it at the end, after duplicate removal.
because `-` iterates over everything and copies everything.

dedmen: Or if we know where it comes from (I guess primary/secondary/handgunWeapon) we can do it right there.
Would be cheaper to do in L58 for just the 3 things, than do it for everything.

commy2: It comes from `p/s/hWeapon` as well as `p/s/hWeaponItems`. The magazine commands report empty arrays if no mag is loaded. I put the array subtraction below the last line that may add empty strings.

The placement is so that the array subtraction command is only executed when the `_weaponItems` flag is set, as this is the only place where the commands may add empty strings. Note that while `p/s/hWeaponItems` has the items in order muzzle/pointer/optics/bipod and fills the gaps with empty strings, assignedItems, while having an order, just changes in array size (and indices) if a specific slot is unoccupied.

If I were to add it to the last line after the `arrayIntersect` command instead, I would slow down the function for every use even when this flag is set to false. I know that a smaller array would speed up the function slightly, but this would mean adding an additional binary command (`-`) for all uses for a small improvement in C++ that itself wouldn't even be a problem in all other use cases. The priority is clear to me.

The only potential optimization here would be moving the weapon(items/magazines) block above the uniform/vest/backpack block.

commy2: Another potential optimization approach would be creating the initial array as `[""]` and then using `pushBackUnique` everywhere. And then using `deleteAt 0` at the end. However, this would need `appendUnique`, because one source of empty strings are arrays here.

dedmen: Maybe instead store _weaponItems seperately, then remove "" from it.
And only then append it to _allItems. That way you don't have to iterate over uniform/vest/backpack items to remove ""

dedmen: Make a new array here
https://github.com/CBATeam/CBA_A3/pull/1157/files#diff-5281788de7c6251ec13f252782a5e72fR54
And then `_allItems append _newArray` after removing the ""

commy2: That would mean more commands in total. I don't think there will be any case where this actually would be faster.

dedmen: but only in the weaponsItems case
which I'd say is rare.

PR #1163, opened 2019-06-14
update scripted optics effect radblur classes

commy2:
- title

CBA_OpticsRadBlur1 weak
CBA_OpticsRadBlur2 medium
CBA_OpticsRadBlur3 strong

Radius of effect so it matches standard optic size.

Names are taken from OpticsBlur1, OpticsBlur2, OpticsBlur3 etc, which follow the same weak/medium/strong pattern.


commy2:
```sqf
apply {
    if (_x isEqualType "") then {
        call compile _x
    } else {_x}
};
```

PR #1165, opened 2019-06-17
handle expressions in CfgOpticsEffect

commy2:
- Convert strings to numbers, so the command doesn't error.

Found this yesterday when testing CBA master with my unit. Related to #1163.


commy2:
- [ ] Test this on the 5:4 TFT to see if the 16/9 in height can be replaced by getResolution.

PR #1166, opened 2019-06-19
replace more %xy in mission names

commy2:
- adds `CBA_fnc_decodeHTML`
- function replaces more %xy than %20 / space

- [x] Is there a difference between HTML and HTML URL? I am not sure about the function name.
- [x] Function doesn't work as is, because some functions are not defined in mission namespace in the briefing. Needs some fixing.
- [x] Will we do the complete HTML character list?


bux:
An extension would be easier.
https://docs.microsoft.com/en-us/dotnet/api/system.net.webutility.htmldecode?view=netframework-4.8#System_Net_WebUtility_HtmlDecode_System_String_

veteran29:
`Is there a difference between HTML and HTML URL? I am not sure about the function name.`
IMO it would fit more to name this `fnc_urlDecode` / `fnc_decodeURL` you don't use percent encoding in HTML, you use this in URLs.

commy2:
No DLL's in CBA, bux.

addons/strings/script_component.hpp
```diff
@@ -10,3 +10,222 @@
 #endif
 
 #include "\x\cba\addons\main\script_macros.hpp"
+
+#define UTF8_TABLE [\
+    ["%21","!"],\
```
dedmen: These are just ascii keycodes
https://community.bistudio.com/wiki/toString

commy2: You have any good way to translate i.e. `%e2%80%98"` using `toString`?

commy2: Tested this:
```
missionName // "%c2%ac%c2%ae%c2%af%c2%b0%c2%b1%c2%b2%c2%b3%c2%b4%c2%b5%c2%b6%c2%b7%c2%b8%c2%b9%c2%ba%c2%bb%c2%bc%c2%bd%c2%be%c2%bf"

missionName call CBA_fnc_decodeURL // "¬®¯°±²³´µ¶·¸¹º»¼½¾¿"

toString (missionName splitString "%" apply {call compile format ["0x%1", _x]} apply { _x}) // "Â¬Â®Â¯Â°Â±Â²Â³Â´ÂµÂ¶Â·Â¸Â¹ÂºÂ»Â¼Â½Â¾Â¿"
```

So it is not exactly the same as ASCII codes. It seems to follow the UTF-8 conversion table instead:
https://www.w3schools.com/tags/ref_urlencode.asp
Except that % is encoded as %% and not %25.

dedmen:
Just noticed switching around terrains in the mission list visibly freezes my game.
Booted up the profiler and checked, its cuz fnc_decodeURL is being called 24 times with 20ms runtime each :thinking: 
each of them calling fnc_replace 217 times.
Even though a total of 0 missions on that terrain contain a % character in the name.. we just use underscores.

PR #1171, opened 2019-06-26
Update loadorder for 1.94 Contact DLC

dedmen:
Update our requiredAddons to include all of Contact DLC.
Don't merge before 1.94.

### DIFF
--- a/addons/common/config.cpp
+++ b/addons/common/config.cpp
@@ -8,7 +8,7 @@ class CfgPatches {
         units[] = {};
         weapons[] = {};
         requiredVersion = REQUIRED_VERSION;
-        requiredAddons[] = {"A3_Data_F_Tank_Loadorder"};
+        requiredAddons[] = {"A3_Data_F_Enoch_Loadorder"};
         version = VERSION;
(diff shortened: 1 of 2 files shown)


commy2:
tf is an "Enoch"?

jonpas:
> tf is an "Enoch"?

Codename for Livonia terrain.

ColdEvul:
Whould this effect load order of cba_settings in anyway?
I've starting this experiencing cba_settings related issues in the init.sqf in Enoch.
It complaining about variable not existing suddenly.

PabstMirror:
I don't think it would matter that much for settings, I think it's more for JAM/JR stuff.
We might want to take a look at the errors, not sure if init order changed in 1.74?
Are you using `CBA_settings_fnc_init` or `CBA_fnc_addSetting`? (later should be safer to use)

ColdEvul:
still use the `CBA_settings_fnc_init` have not gotten to convert this yet.
I will put upp a issue if it still exist after the convert.

dedmen:
>Whould this effect load order of cba_settings in anyway?

CBA still loads after all vanilla stuff, just like it always did.

PR #1172, opened 2019-06-29
Accessiblity - Allow changing marker colors

PabstMirror:
Just proof of concept right now
allows changing lines/marker color for people with color problems
requires mission restart

Colors also usable by other mods?
I had been thinking of ace_nametags but those colors are different from map colors
and sthud and diwako_dui both already have good color options

### DIFF
--- a/addons/accessiblity/initSettings.sqf
+++ b/addons/accessiblity/initSettings.sqf
@@ -1,0 +1,26 @@
+[
+QGVAR(red), "COLOR",
+["str_team_red"],
+["CBA Accessiblity"],
+[0.9, 0, 0, 1],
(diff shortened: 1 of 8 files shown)


addons/accessiblity/initSettings.sqf
```diff
@@ -0,0 +1,26 @@
+[
+QGVAR(red), "COLOR",
+["str_team_red"],
+["CBA Accessiblity"],
+[0.9, 0, 0, 1],
+2, // setting can't be overwritten
+{TRACE_1("red",_this);}
+] call CBA_settings_fnc_init;
```
commy2: `CBA_fnc_addSetting`

PR #1174, opened 2019-07-03
add functions to play and stop sounds with speed of sound delay

commy2:
- description tbd

- [ ] `CBA_play`/`stopSoundSOSLooped`

### DIFF
--- a/addons/events/fnc_executeWhenSoundWaveArrived.sqf
+++ b/addons/events/fnc_executeWhenSoundWaveArrived.sqf
@@ -1,0 +1,65 @@
+#include "script_component.hpp"
+/* ----------------------------------------------------------------------------
+    Function: CBA_fnc_executeWhenSoundWaveArrived
+
+    Description:
+        Executes code snippet if sound from an origin has arrived at the camera.
+
+    Parameters:
+        _origin    - Object or position from where sound emits <OBJECT, PosASL>
+        _code      - Code to execute <CODE>
+        _arguments - Additional arguments to pass to code, default [] <ANY>
+
+    Returns:
+        Nothing.
+
+    Passed Arguments:
+        _origin    - Same as above OR null object <OBJECT>
+        _position  - Current position of origin OR initial position <PosASL>
+        _arguments - Same as above <NUMBER>
+
+    Examples:
+    (begin example)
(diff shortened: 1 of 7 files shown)


addons/events/fnc_executeWhenSoundWaveArrived.sqf
```diff
@@ -0,0 +1,65 @@
...
+        Nothing.
+
+    Passed Arguments:
+        _origin    - Same as above OR null object <OBJECT>
+        _position  - Current position of origin OR initial position <PosASL>
+        _arguments - Same as above <NUMBER>
+
+    Examples:
+    (begin example)
+        [{playSound "Alarm"}, cursorTarget] call CBA_fnc_executeWhenSoundWaveArrived;
```
veteran29: Arguments order seems to be wrong here.

addons/events/initSOS.sqf
```diff
@@ -0,0 +1,81 @@
...
+            _source = (allMissionObjects "#soundonvehicle" - _source) param [0, objNull];
+
+            GVAR(publicSounds) setVariable [_handler, [_source, _origin]];
+        }, [_sound, _handler]] call CBA_fnc_executeWhenSoundWaveArrived;
+    }] call CBA_fnc_addEventHandler;
+
+    ["CBA_stopSoundSOS", {
+        params ["_handler"];
+
+        (GVAR(publicSounds) getVarible [_handler, [nil, objNull]]) params ["", "_origin"];
```
veteran29:
```suggestion
        (GVAR(publicSounds) getVariable [_handler, [nil, objNull]]) params ["", "_origin"];
```

addons/events/fnc_executeWhenSoundWaveArrived.sqf
```diff
@@ -0,0 +1,65 @@
...
+
+    private _position = _origin;
+    if (_position isEqualType objNull) then {
+        _position = getPosASL _position;
+    } else {
+        _origin = objNull;
+    };
+
+    private _distanceToCamera = _position vectorDistance AGLToASL positionCameraToWorld [0,0,0];
+    private _travelledDistance = (CBA_missionTime - _startTime) * SPEED_OF_SOUND;
```
PabstMirror: What if player is underwater?

commy2: I don't want to think about it.

PR #1183, opened 2019-07-25
1.94 - Fix GL magwells

PabstMirror:
> Added: magazineWells to Underbarrel Grenade Launchers

New vanilla configs:
```
    class UGL_F: GrenadeLauncher {
        magazineWell[] = {"UGL_40x36"};
        
    class arifle_MX_GL_F: arifle_MX_Base_F {
        class GL_3GL_F: UGL_F {
                magazineWell[] = {"UGL_40x36","3UGL_40x36"};
```

This just uses `=` instead of `+=` because of problems with inhetiance
The proper way would be to add a subconfig: https://github.com/CBATeam/CBA_A3/commit/8e24cb6788dffe94b1ea8c015342494cea650975
but not sure if we want to add another cfgPatch just for something minor?

### DIFF
--- a/addons/jam/CfgWeapons.hpp
+++ b/addons/jam/CfgWeapons.hpp
@@ -22,14 +22,14 @@ class CfgWeapons {
 
     class GrenadeLauncher;
     class UGL_F: GrenadeLauncher {
-        magazineWell[] += {"CBA_40mm_M203","CBA_40mm_EGLM"}; // empty in vanilla
+        magazineWell[] += {"CBA_40mm_M203","CBA_40mm_EGLM"};
     };
 
     class arifle_MX_Base_F: Rifle_Base_F {
         magazineWell[] += {"CBA_65x39_MX","CBA_65x39_MX_XL"};
 
         class GL_3GL_F: UGL_F {
-            magazineWell[] += {"CBA_40mm_3GL","CBA_40mm_M203","CBA_40mm_EGLM"}; // [] in vanilla
+            magazineWell[] = {"UGL_40x36","3UGL_40x36","CBA_40mm_3GL","CBA_40mm_M203","CBA_40mm_EGLM"}; // just list all wells because of += limitations
         };
     };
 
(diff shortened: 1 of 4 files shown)


commy2:
JR has a prep component and I cut that down to be a very basic sub component. I don't really understand why it is needed here though.

addons/jam/CfgWeapons.hpp
```diff
@@ -22,14 +22,14 @@ class CfgWeapons {
...
-        magazineWell[] += {"CBA_40mm_M203","CBA_40mm_EGLM"}; // empty in vanilla
+        magazineWell[] += {"CBA_40mm_M203","CBA_40mm_EGLM"};
     };
 
     class arifle_MX_Base_F: Rifle_Base_F {
         magazineWell[] += {"CBA_65x39_MX","CBA_65x39_MX_XL"};
 
         class GL_3GL_F: UGL_F {
-            magazineWell[] += {"CBA_40mm_3GL","CBA_40mm_M203","CBA_40mm_EGLM"}; // [] in vanilla
+            magazineWell[] = {"UGL_40x36","3UGL_40x36","CBA_40mm_3GL","CBA_40mm_M203","CBA_40mm_EGLM"}; // just list all wells because of += limitations
```
commy2:
```suggestion
            magazineWell[] += {"CBA_40mm_3GL"};
```
Am I missing something? If I use that ^, result ingame is:
`magazineWell[] = {"CBA_40mm_M203","CBA_40mm_EGLM","CBA_40mm_3GL"};`
Which is exactly what we want, no?

PabstMirror: Missing `"UGL_40x36","3UGL_40x36"`, I think we use `+=` so we don't break compat with mods using vanilla magwells

commy2: Ah, now I understand the problem.

commy2:
It is exactly as you wrote, Pabst. I went with what you suggested, because another config is fine for this imo as it is needed by limitation of +=.

I went with "finish" as the opposite of "**prep**eration" (used in JR and UI).

Merge?

PabstMirror:
:finland: lgtm

PR #1186, opened 2019-07-26
fix minor ui issues with Contact DLC ui modifications

commy2:
This PR fixes:
- when Contact DLC is loaded only, when a tab in the ADDON OPTIONS menu is selected, but not focused, the text appears black on black background, when it should remain white. Without Contact DLC, it is supposed to still turn black.

before:

after:


commy2:
- When Contact DLC is loaded only, the ADDON OPTIONS button (and all other buttons added to the menu by our framework like ACE Fix Headbug etc.) is centered instead of flush left, because with Contact DLC they changed the behavior of RscButtonMain and also replaced these buttons to inherit from RscButtonMenuLeft (a new class to get the legacy behavior?).

before:


after:

Issue #1024, opened 2018-11-15
Task: Update music module


dedmen:
**CBA Version:** `3.9.0` (stable)

**Description:**
CBA Music module isn't really useful in it's current implementation

<https://github.com/CBATeam/CBA_A3/blob/master/addons/music/fnc_isMusicPlaying.sqf> just does a isNil check, but nothing set's that variable to nil if the music stopped. Unless you call `CBA_fnc_getMusicPlaying` before, and even that only detects if the remaining time ran out, not if the music has been stopped by something else.
It should be using the MusicStopped EH <https://community.bistudio.com/wiki/addMusicEventHandler>.

I got the Idea through this thread: https://armaworld.de/index.php?thread/4165-musik-als-dauerschleife-%C3%BCber-description-ext/&postID=33251#post33251

PR #1192, opened 2019-07-29
Put tags in CfgMusic.hpp into quotes

R3voA3:
Put tags in CfgMusic.hpp into quotes

Was suggested by @dedmen in https://github.com/CBATeam/CBA_A3/issues/1024

### DIFF
--- a/addons/music/CfgMusic.hpp
+++ b/addons/music/CfgMusic.hpp
@@ -7,212 +7,212 @@ class CfgMusic
     {
         type = "soundtrack";
         theme = "safe";
-        tags[] = {bis, lead};
+        tags[] = {"bis", "lead"};
     };
     class LeadTrack01a_F
     {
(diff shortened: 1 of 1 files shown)


commy2:
- Opening curly brackets should be at EOL not a new line just for them.
- What is the point of this file anyway? Are these not all base game classes?
- `theme` entry is just copy pasted from the base game config?

dedmen:
>What is the point of this file anyway? Are these not all base game classes?

Yes but I think we are just adding tags to the new music stuff thingy?
When I looked at the music PR this stuff made sense I guess.

dedmen:
>Removed double quotes from some theme entries.

Well as you can see you clearly can't trust my reviews 🥇

RevoArma3:
There are a few tracks missing. Should I add them aswell?

PR #1145, opened 2019-05-14
Adapt to HEMTT

jonpas:
**When merged this pull request will:**
- Use HEMTT as build system (uses armake2)
- Add `build.bat` for quick Windows builds without command line and `build_release.bat` for release builds
- Add script `setup.bat` to download project-specific HEMTT binary
  - No Linux script as downloading an artifact requires authentication
- Swap Travis CI and Circle CI with GitHub Actions
  - _No builds yet as Actions HEMTT is too old version (`/` comment bug)_
- Add custom HEMTT versioning script
  - Update versions in files (like `make.py`) in `prebuild` and reset in `postbuild`
  - Version source is `addons/main/script_version.hpp`
- [Documentation](https://github.com/CBATeam/CBA_A3/wiki/Building-CBA-from-source-(HEMTT))

**TODO armake2/HEMTT:**
- [x] Fix subconfig not processed correctly (https://github.com/KoffeinFlummi/armake2/issues/45)
- [x] Fix preprocessor define and comment on same line (https://github.com/KoffeinFlummi/armake2/issues/24)
  - [x] Remove temporary removal of comments
- [x] Fix preprocessor comment with `/` on same line ([example](https://github.com/CBATeam/CBA_A3/blob/0464ca4a143b92c71a0f8e04c0abac649841358f/addons/jam/CfgMagazineWells.hpp#L59)) (https://github.com/synixebrett/HEMTT/commit/477e44c5ef807f1a4ba903ce00adb7682f132d49)
- [x] Folder copy on release (https://github.com/synixebrett/HEMTT/issues/88)

---

**TODO CBA Later (User):**
- [ ] Use HEMTT file-patching symlink setup instead of `setup.py`
- [ ] Add release/publish procedure (ref. https://github.com/CBATeam/CBA_A3/pull/730#issuecomment-411018642) (HEMTT script)
- ~~Use Windows Installer instead of project-specific binary and download script (https://github.com/synixebrett/HEMTT/pull/60)~~ Will stay project-specific
- [ ] Revert #1199

**TODO HEMTT Later:**
- [ ] Release HEMTT 0.8.0 and use in CI (contains `/` comment and folder copy on release)

PR #1199, opened 2019-08-04
Nobin xeh

commy2:
- fix for bad Mikeros Tools hating on scope entry using simple expressions...


jonpas:
Will be reverted with #1145

PR #1201, opened 2019-08-06
Add Facewear Randomization

jonpas:
- Add working replacement for 'allowedHeadgear' of `BIS_fnc_unitHeadgear` but on unit level config.
  - Supports `BIS_enableRandomization` variable and `disableRandomization[]` mission config.

Integrated from BWMod and modified for better systems support.

**Design decisions:**
- ~~Use vanilla allowedFacewear config entry -> automatically works for all mods~~
- ~~Overwrite `BIS_fnc_unitHeadgear` instead -> yes/no? feasible? -> would require mods to make sure init EH calls it (like `BIS_fnc_initVehicle`), no need with separate function~~


commy2:
>Use vanilla allowedFacewear config entry -> automatically works for all mods
?
Vanilla `allowedFacewear` is a CfgWeapons entry for headgear in vanilla, not a CfgVehicles entry for soldiers. By not having this global space config entry tagged, you introduce a new untagged config entry in CBA that also shares a name with an unrelated base game entry.

Note that randomized facewear without headgear can already be achieved with a dummy headgear class like `H_FakeHeadgear_Syndikat_F`:
```cpp
class CfgWeapons {
    class H_FakeHeadgear_Syndikat_F {
        author = "Bohemia Interactive";
        scope = 0;
        allowedFacewear[] = {
            "G_Aviator", 1,
            "G_Bandanna_aviator", 1,
            "G_Bandanna_beast", 1,
            "G_Bandanna_blk", 1,
            "G_Bandanna_khk", 1,
            "G_Bandanna_oli", 1,
            "G_Bandanna_shades", 1,
            "G_Bandanna_sport", 1,
            "G_Bandanna_tan", 1,
            "G_Shades_Black", 1,
            "G_Shades_Blue", 1,
            "G_Shades_Red", 1,
            "G_Shades_Green", 1
        };
    };
};

class CfgVehicles {
    class I_C_Soldier_base_F: I_G_Soldier_base_F {
        class EventHandlers: EventHandlers {
            init = "if (local (_this select 0)) then {[(_this select 0), [], []] call BIS_fnc_unitHeadgear;};";
        };

        headgearList[] = {
            "", 2,
            "H_FakeHeadgear_Syndikat_F", 5,
            "H_Bandanna_gry", 1,
            "H_Bandanna_blu", 1,
            "H_Bandanna_cbr", 1,
            "H_Bandanna_khk", 1,
            "H_Bandanna_sgg", 1,
            "H_Bandanna_sand", 1,
            "H_Bandanna_surfer_blk", 1,
            "H_Bandanna_camo", 1,
            "H_Booniehat_khk", 1,
            "H_Booniehat_oli", 1,
            "H_Booniehat_tan", 1,
            "H_Cap_blk_Syndikat_F", 1,
            "H_Cap_grn_Syndikat_F", 1,
            "H_Cap_oli_Syndikat_F", 1,
            "H_Cap_tan_Syndikat_F", 1
        };
    };
};
```

addons/common/fnc_randomizeFacewear.sqf
```diff
@@ -0,0 +1,44 @@
...
+
+if (!local _unit || {!(_unit getVariable ["BIS_enableRandomization", true])}) exitWith {true};
+
+private _allowedFacewear = getArray (configFile >> "CfgVehicles" >> typeOf _unit >> "CBA_allowedFacewear");
+
+if (_allowedFacewear isEqualTo []) exitWith {true};
+
+private _facewear = selectRandomWeighted _allowedFacewear;
+
+if (_facewear == "") then {
```
jonpas: This is changed from BWA version, which uses `"None"`. I did this to keep it same as `headgearList[]`.

addons/common/fnc_randomizeFacewear.sqf
```diff
@@ -0,0 +1,44 @@
...
+
+Parameters:
+    _unit - unit <OBJECT>
+
+Returns:
+    true on success, false on error <BOOLEAN>
+
+Examples:
+    (begin example)
+        [unit] call CBA_fnc_randomizeFacewear
```
commy2: ``´suggestion
        [unit] call CBA_fnc_randomizeFacewear;
```

File: addons/common/fnc_randomizeFacewear.sqf [hunk 3]
```diff
@@ -0,0 +1,44 @@
+#include "script_component.hpp"
+/* ----------------------------------------------------------------------------
+Function: CBA_fnc_randomizeFacewear
+
+Description:
+    Add config defined weighted random facewear to unit.
+    Uses same config as headgearList[].
```
commy2:
```suggestion
```

addons/common/fnc_randomizeFacewear.sqf
```diff
@@ -0,0 +1,44 @@
...
+
+Examples:
+    (begin example)
+        [unit] call CBA_fnc_randomizeFacewear
+    (end)
+
+Author:
+    commy2
+---------------------------------------------------------------------------- */
+SCRIPT(randomizeFacewear);
```
commy2:
```suggestion
```

addons/common/fnc_randomizeFacewear.sqf
jonpas: Need to write somehow how the config looks.

addons/common/fnc_randomizeFacewear.sqf
```diff
@@ -0,0 +1,43 @@
...
+
+if (_allowedFacewear isEqualTo []) exitWith {true};
+
+private _facewear = selectRandomWeighted _allowedFacewear;
+
+if (_facewear == "") then {
+    removeGoggles _unit;
+} else {
+    _unit addGoggles _facewear;
+};
```
commy2: Header says returns true on success and false on error. This never returns false though and may return nil.
What would be considered failure? Disabled randomization by class from mission config or by object from object namespace variable?

jonpas: It returns `false` if passed unit is null. Forgot to add a `true` at the end. FYI, I wrote this PR on my phone. :eyes: 

I am not sure.

commy2: Okay, that is acceptable.
```suggestion

true

```

addons/common/fnc_randomizeFacewear.sqf
```diff
@@ -0,0 +1,43 @@
...
+Examples:
+    (begin example)
+        [unit] call CBA_fnc_randomizeFacewear;
+    (end)
+
+Author:
+    commy2
+---------------------------------------------------------------------------- */
+
+params [["_unit", objNull]];
```
commy2:
```suggestion
 params [["_unit", objNull, [objNull]]];
```
Public function params trinity

addons/common/fnc_randomizeFacewear.sqf
commy2:
```suggestion

CBA_facewearList[] = {}; // default: ignore and use facewear from BIS_fnc_unitHeadgear instead
CBA_facewearList[] = {"", 1}; // no facewear, delete and overwrite facewear from BIS_fnc_unitHeadgear

// 40% no facewear, 30% balaclava, 30% bandana. Weighted randomization: Sum of propabilites must not necessarily equal 1.
CBA_facewearList[] = {"", 0.4, "G_Balaclava_blk", 0.3, "G_Bandanna_blk", 0.3};
```

addons/common/fnc_randomizeFacewear.sqf
```diff
@@ -0,0 +1,43 @@
...
+
+if (isNull _unit) exitWith {
+    TRACE_1("unit is null",_unit);
+    false
+};
+
+if (!local _unit || {!(_unit getVariable ["BIS_enableRandomization", true])}) exitWith {true};
```
commy2: Should support https://community.bistudio.com/wiki/Description.ext#disableRandomization
with same precedence rules as described.
findIf loop for case insensitivity.

addons/common/fnc_randomizeFacewear.sqf
```diff
@@ -0,0 +1,43 @@
...
+    (end)
+
+Author:
+    commy2
+---------------------------------------------------------------------------- */
+
+params [["_unit", objNull]];
+
+if (isNull _unit) exitWith {
+    TRACE_1("unit is null",_unit);
```
commy2: WARNING_1 macro instead? (has to use %1)

addons/common/fnc_randomizeFacewear.sqf
```diff
@@ -0,0 +1,43 @@
...
+    false
+};
+
+if (!local _unit || {!(_unit getVariable ["BIS_enableRandomization", true])}) exitWith {true};
+
+private _allowedFacewear = getArray (configFile >> "CfgVehicles" >> typeOf _unit >> "CBA_allowedFacewear");
```
commy2: `CBA_facewearList` and `_facewearList`

addons/common/fnc_randomizeFacewear.sqf
commy2: This is good. BI seems to have changed their system, it used NONE dummy class in CfgGlasses when I wrote the original.

addons/common/fnc_randomizeFacewear.sqf
jonpas: Other things just use a TRACE, but I guess they are meant to fail silently.

jonpas:
Tested working with following config:
```
CBA_facewearList[] = {
    "", 0.5,
    "G_Shades_Black", 0.25,
    "G_Shades_Blue", 0.25,
    "G_Shades_Green", 0.25,
    "G_Shades_Red", 0.25
};
```


`disableRandomization[]` and `BIS_enableRandomization -> false` also work correctly: they get `"G_Shades_Black"` applied and no headgear, which is correct as per their default loadout.

commy2:
Wikipage for doc?

addons/common/fnc_randomizeFacewear.sqf
commy2: Thought more about this.

BI still has this `CfgGlasses/None` class. While public scope, it clearly is not intended to be worn. It even has a bug where it displays the DLC icon in inventory (yet doesn't trigger the DLC ads spam).

I'd suggest this just to be rigorous:
```suggestion
if (toLower _facewear in ["", "none"]) then {
```

PR #1202, opened 2019-08-09
6.5mm magwells fix

Loupeznik:
- Add colored MX magazines to both MX magwells
- Remove unintended Katiba magazines from 30rnd MX magwell

MX magwells seemed broken if combined with other addons (e.g. colored mags wouldn't appear in arsenal, don't know which mod causes this exactly), also 30rnd Katiba mags shouldn't be able to be loaded into MX rifles.

### DIFF
--- a/addons/jam/magwells_65x39.hpp
+++ b/addons/jam/magwells_65x39.hpp
@@ -1,15 +1,21 @@
     class CBA_65x39_MX {
         BI_mags[] = {
             "30Rnd_65x39_caseless_mag",
-            "30Rnd_65x39_caseless_green",
             "30Rnd_65x39_caseless_mag_Tracer",
-            "30Rnd_65x39_caseless_green_mag_Tracer"
+            "30Rnd_65x39_caseless_khaki_mag_Tracer",
+            "30Rnd_65x39_caseless_khaki_mag",
+            "30Rnd_65x39_caseless_black_mag_Tracer",
+            "30Rnd_65x39_caseless_black_mag",
         };
     };
     class CBA_65x39_MX_XL {
         BI_mags[] = {
             "100Rnd_65x39_caseless_mag",
-            "100Rnd_65x39_caseless_mag_Tracer"
+            "100Rnd_65x39_caseless_mag_Tracer",
+            "100Rnd_65x39_caseless_khaki_mag_Tracer",
+            "100Rnd_65x39_caseless_khaki_mag",
+            "100Rnd_65x39_caseless_black_mag_Tracer",
+            "100Rnd_65x39_caseless_black_mag",
         };
     };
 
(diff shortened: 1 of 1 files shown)


addons/jam/magwells_65x39.hpp
```diff
@@ -1,15 +1,21 @@
     class CBA_65x39_MX {
         BI_mags[] = {
             "30Rnd_65x39_caseless_mag",
-            "30Rnd_65x39_caseless_green",
             "30Rnd_65x39_caseless_mag_Tracer",
-            "30Rnd_65x39_caseless_green_mag_Tracer"
+            "30Rnd_65x39_caseless_khaki_mag_Tracer",
+            "30Rnd_65x39_caseless_khaki_mag",
+            "30Rnd_65x39_caseless_black_mag_Tracer",
+            "30Rnd_65x39_caseless_black_mag",
```
commy2: - remove extra comma after last array element
- please sort _Tracer after normal mag

addons/jam/magwells_65x39.hpp
```diff
@@ -1,15 +1,21 @@
...
     };
     class CBA_65x39_MX_XL {
         BI_mags[] = {
             "100Rnd_65x39_caseless_mag",
-            "100Rnd_65x39_caseless_mag_Tracer"
+            "100Rnd_65x39_caseless_mag_Tracer",
+            "100Rnd_65x39_caseless_khaki_mag_Tracer",
+            "100Rnd_65x39_caseless_khaki_mag",
+            "100Rnd_65x39_caseless_black_mag_Tracer",
+            "100Rnd_65x39_caseless_black_mag",
```
commy2: - same as above

commy2:
Note that this is not needed for the MX after upcoming update CBA v3.12.1, because the MX will now be using the base game magazine class. But it is still desired for third party weapons that use this jam magwell.

commy2:
```

"30Rnd_65x39_caseless_mag",
"30Rnd_65x39_caseless_mag_Tracer",
"30Rnd_65x39_caseless_khaki_mag",
"30Rnd_65x39_caseless_khaki_mag_Tracer",
"30Rnd_65x39_caseless_black_mag",
"30Rnd_65x39_caseless_black_mag_Tracer"
```
This is the best looking, don't you agree?

Loupeznik:
> 
> 
> ```
> 
> "30Rnd_65x39_caseless_mag",
> "30Rnd_65x39_caseless_mag_Tracer",
> "30Rnd_65x39_caseless_khaki_mag",
> "30Rnd_65x39_caseless_khaki_mag_Tracer",
> "30Rnd_65x39_caseless_black_mag",
> "30Rnd_65x39_caseless_black_mag_Tracer"
> ```
> 
> This is the best looking, don't you agree?

yeah, it looks better. Corrected.

PR #1203, opened 2019-08-12
Czech translation

Loupeznik:
- Add full Czech translation

### DIFF
--- a/addons/strings/stringtable.xml
+++ b/addons/strings/stringtable.xml
@@ -17,7 +17,7 @@
         </Key>
         <Key ID="STR_CBA_FORMAT_NUMBER_DECIMAL_POINT">
             <English>.</English>
-            <Czech>.</Czech>
+            <Czech>,</Czech>
             <German>,</German>
             <Polish>,</Polish>
             <Russian>,</Russian>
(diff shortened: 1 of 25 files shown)


addons/strings/stringtable.xml
```diff
@@ -31,12 +32,8 @@
         </Key>
         <Key ID="STR_CBA_FORMAT_NUMBER_THOUSANDS_SEPARATOR">
             <English>,</English>
-            <Czech> </Czech>
+            <Czech>,</Czech>
             <German>.</German>
-            <Polish> </Polish>
```
PabstMirror: don't want to drop these

Loupeznik: Not quite sure how they got removed in the first place, I wasn't doing anything with those. Maybe some Tabler issue, I'll revise the XMLs manually.

commy2: Probsbly tabler cleaning white space at EOL and then the entries being empty so they get removed.

commy2:
This is done?

Loupeznik:
Yeah, it should be complete

bux:
i love pabst's subtle humour

commy2:
Thank you!

PR #1207, opened 2019-08-15
Quote VERSION in VERSION_CONFIG

### LINKED ISSUE #1206: unquoted string warning
**Arma 3 Version:** `latest` (stable)
**CBA Version:** `latest` (stable)

**Mods:**
```
- CBA_A3
```

**Description:**
Sorry to bother you with such a minor question but i am a beginner and would like be sure that things i do, are correct.

I've got i warning about "unquoted string" in config.cpp compiled with armake2. My issue is more sort of question. Are you sure you macro ´´´VERSION_CONFIG´´´ definition/usage  is correct? if so this would a armake2 issue not cba ...

```
In file addons\main\config.cpp:11: warning: String value "0.1.0.0" is not quoted properly. [unquoted-string]
```
addons\main\config.cp:
```
class CfgPatches {
    class ADDON {
        name = CSTRING(COMPONENT);
        units[] = {};
        weapons[] = {};
        requiredVersion = REQUIRED_VERSION;
        requiredAddons[] = {};
        author = "test";
        VERSION_CONFIG;
    };
};
````
**Steps to reproduce:**
1. hemtt create
...
2. hemtt build

using hemtt / armake2 toolset.

**RPT log file:**

### PR DESCRIPTION
**Pull request by jonpas**
- Quote `VERSION` in `VERSION_CONFIG` macro properly - fix #1206 
- Fix BOM (and excessive space) in some files.

Cherry-picked from #1145.

### DIFF
--- a/addons/main/script_macros_common.hpp
+++ b/addons/main/script_macros_common.hpp
@@ -60,7 +60,7 @@
 #endif
 
 #ifndef VERSION_CONFIG
-    #define VERSION_CONFIG version = VERSION; versionStr = QUOTE(VERSION); versionAr[] = {VERSION_AR}
+    #define VERSION_CONFIG version = QUOTE(VERSION); versionStr = QUOTE(VERSION); versionAr[] = {VERSION_AR}
 #endif
 
 #define ADDON DOUBLES(PREFIX,COMPONENT)
--- a/addons/optics/script_component.hpp
+++ b/addons/optics/script_component.hpp
@@ -1,4 +1,4 @@
-﻿#define COMPONENT optics
+#define COMPONENT optics
 #include "\x\cba\addons\main\script_mod.hpp"
 
 //#define DEBUG_MODE_FULL
--- a/addons/ui/flexiMenu/data/menu_arma2.hpp
+++ b/addons/ui/flexiMenu/data/menu_arma2.hpp
@@ -1,4 +1,4 @@
-﻿//#include "\x\cba\addons\ui\script_component.hpp"
+//#include "\x\cba\addons\ui\script_component.hpp"
 
 #define _DefaultAspectRatio 3 / 4
 #define _SX (safeZoneX+safeZoneW / 2) // screen centre x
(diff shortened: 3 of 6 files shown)

### CODE REVIEW

File: addons/main/script_macros_common.hpp [hunk 1]
```diff
@@ -60,7 +60,7 @@
 #endif
 
 #ifndef VERSION_CONFIG
-    #define VERSION_CONFIG version = VERSION; versionStr = QUOTE(VERSION); versionAr[] = {VERSION_AR}
+    #define VERSION_CONFIG version = QUOTE(VERSION); versionStr = QUOTE(VERSION); versionAr[] = {VERSION_AR}
```
commy2: If `versionStr` is `QUOTE(VERSION)`, then why should `version` be `QUOTE(VERSION)` as well?

Isn't it more likely that version is supposed to be a number, which works as long as there is just one comma inside the version number? After this we convert stuff like 1.00 to "1.00", even though that versionStr was already "1.00".

I don't think the choice of tool should dictate this.

I know it is a shitty way of writing this, but this we inherited from CBA_A2.

jonpas: Even CBA doesn't have a version number like that, but one with 2 dot separators, which means Arma quotes it anyways (thus is wrong).

Tool is not dictating this, Arma and the macro use in CBA and most mods is. Any proper tool will give you a warning for missing quoting. I thought stuff from A2 CBA is supposed to be rewritten anyways. ;)

I don't have a solution to provide backwards compatibility, but CBA should be without inconsistencies like this.

commy2:
Don't like this for its potential to turn NUMBER tokens into STRINGs for addons that use CBA and want to entry to remain the type it is.

Close?

jonpas:
As stated above, all mods using 3-number versioning already convert this to string, Arma does it.

This is technically wrong, Arma converts it, so it's not a number, and CBA shouldn't have magic conversions. 

We could change it so this part only takes MAJOR/MINOR in mods (such as CBA itself), that way it will actually stay a number in all cases.

PR #1209, opened 2019-08-17
Common - Prevent endless loop with waitAndExec

### LINKED ISSUE #1208: crash on CBA_fnc_execNextFrame and map close
**Arma 3 Version:** `1.94.145977` stable
**CBA Version:** ` 3.12.1` stable (steam workshop)

**Mods:**
```
- CBA_A3
```

**Description:**
Arma crash every time when CBA_fnc_execNextFrame is used in loop on map close event.

**Steps to reproduce:**
```
systemChatfnc = {
    systemChat str time;
    [
        {
            if (testFnc) then {
                call systemChatfnc
            };
        },
        []
    ]call CBA_fnc_execNextFrame;
};
testFnc = true;
call systemChatfnc;
```

open and close map. Just before map close, arma will crash.

Editor (Singleplayer) / Editor (Multiplayer) (only here tested)

**RPT log file:**
unnecessary, nothing valuable is there, as the arma crashes completely. Tested on 2 PC.
edit:
rpt and mdmp file:
https://github.com/CBATeam/CBA_A3/files/3511987/Arma.3.zip

PabstMirror:
Fix #1208

### DIFF
--- a/addons/common/init_perFrameHandler.sqf
+++ b/addons/common/init_perFrameHandler.sqf
@@ -3,8 +3,6 @@
 
 #include "script_component.hpp"
 
-#define DELAY_MONITOR_THRESHOLD 1 // Frames
-
 GVAR(perFrameHandlerArray) = [];
 GVAR(perFrameHandlersToRemove) = [];
 GVAR(lastTickTime) = diag_tickTime;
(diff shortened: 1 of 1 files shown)


addons/common/init_perFrameHandler.sqf
```diff
@@ -56,7 +56,10 @@ GVAR(waitUntilAndExecArray) = [];
         _delete = false;
     };
 
-
+    if (diag_frameno != GVAR(nextFrameNo)) then { // in case we skipped a frame, ensure we don't add to nextFrameBufferA while iterating
```
PabstMirror: this should be moved up to line 22 to ensure execNextFrame is also accurate for pfeh and waitAndExec

commy2:
WIP or done?

PabstMirror:
Should be good to go,

PR #1210, opened 2019-08-19
Update online docs for 3.12.1

Teqashi:
@veteran29 complained about it being outdated and his lack of time to update it, so here I am 😄 

Can be tested on my lab:
https://jira.sdsznsk.pl/files/overview-txt.html
(Ignore TLD 🙃 )

// Built on Debian 9.7 with naturaldocs 1.51-2 package
// Tested on newest Opera and Edge

- Update online docs to newest version that is on master currently


jonpas:
Nice!

We should automate that with GitHub Actions (when it releases) if possible.

commy2:
Did we really add so much stuff that it adds up to 1000 lines of documentation?

jonpas:
@commy2 There is also NaturalDocs version increase and probably some structure changes with that.

PR #1211, opened 2019-08-19
make_docs script fix, ndocs menu items fix

Teqashi:
Fixed code in make_docs shell script so it will run (at least on Debian 9). Added ndocs-project menu categories for new functions (like disposables or optics). Also some minor fixes in syntax (when pulled, wouldn't even run in Deb 9.7)

// Tested and built on Debian 9.7 with naturaldocs 1.51-2 package

- Provide compatibility with Debian 9+ (Fix syntax)
- Fix menu items (new categories)
- Update ndocs-project tree for newest functions

### DIFF
--- a/tools/ndocs-project/Menu.txt
+++ b/tools/ndocs-project/Menu.txt
@@ -1,4 +1,4 @@
-Format: 1.52
+Format: 1.51
 
 
 Title: CBA
@@ -67,13 +67,14 @@ Group: Functions  {
    Group: Ai  {
 
       File: CBA_fnc_addWaypoint  (ai/fnc_addWaypoint.sqf)
+      File: CBA_fnc_buildingPositions  (ai/fnc_buildingPositions.sqf)
       File: CBA_fnc_clearWaypoints  (ai/fnc_clearWaypoints.sqf)
       File: CBA_fnc_searchNearby  (ai/fnc_searchNearby.sqf)
       File: CBA_fnc_taskAttack  (ai/fnc_taskAttack.sqf)
       File: CBA_fnc_taskDefend  (ai/fnc_taskDefend.sqf)
       File: CBA_fnc_taskPatrol  (ai/fnc_taskPatrol.sqf)
       File: CBA_fnc_taskSearchArea  (ai/fnc_taskSearchArea.sqf)
-      File: CBA_fnc_waypointGarrison  (ai/fnc_waypointGarrison.sqf)
+      File: fnc_waypointGarrison  (ai/fnc_waypointGarrison.sqf)
       }  # Group: Ai
 
    Group: Arrays  {
@@ -92,6 +93,7 @@ Group: Functions  {
       File: CBA_fnc_join  (arrays/fnc_join.sqf)
       File: CBA_fnc_reject  (arrays/fnc_reject.sqf)
       File: CBA_fnc_select  (arrays/fnc_select.sqf)
+      File: CBA_fnc_selectRandomArray  (arrays/fnc_selectRandomArray.sqf)
       File: CBA_fnc_shuffle  (arrays/fnc_shuffle.sqf)
       File: CBA_fnc_sortNestedArray  (arrays/fnc_sortNestedArray.sqf)
       }  # Group: Arrays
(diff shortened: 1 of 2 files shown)


tools/ndocs-project/Menu.txt
```diff
@@ -67,13 +67,14 @@ Group: Functions  {
...
       File: CBA_fnc_addWaypoint  (ai/fnc_addWaypoint.sqf)
+      File: CBA_fnc_buildingPositions  (ai/fnc_buildingPositions.sqf)
       File: CBA_fnc_clearWaypoints  (ai/fnc_clearWaypoints.sqf)
       File: CBA_fnc_searchNearby  (ai/fnc_searchNearby.sqf)
       File: CBA_fnc_taskAttack  (ai/fnc_taskAttack.sqf)
       File: CBA_fnc_taskDefend  (ai/fnc_taskDefend.sqf)
       File: CBA_fnc_taskPatrol  (ai/fnc_taskPatrol.sqf)
       File: CBA_fnc_taskSearchArea  (ai/fnc_taskSearchArea.sqf)
-      File: CBA_fnc_waypointGarrison  (ai/fnc_waypointGarrison.sqf)
+      File: fnc_waypointGarrison  (ai/fnc_waypointGarrison.sqf)
```
jonpas: Why is this one different?

Teqashi: It's because in `File: XYZ` line, XYZ is a string, generated from header in a files, eg.
 `addons/ai/fnc_waypointGarrison.sqf`
```
Script: fnc_waypointGarrison
```

Notice that for example
`addons/ai/fnc_taskAttack.sqf`
```/* ----------------------------------------------------------------------------
Function: CBA_fnc_taskAttack
```
has a proper header, so ndocs generated `File: CBA_fnc_taskAttack` for it.

It's just a matter of fixing `Function` headers inside of fnc_ files

jonpas: Ah! In that case, let's fix the function headers instead. :)

Teqashi: Fixed headers in https://github.com/CBATeam/CBA_A3/pull/1211/commits/a467239b80af3bfa8b16e8975fc05b57166068f3 and https://github.com/CBATeam/CBA_A3/pull/1211/commits/3473b0f51d70042923a87e99791632f34bcc07f0

addons/ai/fnc_waypointGarrison.sqf
```diff
@@ -1,6 +1,6 @@
 #include "script_component.hpp"
 /* ----------------------------------------------------------------------------
-Script: fnc_waypointGarrison
+Function: CBA_fnc_waypointGarrison
```
commy2: But that's wrong. There is no such function defined, as it not appears in CfgFunctions.
This script is used in CfgWaypoints, which, for BI knows why, uses paths to script files to operate.
"Script" instead of "Function" to indicate its usage as scheduled _script_ (as in, not a _function_, because no return value).

jonpas: Oh my, that's retarded. It is usable outside of `CfgWaypoints` though? Could be compiled under `CfgFunctions` in that case.

commy2: I suppose it could be used as script outside of CfgWaypoints, but that would have to be in scheduled environment and would stall the current script instance, or... the "function" would have to be "spawned", which is the absolutely terrible BI way of doing things. Don't want that. Can't we just have the doc support "scripts" instead? Preferably ignore them, because proper code will only ever use them for such niche applications. Ignore as in just not list in the doc.

jonpas: Ignore sounds good.

tools/ndocs-project/Data/FileInfo.nd
```diff
@@ -1,521 +1,671 @@
-1.52
```
dedmen: Is this file supposed to be in here? It seems auto generated and environment dependent.

Teqashi: As far as I can see FileInfo.nd is:
```
An index of the state of the files as of the last parse.  Used to determine if files were added, deleted, or changed.
```
https://www.naturaldocs.org/archive/documentation/1.4/files/Modules/NaturalDocs/Project-pm.html#FileInfo.nd

Because ndocs updates only changed files during its rebuild
```
Natural Docs' build process is differential, which means each run it looks for what changed since its last run and then updates only those parts of the documentation.  This means that later runs are much faster than the first so you can run it often as you work.
```
https://www.naturaldocs.org/getting_started/building_the_documentation/

veteran29: Should be in. gitignore then.

jonpas: I say keep it, it'll be useful once we do it via CI.

addons/ai/fnc_waypointGarrison.sqf
```diff
@@ -1,6 +1,6 @@
 #include "script_component.hpp"
 /* ----------------------------------------------------------------------------
-Script: fnc_waypointGarrison
+Internal Script: CBA_fnc_waypointGarrison
```
jonpas:
```suggestion
Internal Script: fnc_waypointGarrison
```
It's not compiled with prefix.

Teqashi: Yeah, I just noticed it, fixing it rn along with some other minor errors.

tools/ndocs-project/Data/FileInfo.nd
dedmen: It will be useful once we do it via CI yes, but with CI filepaths. These are environment dependent file paths. So currently that file is useless, we should re-add it with CI tho yes.

veteran29: I agree with Dedmen.

jonpas: Alright, I see what you mean.

addons/ai/fnc_waypointGarrison.sqf
veteran29: Bump ( ͡° ͜ʖ ͡°)

commy2:
State?

commy2:
state?

PR #1212, opened 2019-08-19
add CONTACT SERVER ADMIN header to signature mismatch message

commy2:
- Tired of the workshop drones, so append the crappy BI error messages so blame does not get shifted to this (so _me_ by association).

Open for more suggestions as error messages in this game are generally horrendous and offensively so.


jonpas:
Screenshot example?

PR #1213, opened 2019-08-21
Default transition targetstate to transition configName

PiZZAD0X:
- Change the default transition targetstate value to the transition configName, so that most transitions between states can have their transition state and targetstate defined without a near duplicate entry.

Example: 

```
class Init {
    onStateEntered = QFUNC(onCacheInit);
    class Cache {
        targetState = "Cache";
        events[] = {QGVAR(cache)};
    };

    class Uncache {
        targetState = "Uncache";
        events[] = {QGVAR(uncache)};
    };
};
```

Could now be written as: 

```
class Init {
    onStateEntered = QFUNC(onCacheInit);
    class Cache {
        events[] = {QGVAR(cache)};
    };

    class Uncache {
        events[] = {QGVAR(uncache)};
    };
};
```

while targetState 's that deviate from the transition configName are supported.

### DIFF
--- a/addons/statemachine/fnc_createFromConfig.sqf
+++ b/addons/statemachine/fnc_createFromConfig.sqf
@@ -45,7 +45,10 @@ private _stateMachine = [_list, _skipNull] call FUNC(create);
     private _state = configName _x;
     {
         private _transition = configName _x;
-        private _targetState = [_x,"targetState",_transition] call BIS_fnc_returnConfigEntry;
+        private _targetState = _transition;
+        if (isText (_x >> "targetState")) then {
+            _targetState = getText (_x >> "targetState");
+        };
         GET_FUNCTION(_condition,_x >> "condition");
         GET_FUNCTION(_onTransition,_x >> "onTransition");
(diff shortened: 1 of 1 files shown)


addons/statemachine/fnc_createFromConfig.sqf
```diff
@@ -45,7 +45,7 @@ private _stateMachine = [_list, _skipNull] call FUNC(create);
     private _state = configName _x;
     {
         private _transition = configName _x;
-        private _targetState = getText (_x >> "targetState");
+        private _targetState = [_x,"targetState",_transition] call BIS_fnc_returnConfigEntry;
```
veteran29:
```suggestion
        private _targetState = [_x >> "targetState", "text", _transition] call CBA_fnc_getConfigEntry;
```

PiZZAD0X: This or is/getText method?

commy2: isText

BaerMitUmlaut:
Sure why not.

PR #1214, opened 2019-08-23
handle returned assignments in key events

commy2:
- Atm. keybindings that return assignments error:

Test script (15 is tab key):
```
["Test001", "Action001", "Test Text", {
    systemChat str diag_frameNo;
    0 = 0
}, {}, [15, [false, false, false]], false] call CBA_fnc_addKeybind;
```
Before: generic error in expression
After: works, assignment treated as "false" (like any other non boolean return value).

### DIFF
--- a/addons/events/fnc_keyHandlerDown.sqf
+++ b/addons/events/fnc_keyHandlerDown.sqf
@@ -72,7 +72,7 @@ private _blockInput = false;
             _params pushBack + _keybindParams;
             _params pushBack _x;
 
-            _blockInput = ([_params call _code] param [0, false] isEqualTo true) || {_blockInput};
+            _blockInput = ([nil] apply {_params call _code} param [0, false] isEqualTo true) || {_blockInput};
         };
     };
 } forEach (GVAR(keyDownStates) param [_inputKey, []]);


addons/events/fnc_keyHandlerDown.sqf
```diff
@@ -72,7 +72,7 @@ private _blockInput = false;
             _params pushBack + _keybindParams;
             _params pushBack _x;
 
-            _blockInput = ([_params call _code] param [0, false] isEqualTo true) || {_blockInput};
+            _blockInput = ([nil] apply {_params call _code} param [0, false] isEqualTo true) || {_blockInput};
```
dedmen: `0` would be faster than `nil` but would ofc define _x, doubt that would break anyones code?

commy2: `_x` previously leaked into the user's code block alongside all the other locals like `_params`, `_code` and `_keybindParams` etc.
One could also use [_x] I suppose to keep the _x the same.
I just don't care about micro optimizations like that. The code is still a fifth of what it used to be before my rewrite and no one ever complained about it.
_x being nil was just the logical thing, as in "this variable I never defined is undefined", from the users perspective.

PR #1216, opened 2019-08-26
Add Spanish translation for UI category

jonpas:
- Add Spanish translation for UI category by @TheMagnetar


commy2:
Why not translate a bit more than that?

jonpas:
Because I am not Spanish and just copied that from ACRE2 (where it was added) as we copied the rest from CBA.

commy2:
ok

PR #1222, opened 2019-09-06
Cleanup CfgPatches and use VERSION_CONFIG consistently

jonpas:
- Cleanup all `CfgPatches` for consistency - follow order in ACE3.
- Use `VERSION_CONFIG` in all components - compliments #1221, previously only `version` was defined so Credits showed short version `v3.12`.
- Use `requiredVersion = REQUIRED_VERSION;` consistently.
- Remove now (probably) outdated `4.0.0` version from XEH component.
- Remove outdated `versionDesc`.

### DIFF
--- a/addons/main/config.cpp
+++ b/addons/main/config.cpp
@@ -37,7 +37,6 @@ class CfgPatches {
         authors[] = {};
         url = "$STR_CBA_URL";
         VERSION_CONFIG;
-        versionDesc = "C.B.A.";
     };
 };
 
--- a/addons/main/config.cpp
+++ b/addons/main/config.cpp
@@ -3,9 +3,7 @@
 // Simply a package which requires other addons.
 class CfgPatches {
     class ADDON {
-        author = "$STR_CBA_Author";
         name = CSTRING(component);
-        url = "$STR_CBA_URL";
         units[] = {};
         weapons[] = {};
         requiredVersion = REQUIRED_VERSION;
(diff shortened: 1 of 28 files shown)


addons/main/config.cpp
```diff
@@ -35,9 +33,11 @@ class CfgPatches {
             "cba_optics",
             "cba_disposable"
         };
-        versionDesc = "C.B.A.";
-        VERSION_CONFIG;
+        author = "$STR_CBA_Author";
         authors[] = {};
+        url = "$STR_CBA_URL";
+        VERSION_CONFIG;
+        versionDesc = "C.B.A.";
```
commy2:
```suggestion
        versionDesc = "$STR_CBA_Author";
```
This is weird atm.

jonpas: Agreed, remnant from CBA_A2. I would not put author there though, it's supposed to be a short description, ~~but I don't even know where it's used~~ apparently not used anywhere, just a dummy config entry.

commy2: I don't think it is even read by anything. It is probably an entry from CBA_A2 that is superseded by the `name` entry from A3 base game.

jonpas: Yeah, sounds good, I'll delete.

jonpas: Also removed wiki section on that.

commy2:
Merge now?

PR #1223, opened 2019-09-07
report backpacks with CBA_fnc_getItemConfig

commy2:
- title

Before:
```sqf
isNull ("B_AssaultPack_blk" call CBA_fnc_getItemConfig) // true
```

After:
```sqf
isNull ("B_AssaultPack_blk" call CBA_fnc_getItemConfig) // false
"B_AssaultPack_blk" call CBA_fnc_getItemConfig // bin\config.bin/CfgVehicles/B_AssaultPack_blk
```

### DIFF
--- a/addons/common/fnc_getItemConfig.sqf
+++ b/addons/common/fnc_getItemConfig.sqf
@@ -6,7 +6,7 @@ Description:
     A function used to return the config of an item.
 
 Parameters:
-    _weapon - Any kind of item, weapon or magazine class name <STRING>
+    _item - Any kind of item, weapon or magazine class name <STRING>
 
 Returns:
     _config - Item config. <CONFIG>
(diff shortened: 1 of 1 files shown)


addons/common/fnc_getItemConfig.sqf
```diff
@@ -23,16 +23,22 @@ Author:
...
 {
-    private _config = configFile >> _x >> _weapon;
+    private _config = configFile >> _x >> _item;
 
     if (isClass _config) exitWith {
         _result = _config;
     };
 } forEach ["CfgWeapons", "CfgMagazines", "CfgGlasses"];
 
+private _config = configFile >> "CfgVehicles" >> _item;
```
jonpas: Only do this entire new section if result is null still?

commy2: I suppose?

It wouldn't change the return value though.

jonpas: It would be mildly faster for non-backpack cases.

commy2: Yes, and slower for backpack classes.

jonpas: One would think that's least used.

PR #1235, opened 2019-10-01
Add JR FrontSideRail to MX

dedmen:
As per https://github.com/acemod/ACE3/issues/7182
The MX is missing the side rail, I don't know how that was not noticed before.

Also I noticed that we can directly modify in `arifle_MX_Base_F` as we already have the "WeaponSlotsInfo" declaration to inherit from, inside Rifle_Base_F. Thus we can do some optimizations here.

Same thing for `mk20_base_F`

### DIFF
--- a/addons/jr/CfgWeapons.hpp
+++ b/addons/jr/CfgWeapons.hpp
@@ -336,36 +336,6 @@ class CfgWeapons {
     };
 
     class mk20_base_F: Rifle_Base_F {
-        class WeaponSlotsInfo;
-    };
-
-    class arifle_Mk20_F: mk20_base_F {
-        class WeaponSlotsInfo: WeaponSlotsInfo {
-            class CowsSlot: asdg_OpticRail1913 {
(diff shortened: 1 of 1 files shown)


commy2:
This PR is wrong. Since 1.94, BI added their own "JR" class to the pointer and muzzle slot of the MX. Subsequently these CBA_JR classes were removed to not break compatibility with (potential) Vanilla/Non-CBA addons.

Example for modder:
```cpp
class PointerSlot;
class PointerSlot_Rail: PointerSlot {
    class compatibleItems {
        tag_banana = 1;
    };
};
```

User can get legacy CBA_JR classes for the MX and AK12 using this component: https://github.com/CBATeam/CBA_A3/pull/1189

Also this is an ongoing issue. Expect BI to add more and more "JR" classes and CBA to remove theirs.

Close?

dedmen:
So I guess the bug is in ACE then, not adding its attachments to both the JR and the Vanilla class :/

PR #1237, opened 2019-10-02
Add turret event

neilzar:
- Add a turret event that will send the event to the correct turret owner. For use with the commands that need to be run where the turret is local.

Still a bit unsure whether TUEVENT is a good choice, but I couldn't think of something else.

### DIFF
--- a/addons/events/fnc_turretEvent.sqf
+++ b/addons/events/fnc_turretEvent.sqf
@@ -16,7 +16,7 @@ Returns:
 
 Examples:
     (begin example)
-        ["test", ["target"], cursorObject, [-1]] call CBA_fnc_turretEvent;
+        ["test", ["turret"], cursorObject, [-1]] call CBA_fnc_turretEvent;
     (end)
 
 Author:
@@ -42,7 +42,7 @@ if (isServer) then {
     SEND_EVENT_TO_CLIENT(_params,_eventName,_turretOwner);
 } else {
     // only server knows object owners. let server handle the event.
-    SEND_TEVENT_TO_SERVER(_params,_eventName,_remoteTargets);
+    SEND_TUEVENT_TO_SERVER(_params,_eventName,_vehicle,_turretPath);
 };
 
 nil
--- a/addons/events/fnc_turretEvent.sqf
+++ b/addons/events/fnc_turretEvent.sqf
@@ -1,0 +1,48 @@
+#include "script_component.hpp"
+/* ----------------------------------------------------------------------------
+Function: CBA_fnc_turretEvent
+
+Description:
+    Raises a CBA event on the machine where the vehicle's turret is local.
+
+Parameters:
+    _eventName  - Type of event to publish. <STRING>
+    _params     - Parameters to pass to the event handlers. <ANY>
+    _vehicle    - Vehicle to which the turret belongs. <OBJECT>
+    _turretPath - The turret to execute on. <ARRAY>
+
+Returns:
+    None
+
+Examples:
+    (begin example)
+        ["test", ["target"], cursorObject, [-1]] call CBA_fnc_turretEvent;
+    (end)
+
+Author:
+    NeilZar
+---------------------------------------------------------------------------- */
+SCRIPT(turretEvent);
+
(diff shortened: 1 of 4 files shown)


addons/events/script_component.hpp
```diff
@@ -31,6 +31,12 @@
 
 #define SEND_TEVENT_TO_SERVER(params,name,targets) TEVENT_PVAR = [name, params, targets]; publicVariableServer TEVENT_PVAR_STR
 
+// turret events
+#define TUEVENT_PVAR CBAr
```
commy2: This is the most beautiful pull request I have ever seen, however, CBAr and CBAt are the predecessors of CBAs and CBAu and they were changed for reasons. Literally every other character is free, but not r and t.
```suggestion
#define TUEVENT_PVAR CBAv
```

neilzar: 4 characters out of 26 that were not available, I managed to pick one. 😂

commy2: That's why I told you which ones, because the next guess would've been CBAt for "turret".

addons/events/fnc_turretEvent.sqf
```diff
@@ -0,0 +1,47 @@
...
+if (_vehicle turretLocal _turretPath) exitWith {
+    CALL_EVENT(_params,_eventName);
+};
+
+if (isServer) then {
+    // retrieve the turret owner and send the event
+    private _turretOwner = _vehicle turretOwner _turretPath;
+
+    if (_turretOwner == 0) then {
+        private _turretOwner = owner _vehicle;
```
mharis001: `_turretOwner` is private to the inner scope, doesn't modify outer scope's `_turretOwner`.

neilzar: Good catch, I had something else there at first that required the private but forgot to remove it when I changed the implementation.

commy2:
When does `_turretOwner` report 0?

neilzar:
I believe when there is no unit in the turret, haven't been able to properly test it in dedicated MP though. It is done this way in both ACE and ZEN, so I assumed they have found issues with it in the past.

I would assume that turretOwner would return the vehicle owner when the turret is empty because that is the turret locality at that time, but I guess there is some weird stuff in the engine that only returns the unit in the turret as the owner and throws 0 if there is no unit.

commy2:
In my testing it worked as expected, running on the owner of the vehicle when the turret was not crewed, despite this code block being bugged. I didn't test much with the driver turret [-1] though.

I suppose this would benefit from a MP unit test script.

neilzar:
Got to test this with one other player in local hosted MP, results were as expected. Tested it with a modified version of ZEN that uses the turret event to set the ammo on a vehicle. He placed a vehicle in Zeus, and I adjusted the ammo slider. These are the results I got:

- When the vehicle was empty, it would run on his machine.
- When a unit local to him was in any spot of the vehicle, it ran on his machine.
- When a unit local to me was in any spot of the vehicle, it ran on my machine.
- When his unit was driver and mine was in a turret, the events for the turret ran on my machine, and the one for the driver ran on his. Same the other way around.

These are the results that I expected.

commy2:
A change after I approved lies in how it handles non existing turrets I believe. What would the expected behavior be for non existing turrets? Before I approved the event would trigger nowhere. Now I think it falls back to vehicle locality, but I haven't tested.
Spontaneously I would say that a non existing turret should be treated like a null object in targetEvent and raise the event nowhere (silent fail).

neilzar:
I think a silent fail would be a good solution, I'll work on implementing it.

addons/events/fnc_turretEvent.sqf
```diff
@@ -0,0 +1,52 @@
...
+
+Author:
+    NeilZar
+---------------------------------------------------------------------------- */
+SCRIPT(turretEvent);
+
+params [["_eventName", "", [""]], ["_params", []], ["_vehicle", objNull, [objNull]], ["_turretPath", [-1], [[]]]];
+
+if (_turretPath isEqualTo []) then {
+   _turretPath = [-1];
```
jonpas:
```suggestion
    _turretPath = [-1];
```

PR #1242, opened 2019-10-18
 Replace all use of count as a forEach

PabstMirror:
Trivial perf improvement, but also called a few times each frame

### DIFF
--- a/addons/statemachine/fnc_clockwork.sqf
+++ b/addons/statemachine/fnc_clockwork.sqf
@@ -23,7 +23,6 @@ SCRIPT(clockwork);
     private _stateMachine = _x;
     private _list = _stateMachine getVariable QGVAR(list);
     private _skipNull = _stateMachine getVariable QGVAR(skipNull);
-    private _updateCode = _stateMachine getVariable QGVAR(updateCode);
     private _id = _stateMachine getVariable QGVAR(ID);
     private _tick = _stateMachine getVariable QGVAR(tick);
 
@@ -36,6 +35,7 @@ SCRIPT(clockwork);
 
     // When the list was iterated through, jump back to start and update it
     if (_tick >= count _list) then {
+        private _updateCode = _stateMachine getVariable QGVAR(updateCode);
         _tick = 0;
         if !(_updateCode isEqualTo {}) then {
             _list = [] call _updateCode;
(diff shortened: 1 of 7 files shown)


addons/xeh/fnc_preInit.sqf
```diff
@@ -63,11 +63,11 @@ GVAR(EventsLowercase) = [];
         FUNC(Init) = compileFinal (_header + "(_this select 0) call CBA_fnc_initEvents; (_this select 0) call CBA_fnc_init");
     } else {
         if (_x isEqualTo "HitPart") then {
-            FUNC(HitPart) = compileFinal (_header + format ['{call _x; nil} count ((_this select 0 select 0) getVariable QGVAR(%1))', _x]);
+            FUNC(HitPart) = compileFinal (_header + format ['{call _x} forEach ((_this select 0 select 0) getVariable QGVAR(%1))', _x]);
         } else {
             missionNamespace setVariable [
                 format [QFUNC(%1), _x],
-                compileFinal (_header + format ['{call _x; nil} count ((_this select 0) getVariable QGVAR(%1))', _x])
+                compileFinal (_header + format ['{call _x} forEach ((_this select 0) getVariable QGVAR(%1))', _x])
```
PabstMirror: @commy2 - Is this safe, or do we have to worry about the return type changing? 
forEach returns last run
instead of `nil } count` always returns 0

commy2: No, this is fine.

addons/statemachine/fnc_clockwork.sqf
```diff
@@ -36,6 +35,7 @@ SCRIPT(clockwork);
 
     // When the list was iterated through, jump back to start and update it
     if (_tick >= count _list) then {
+        private _updateCode = _stateMachine getVariable QGVAR(updateCode);
```
commy2: Can we be sure this variable doesn't change between function start and this, or between iterations?

dedmen: This is unscheduled afaik?
And there is nothing that executes user defined code between old L26 and the new L38.
So I'd call that safe.

addons/xeh/fnc_startFallbackLoop.sqf
```diff
@@ -28,8 +28,7 @@ GVAR(fallbackRunning) = true;
 {
     // don't run init and initPost event handlers on objects that already exist
     SETINITIALIZED(_x);
```
dedmen: SETINITIALIZED is a setVariable which returns nil, which would make the count work here.

commy2: As this is something that runs every frame and the array can be big in certain missions, I agree that it can use count loop here.

PR #1243, opened 2019-10-19
Attack Module Override enables AI pathing

zanven42:
Issue: when a group is set to defend with a high percentage of "hold" and a trigger activates a sync'd attack module which wants all units on defend to attack a location the behaviour currently is that all units will have the new waypoint and all units with pathing disabled because of "holding" will stand still which presents a inflexible scenario without custom logic elements to enable good timing of releasing "waves" of units as a Zeus.

Proposal: Enable the override parameter to re enable pathing so it can override any blocking previous command given to the unit such as the "defend module's hold". This should make that option act more as expected by overriding previous orders and allow more dynamic options for mission makers.

### DIFF
--- a/addons/ai/fnc_taskAttack.sqf
+++ b/addons/ai/fnc_taskAttack.sqf
@@ -34,6 +34,7 @@ if !(local _group) exitWith {}; // Don't create waypoints on each machine
 // Allow TaskAttack to override other set waypoints
 if (_override) then {
     [_group] call CBA_fnc_clearWaypoints;
+    { _x enableAI "PATH"; } foreach (units _group);
 };
 
 [_group, _position, _radius, "SAD", "COMBAT", "RED"] call CBA_fnc_addWaypoint;
--- a/addons/ai/fnc_taskAttack.sqf
+++ b/addons/ai/fnc_taskAttack.sqf
@@ -34,7 +34,9 @@ if !(local _group) exitWith {}; // Don't create waypoints on each machine
 // Allow TaskAttack to override other set waypoints
 if (_override) then {
     [_group] call CBA_fnc_clearWaypoints;
-    { _x enableAI "PATH"; } foreach (units _group);
+    {
+        _x enableAI "PATH";
+    } forEach units _group;
 };
 
 [_group, _position, _radius, "SAD", "COMBAT", "RED"] call CBA_fnc_addWaypoint;


commy2:
>when a group is set to defend with a high percentage of "hold" and a trigger activates a sync'd attack module which wants all units on defend to attack a location the behaviour currently is that all units will have the new waypoint and all units with pathing disabled because of "holding" will stand still

To clarify why this fix is specifically about PATH and not MOVE or AUTOTARGET etc, @zanven42 is refering to this line:
https://github.com/CBATeam/CBA_A3/blob/master/addons/ai/fnc_taskDefend.sqf#L110
```sqf
                    if (random 1 < _hold) then {
                        _unit disableAI "PATH";
                    } else {
                        doStop _unit;
                    };
```

Basically taskAttack and taskDefend should play nice together.

The Attack module forwards to the task function, but that place probably also needs some fix, because currently it does not use the _override flag:
https://github.com/CBATeam/CBA_A3/blob/master/addons/modules/fnc_moduleAttack.sqf

Question is, should PATH not also be reenabled for taskPatrol, taskSearchArea and taskDefend itself?
https://github.com/CBATeam/CBA_A3/blob/master/addons/ai/fnc_taskDefend.sqf#L55
https://github.com/CBATeam/CBA_A3/blob/master/addons/ai/fnc_taskPatrol.sqf#L51
https://github.com/CBATeam/CBA_A3/blob/master/addons/ai/fnc_taskSearchArea.sqf#L56

addons/ai/fnc_taskAttack.sqf
```diff
@@ -34,6 +34,7 @@ if !(local _group) exitWith {}; // Don't create waypoints on each machine
 // Allow TaskAttack to override other set waypoints
 if (_override) then {
     [_group] call CBA_fnc_clearWaypoints;
+    { _x enableAI "PATH"; } foreach (units _group);
```
commy2: formatting
```suggestion
    {
        _x enableAI "PATH";
    } forEach units _group;
```

zanven42:
yeah it probably should, i only encountered the issue a few hours before this PR as i am only using the Defend and attack modules intensely for a mission i am making and wanted them to play nice.

Another question is should override be required to take people off defend onto the other tasks. I lean towards not required but to be safe i added it to the override section to maintain backwards expectations.

commy2:
I agree. Overwrite should not be required, or at least be the default behavior. The other task functions should re-enable PATH at the points where they clear the waypoints.

commy2:
State? Or you want me to finish it?

PR #1244, opened 2019-10-23
Add CBA_fnc_selectBest function

PabstMirror:
Select best element from an array

Alternate way, but I found no real speed diff,
```
_array = _array apply {  [_x call _criteria, _x]; };
_array sort false; 
_array # 0 # 1
```

### DIFF
--- a/addons/arrays/fnc_selectBest.sqf
+++ b/addons/arrays/fnc_selectBest.sqf
@@ -1,0 +1,40 @@
+#include "script_component.hpp"
+/* ----------------------------------------------------------------------------
+Function: CBA_fnc_selectBest
+
+Description:
+    Select best element from an array
+
+Parameters:
+    _array - Input Array [Array]
+    _criteria - Code that is passed element and should return an integer value [Code]
+    _return - default return value if array is empty (optional, default nil) [Any]
+
+Returns:
+    Element that scores the highest [Any]
+
+Example:
+    (begin example)
+        _result = [[2, -6, 4], {abs _x}] call CBA_fnc_selectBest;
+        // _result => -6
+        _result =  [[q1, q2], {-(t1 distance _x)}, objNull] call CBA_fnc_selectBest; // selects closest target (negative distance)
+        // _result => q1
+    (end)
+
+Author:
+    PabstMirror
+---------------------------------------------------------------------------- */
+SCRIPT(selectBest);
+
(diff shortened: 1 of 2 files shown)


dedmen:
There should be a speed diff tho 🤔 if/then evaluation and opening a new scope every iteration should be more expensive than creating an array and running a sort.

commy2:
I like the alternative in the opening post more for reability alone.
Last line could be:
`_array param [0, [0, _default]] select 1`
and so handle the default value and the whole script would be 4 lines.

PabstMirror:
```
a = {  
    params ["_array", "_criteria", "_return"]; 
    private _bestScore = -1e99; 
    { 
        private _xScore = _x call _criteria; 
        if (_xScore > _bestScore) then { 
            _return = _x; 
            _bestScore = _xScore; 
        }; 
    } forEach _array; 
    _return 
};  
b = {  
    params ["_array", "_criteria", "_default"];  
    _array = _array apply { [_x call _criteria, _x]; };  
    _array sort false;  
    _array param [0, [0, _default]] select 1 
};  
x = [1,2,3,4,5,6,7,8,9,-10];  
c = { abs _this}; 
 
Result: 0.264271 ms
10x [x,c] call a; 

Result: 0.28169 ms
10x [x,c] call b; 
```
Alt way has problems with undefined in scheduled
```
[] spawn {[[],{_x}] call b}
Error Undefined variable in expression: _default
```
You also get `_forEachIndex` with a

commy2:
RETNIL(_var) macro to get around the `_default is nil` problem.

addons/arrays/fnc_selectBest.sqf
```diff
@@ -0,0 +1,40 @@
...
+private _bestScore = -1e99;
+{
+    private _xScore = _x call _criteria;
+    if (_xScore > _bestScore) then {
+        _return = _x;
+        _bestScore = _xScore;
+    };
+} forEach _array;
+
+_return
```
commy2:
```suggestion
RETNIL(_return)
```

commy2:
One problem I have with this is that the variables `_bestScore, __xScore, _array, _criteria, _return` leak into the code block.

PabstMirror:
Just to note, criteria function can return special value of -1e99 and item will be ignored 
(e.g. will return default value if no item returns a value higher)

PR #1248, opened 2019-10-31
Extension Callbacks - Events

### LINKED ISSUE #1196: Feature Request: ExtensionCallback helpers
With the addition of Extension Callbacks in 1.95 it would be nice to have a set of CBA helper functions exposed to extensions to reduce rewritten code.

Each event has a name, function, and args.

There could be things like
`cba` `hint`
`cba` `notify`
and more that would make it easy to use common functions from extensions without needing to re-implement them for your addon.

synixebrett:
- Close #1196

This will allow extension developers to use these functions easily without creating an unnecessary additional `addMissionEventHandler`

### DIFF
--- a/addons/events/XEH_postInit.sqf
+++ b/addons/events/XEH_postInit.sqf
@@ -86,3 +86,32 @@ if (isServer) then {
         GVAR(triggerPressed) = false;
     };
 }] call CBA_fnc_addDisplayHandler;
+
+addMissionEventHandler ["ExtensionCallback", {
+	params ["_name", "_function", "_data"];
+
+    if !(_name isEqualTo "cba_events") exitWith {};
+
+    private _data = parseSimpleArray _data;
+    _data params ["_event", "_params", "_targets"];
+    switch (_function) do {
+        case "localEvent": {
+            [_event, _params] call CBA_fnc_localEvent;
+        };
+        case "targetEvent": {
+            [_event, _params, _targets] call CBA_fnc_targetEvent;
+        };
+        case "serverEvent": {
+            [_event, _params] call CBA_fnc_serverEvent;
+        };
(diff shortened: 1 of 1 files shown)


PabstMirror:
long enought that it might be worth to put in func 
if bis eventHandlers are still compiled each time?

dedmen:
> if bis eventHandlers are still compiled each time?
not in 1.96 with script caching, but the memory leak is back and the caching will be removed again.
So yes.

synixebrett:
By created, a mean writing code, not created in the game engine.

Without this you might have something like
```sqf
addMissionEventHandler ["ExtensionCallback", {
  params ["_name", "_function", "_data"];
  if (_name isEqualTo "myCoolMod") then {
    if (_function isEqualTo "sendEvent") then {
      ["myCoolModEvent", _data] call CBA_fnc_localEvent;
    };
  };
}];
```

If this PR was part of CBA you would need that at all, since you could just use the `CBA` `localEvent` extension callback. From the engine perspective an `addMissionEventHandler` is being compiled either way.

commy2:
And? That seems simple enough to me.

synixebrett:
If more mods were doing that there would be more `addMissionEventHandler`s compared to just using the CBA one. This may not be the most important thing to add to CBA but it's not like it is hurting anything. Just makes things a little bit simpler for extension developers

PabstMirror:
```
cba_events_fnc_addExtensionCallback = {
    if (missionNamespace getVariable [QGVAR(extCallbackAdded), false]) exitWith {};
    diag_log "[CBA] Adding Extension callback handler";
    GVAR(extCallbackAdded) = true;
    ...
};
```
Can be called zero or multiple times, and code will be added only once and only if needed

PR #1253, opened 2019-11-15
Fix JIP event removal on dedicated client

mharis001:
- Fix `CBA_fnc_removeGlobalEventJIP` not correctly removing events when called in "wait for object deletion" mode on non-server machine when the object is deleted after the client disconnects
- "Deleted" EH won't exist after the client disconnects and as such the event is never removed, unless added on server


commy2:
3.13?
Did you actually run into an issue with this?

mharis001:
Yes, nothing major. Just code that would silently fail when given `objNull`, this will ensure it doesn't run at all.

Issue #1234, opened 2019-10-01
cba optics and using Arsenal in Zeus don't return camera to its original position


bux:
**Mods (complete and add to the following information):**
- **Arma 3:** stable
- **CBA:** `3.12.2`
- **BWMod:** stable

**Description:**
Redirected here by @commy2 
When adding BWMod to the mod list closing the Zeus Arsenal doesn't return the camera to the position when using the Arsenal module on an editable unit but to the first person view of either the unit one is working on or the player unit.

**Steps to reproduce:**
- Add player to a mission
- Add game master module and sync to player
- Start mission
- Open Zeus and place a new unit
- Use the Arsenal module to open Arsenal of newly placed unit
- Close the Arsenal

**Expected behavior:**
Camera returns to the last position where one opened the Zeus Arsenal instead of first person view of a unit

**Where did the issue occur?**
- Editor SP

**Log Files:**
https://pastebin.com/Uddqdx1c

**Additional context:**

**Screenshots:**

Blutze:
Does NOT replicate with [RKSL Attachments](https://steamcommunity.com/sharedfiles/filedetails/?id=1661066023) instead of BWMod, so just getting scripted optics to init is not enough. Maybe the Brits are doing something else janky, or there are remnants of the old BWMod-only implementation.

commy2:
Zeus Arsenal.
What about a regular Mission Arsenal?
What about the 3den Arsenal?
What about the Arsenal Mission from main menu?

Blutze:
> What about a regular Mission Arsenal?
> What about the 3den Arsenal?
> What about the Arsenal Mission from main menu?

Those do not have this issue.

Just remembered I still have [this](https://gist.github.com/Blutze/ac4c972111676349636e039dfaf9a92c). Pay attention to the date and versions, this is before scripted optics got ported to CBA I think (all three mods were Steam Workshop downloads, up to date at that time).

bux:
Any more info needed?

commy2:
No, I just need to find the time to look at it.

commy2:
(Mostly) fixed in a commit that will appear here: https://git.koffeinflummi.de/bwmod-team/bwa3-public/commits/master

commy2:
https://git.koffeinflummi.de/bwmod-team/bwa3-public/commit/f93693a62948018b806586e1df81c90daad829c4

PR #1254, opened 2019-11-16
Fix optics camera restart handling for Zeus

mharis001:
- Fix camera restart preventing returning to `curatorCamera` when using arsenal through Zeus (module)
- Related to #1234

### DIFF
--- a/addons/optics/script_component.hpp
+++ b/addons/optics/script_component.hpp
@@ -1,9 +1,9 @@
-﻿#define COMPONENT optics
+#define COMPONENT optics
 #include "\x\cba\addons\main\script_mod.hpp"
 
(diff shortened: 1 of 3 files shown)


addons/optics/script_component.hpp
```diff
@@ -1,9 +1,9 @@
-﻿#define COMPONENT optics
+#define COMPONENT optics
```
commy2: ?

jonpas: Probably removal of BOM.

mharis001: Yes, removed BOM.

commy2: Removal is good.

PR #1259, opened 2019-11-29
add\update Tchinese Translation

dfr1238:
add some missing string on tchinese


commy2:
This is the last missing entry: https://github.com/CBATeam/CBA_A3/blob/master/addons/optics/stringtable.xml#L5-L11

PR #1263, opened 2019-12-10
FSM Editor Config for CBA Statemachine

Freddo3000:
- Add a FSM Editor config file that allows it to create and edit CBA Statemachine configs

<details>
<summary>Included instructions</summary>

```
CBA Statemachine Config for use with FSM Editor

Description:
    This is a config that allows you to create and edit CBA Statemachines using
    the Arma Tools FSM editor, for increased readability and expandability.

    It isn't by any means perfect, but it should be useable
    
Usage:
    1. Select this config using FsmAttributes > Compile Config
    2. Change FSM attributes under FsmAttributes > Attributes from Compile Config
        2.1 'list' parameter is parsed as list = QUOTE(x);
        2.2 'skipNull' parameter is parsed as skipNull = x;
    3. States
        3.1 'initCode' is parsed as is, and may contain onState = "", 
            onStateEntered = "" and/or onStateLeaving = ""
    4. Transitions
        4.1 'targetState' does not need to be entered, and simply follows the
            transition link
        4.2 'condition' is parsed as condition = QUOTE(x)
        4.3 'action' is parsed as is, and may contain onTransition = "" 
            and/or events[] = {}
			
Dependencies:
	In order to properly function, script_macros_common need to be included
Author:
    Freddo 2019-12-10
```
</details>

Example FSM: 
fsm_editor

Same FSM but as text:
https://gist.github.com/Freddo3000/244d94c0056f2b4afb0f52fec73a57a2

Once ran through pboProject:
https://gist.github.com/Freddo3000/eab4a2afd7a4ffcbc688bcf37ce7e91d

While it could be prettier, it does its job well enough.

### DIFF
--- a/addons/statemachine/CBA_FSMEditor.cfg
+++ b/addons/statemachine/CBA_FSMEditor.cfg
@@ -1,0 +1,103 @@
+/* ----------------------------------------------------------------------------
+CBA Statemachine Config for use with FSM Editor
+
+Description:
+    This is a config that allows you to create and edit CBA Statemachines using
+    the Arma Tools FSM editor, for increased readability and expandability.
+	
+	It isn't by any means perfect, but it should be useable
+    
+Usage:
+    1. Select this config using FsmAttributes > Compile Config
+    2. Change FSM attributes under FsmAttributes > Attributes from Compile Config
+        2.1 'list' parameter is parsed as list = QUOTE(x);
+        2.2 'skipNull' parameter is parsed as skipNull = x;
+    3. States
+        3.1 'initCode' is parsed as is, and may contain onState = "", 
+            onStateEntered = "" and/or onStateLeaving = ""
+    4. Transitions
+        4.1 'targetState' does not need to be entered, and simply follows the
+            transition link
+        4.2 'condition' is parsed as condition = QUOTE(x)
+        4.3 'action' is parsed as is, and may contain onTransition = "" 
+            and/or events[] = {}
+			
+Dependencies:
+	In order to properly function, script_macros_common need to be included
+
+Author:
+    Freddo 2019-12-10
+---------------------------------------------------------------------------- */
+
+class Attributes
+{
+  names[] = {"list", "skipNull"};
+};
+
+class Compile
+{
+    clearNewLines = 0;
+//  indenting = -1;  //no indenting
+    rewriteFile = 1; //do not append result to the end of file, but rewriteIt
+    class Pass_States
+    {
+        print_sm =           "class %(fsmname)\n{\n";
+        indent_smOpen =      4;
+        print_attrList =     "list = QUOTE(%(list));\n";
+        print_attrSkipNull = "skipNull = %(skipNull);\n";
+        class State_Any
+        {
+            print_state =          "class %(statename)\n{\n";
+			indent_stateContents = 0; //Neccessary to preserve formatting
+            print_stateContents =  "%(stateinit)\n";
+			indent_stateOpen =     8;
+            class Links
+            {
+                    print_transition =          "class %(linkname)\n{\n";
+                    indent_transitionOpen =     12;
+                    print_target =             "targetState=QUOTE(%(to));\n";
+                    print_condition =           "condition=QUOTE(%(condition));\n";
+					indent_transitionContents = 0; //Neccessary to preserve formatting
+                    print_transitionContents =  "%(action)\n";
+                    indent_transitionClose =    8;
+                    print_transitionClose =     "};\n";
+            }
+            indent_stateClose = 4;
+            print_stateClose =  "};\n";
+        }
+        indent_smClose = 0;
+        print_smClose =  "};\n";
+    };
+};
+
+class Decompile
+{
+    process =         1; //0 for not processing Decompile info
+    FSMLeft =      "/*";
+    FSMRight =     "*/";
+    class FSMPrefix
+    {
+        default =    "";
+    };
+    class FSMPrefix2
+    {
+        default =    "";
+        head =     "*/";
+    };
+    class FSMSufix
+    {
+        default =    "";
+        compile =  "\n";
+        state =    "\n";
+        head = "\n/*\n";
+        link =     "\n";
+    };
+    class FSMSufix2
+    {
+        default =    "";
+        compile =    "";
+        state =    "\n";
+        head =     "\n";
+        link =     "\n";
+    };
+};
--- a/addons/statemachine/CBA_FSMEditor.cfg
+++ b/addons/statemachine/CBA_FSMEditor.cfg
@@ -4,96 +4,86 @@ CBA Statemachine Config for use with FSM Editor
 Description:
     This is a config that allows you to create and edit CBA Statemachines using
     the Arma Tools FSM editor, for increased readability and expandability.
-	
-	It isn't by any means perfect, but it should be useable
-    
+
+    It isn't by any means perfect, but it should be useable
+
 Usage:
     1. Select this config using FsmAttributes > Compile Config
-    2. Change FSM attributes under FsmAttributes > Attributes from Compile Config
+    2. Change FSM attributes under FsmAttributes > Attributes from Compile Config.
         2.1 'list' parameter is parsed as list = QUOTE(x);
         2.2 'skipNull' parameter is parsed as skipNull = x;
     3. States
-        3.1 'initCode' is parsed as is, and may contain onState = "", 
+        3.1 'initCode' is parsed as is, and may contain onState = "",
             onStateEntered = "" and/or onStateLeaving = ""
     4. Transitions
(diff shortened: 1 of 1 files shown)


addons/statemachine/CBA_FSMEditor.cfg
```diff
@@ -0,0 +1,103 @@
...
+			
+Dependencies:
+	In order to properly function, script_macros_common need to be included
+
+Author:
+    Freddo 2019-12-10
+---------------------------------------------------------------------------- */
+
+class Attributes
+{
```
commy2: Does it have to have open curly bracket after newline? Goes against the rest of the code base.

addons/statemachine/CBA_FSMEditor.cfg
```diff
@@ -0,0 +1,103 @@
...
+            transition link
+        4.2 'condition' is parsed as condition = QUOTE(x)
+        4.3 'action' is parsed as is, and may contain onTransition = "" 
+            and/or events[] = {}
+			
+Dependencies:
+	In order to properly function, script_macros_common need to be included
+
+Author:
+    Freddo 2019-12-10
```
commy2: No reason to put the date here. History is saved on git, not the files. Would keep those things apart.

addons/statemachine/CBA_FSMEditor.cfg
```diff
@@ -0,0 +1,103 @@
...
+            onStateEntered = "" and/or onStateLeaving = ""
+    4. Transitions
+        4.1 'targetState' does not need to be entered, and simply follows the
+            transition link
+        4.2 'condition' is parsed as condition = QUOTE(x)
+        4.3 'action' is parsed as is, and may contain onTransition = "" 
+            and/or events[] = {}
+			
+Dependencies:
+	In order to properly function, script_macros_common need to be included
```
commy2: This is a tab?
```suggestion
    In order to properly function, script_macros_common need to be included.
```

addons/statemachine/CBA_FSMEditor.cfg
Freddo3000: I were simply trying to follow the template here: https://github.com/CBATeam/CBA_A3/wiki/Documenting-SQF-functions
```
Author:
    Spooner 2009-01-05
```

addons/statemachine/CBA_FSMEditor.cfg
```diff
@@ -0,0 +1,103 @@
...
+    1. Select this config using FsmAttributes > Compile Config
+    2. Change FSM attributes under FsmAttributes > Attributes from Compile Config
+        2.1 'list' parameter is parsed as list = QUOTE(x);
+        2.2 'skipNull' parameter is parsed as skipNull = x;
+    3. States
+        3.1 'initCode' is parsed as is, and may contain onState = "", 
+            onStateEntered = "" and/or onStateLeaving = ""
+    4. Transitions
+        4.1 'targetState' does not need to be entered, and simply follows the
+            transition link
```
commy2:
```suggestion
            transition link.
```

addons/statemachine/CBA_FSMEditor.cfg
```diff
@@ -0,0 +1,103 @@
...
+
+Description:
+    This is a config that allows you to create and edit CBA Statemachines using
+    the Arma Tools FSM editor, for increased readability and expandability.
+	
+	It isn't by any means perfect, but it should be useable
+    
+Usage:
+    1. Select this config using FsmAttributes > Compile Config
+    2. Change FSM attributes under FsmAttributes > Attributes from Compile Config
```
commy2:
```suggestion
    2. Change FSM attributes under FsmAttributes > Attributes from Compile Config.
```

addons/statemachine/CBA_FSMEditor.cfg
```diff
@@ -0,0 +1,103 @@
...
+        print_sm =           "class %(fsmname)\n{\n";
+        indent_smOpen =      4;
+        print_attrList =     "list = QUOTE(%(list));\n";
+        print_attrSkipNull = "skipNull = %(skipNull);\n";
+        class State_Any
+        {
+            print_state =          "class %(statename)\n{\n";
+			indent_stateContents = 0; //Neccessary to preserve formatting
+            print_stateContents =  "%(stateinit)\n";
+			indent_stateOpen =     8;
```
commy2: Tabs.

addons/statemachine/CBA_FSMEditor.cfg
```diff
@@ -0,0 +1,103 @@
...
+    class Pass_States
+    {
+        print_sm =           "class %(fsmname)\n{\n";
+        indent_smOpen =      4;
+        print_attrList =     "list = QUOTE(%(list));\n";
+        print_attrSkipNull = "skipNull = %(skipNull);\n";
+        class State_Any
+        {
+            print_state =          "class %(statename)\n{\n";
+			indent_stateContents = 0; //Neccessary to preserve formatting
```
commy2: Tabs.

addons/statemachine/CBA_FSMEditor.cfg
commy2: Yeah, we don't do that. I changed the wiki.

addons/statemachine/CBA_FSMEditor.cfg
```diff
@@ -0,0 +1,103 @@
...
+            print_stateContents =  "%(stateinit)\n";
+			indent_stateOpen =     8;
+            class Links
+            {
+                    print_transition =          "class %(linkname)\n{\n";
+                    indent_transitionOpen =     12;
+                    print_target =             "targetState=QUOTE(%(to));\n";
+                    print_condition =           "condition=QUOTE(%(condition));\n";
+					indent_transitionContents = 0; //Neccessary to preserve formatting
```
commy2: Tabs.

addons/statemachine/CBA_FSMEditor.cfg
```diff
@@ -0,0 +1,103 @@
...
+                    print_transitionContents =  "%(action)\n";
+                    indent_transitionClose =    8;
+                    print_transitionClose =     "};\n";
+            }
+            indent_stateClose = 4;
+            print_stateClose =  "};\n";
+        }
```
commy2: Should have semi colon here.

commy2:
For the end user, why is the `CBA_FSMEditor.cfg` file inside the `addons\cba_statemachine.pbo` file?
That means they have to extract the PBO first to get it.

Should this file instead be on the wiki? If it is included into the release, should it not instead be inside `optionals` or some new folder?

Freddo3000:
The reason I placed it here was due to the statemachine [Example.sqf](https://github.com/CBATeam/CBA_A3/blob/master/addons/statemachine/example.sqf) and [Example.hpp](https://github.com/CBATeam/CBA_A3/blob/master/addons/statemachine/example.hpp) being present here. I'm fine with placing it anywhere though.

commy2:
@BaerMitUmlaut ? ^

PiZZAD0X:
Great job, I wasn't aware this was possible with the BI FSM editor! I intend to use this config structure to build some more complex AI systems in a mod using the great state machine system set up by @BaerMitUmlaut. The two examples of the standard SM are shown on the wiki and various biforums posts so I think the config should be available via a wiki link in the hope of encouraging more regular usage of the CBA SM.

PiZZAD0X:
I have tweaked this config with the intention of publishing configs directly from the FSMEditor. Saving in the .bifsm format preserves the FSM data for use in the editor, while saving the statemachine in the .fsm aka compiled format produces a CBA SM config. 

With this tweaked config: https://gist.github.com/PiZZAD0X/3b4ec1ae33cdbc76b3dc2626a5559772

A .bifsm of https://gist.github.com/PiZZAD0X/0d656d5a22453ec6c36b5340d4a184bb
produces: https://gist.github.com/PiZZAD0X/0cc91850e2f3df40571f1ecc0d562d87 when compiled.

commy2:
Question:
```sqf
    class CombatStance {
        onState = QUOTE(
            DEBUG_TAG('Combat stance');
            // Don't overide explicitly set stance
            private _hasStance = _this getVariable [QQGVAR(hasStance), false];
            if (!_hasStance && toLower (unitPos _this) != 'auto') exitWith {
                DEBUG_TAG('Skip stance');
            };
```

Does this work? There are newlines inside the macro (QUOTE()) argument. This should fail.

PiZZAD0X:
Yes it does, but it works just fine throwing in the necessary \ Parser works fine with the \

Isn't best practice to throw these to a QFUNC instead of commands inside a quote inside config?

commy2:
Yeah, excessive script in config is bad, regardless of this.

commy2:
I added a barebone doc page for the statemachine that includes this:
https://github.com/CBATeam/CBA_A3/wiki/Statemachine

Issue #1266, opened 2019-12-17
JSON parse


ColdEvul:
**Is your feature request related to a problem?**
No

**Solution you'd like:**
I whould like the json parser functions from ALIVE to be integrated in to cba. Today a YAML parser exist but one for JSON format whould also be greate. 
Function links below:
- https://github.com/ALiVEOS/ALiVE.OS/blob/4f7ac35a2cc6b257a795eed8d2fbe52e07913c51/addons/x_lib/functions/data/fnc_parseJSON.sqf
- https://github.com/ALiVEOS/ALiVE.OS/blob/4f7ac35a2cc6b257a795eed8d2fbe52e07913c51/addons/x_lib/functions/data/fnc_JSON.sqf
- https://github.com/ALiVEOS/ALiVE.OS/blob/4f7ac35a2cc6b257a795eed8d2fbe52e07913c51/addons/x_lib/functions/data/fnc_encodeJSON.sqf

**Alternatives you've considered:**
Contact maintainers of ALIVE is for rewire and copy paste or rewrite the YAML one. 

**Additional context:**
None

commy2:
Well, am I allowed to just copy these? They look pretty nice aside from small imperfections, mostly newlines and legacy `private array` usage. Can't say if they work as expected though, because idk anything about JSON.

ColdEvul:
`JSON` is basicly `YAML` with `{}`

```
     "firstName": "Jason",
     "lastName": "Smith",
     "age": 25,
     "address": {
         "streetAddress": "21 2nd Street",
         "city": "New York",
         "state": "NY",
         "postalCode": "10021"
     },
     "phoneNumber": [
         { "type": "home", "number": "212 555-1234" },
         { "type": "fax", "number": "646 555-4567" }
     ],
     "newSubscription": false,
     "companyName": null
 }
```

bux:
but indentation is not forced as in yaml.

```json
{"firstname":"Jason","lastName":"Smith"}
```

PR #1269, opened 2019-12-21
add CBA_fnc_parseJSON and CBA_fnc_encodeJSON ports from ALiVE

commy2:
- close https://github.com/CBATeam/CBA_A3/issues/1266


commy2:
Current:
```sqf
_hash = [loadFile "\x\cba\addons\hashes\test_parseJSON_config.json"] call CBA_fnc_parseJSON;

[_hash, "
    firstName"] call CBA_fnc_hashGet // " Jason"
```
So it has problems with white space apparently.

commy2:
Baer wrote a parser that works better (at all?) compared to the ports.
superseded by #1272

PR #1238, opened 2019-10-03
Filtered getUnitLoadout

synixebrett:
**When merged this pull request will:**
- Close #1236 
- `CBA_fnc_getUnitLoadout`
- `CBA_fnc_filterLoadout`
- `CBA_fnc_addLoadoutFilter`
- `CBA_fnc_removeLoadoutFilter`

```sqf
private _loadout = [player] call CBA_fnc_getUnitLoadout; // Loadout with ItemRadio

private _handle = [{ 
    params ["_loadout", "_handle"]; 
    if ((_loadout select 9) select 2 == "ItemRadio") then { 
        (_loadout select 9) set [2, ""]; 
    }; 
    _loadout
}] call CBA_fnc_addLoadoutFilter;

_loadout = [_loadout] call CBA_fnc_filterLoadout; // Previous loadout with ItemRadio removed
_loadout = [player] call CBA_fnc_getUnitLoadout; // Loadout without ItemRadio

[_handle] call CBA_fnc_removeLoadoutFilter;

_loadout = [player] call CBA_fnc_getUnitLoadout; // Loadout with ItemRadio
```

PR #1270, opened 2019-12-21
Filtered Loadouts

### LINKED ISSUE #1236: Filtered getUnitLoadout
**Is your feature request related to a problem?**
Some mods (TFAR, ACRE, Unique Maps) use an ID system to track individual items. This will cause issues with `getUnitLoadout` and `setUnitLoadout`.

**Solution you'd like:**
`CBA_fnc_getUnitLoadout` that is a drop in replacement for its vanilla counterpart. Also create `CBA_fnc_addUnitLoadoutFilter` and `CBA_fnc_removeUnitLoadoutFilter` that take code as a parameter, that code takes a `getUnitLoadout` array and must return a `getUnitLoadout` array. The code will be used to replace any ID items with their base version.

**Additional context:**
[ACRE Example](https://github.com/IDI-Systems/acre2/pull/848/files#diff-b772f408a03abd22c2dab22d1dfc5ba2)

BrettMayson:
- Alternate to #1238 
- Close #1236
- `CBA_fnc_getUnitLoadout`
- `CBA_fnc_filterLoadout`
- `CBA_fnc_addLoadoutFilter`
- `CBA_fnc_removeLoadoutFilter`

```sqf
private _loadout = [player] call CBA_fnc_getUnitLoadout; // Loadout with ItemRadio

private _handle = [{ 
    params ["_loadout", "_handle"]; 
    if ((_loadout select 9) select 2 == "ItemRadio") then { 
        (_loadout select 9) set [2, ""]; 
    }; 
    _loadout
}] call CBA_fnc_addLoadoutFilter;

_loadout = [_loadout] call CBA_fnc_filterLoadout; // Previous loadout with ItemRadio removed
_loadout = [player] call CBA_fnc_getUnitLoadout; // Loadout without ItemRadio

[_handle] call CBA_fnc_removeLoadoutFilter;

_loadout = [player] call CBA_fnc_getUnitLoadout; // Loadout with ItemRadio
```


commy2:
Explain the purpose of this in CBA in 500 words or less.

synixebrett:
# Filtered Loadouts: A Case For

Loadouts are an integral aspect of the Arma 3 gameplay, an aspect that players of any game mode are familiar. `getUnitLoadout` is a command many are familiar with due to its importance in persistent missions, loadout management, dynamic role selection, and overall utility. The inclusion in CBA would remedy the downfalls of current solutions and benefit the entire community. This implementation would be a small change for a lot of lasting good.

## Persistent Missions

There are several communities and game modes that take advantage of persistence spanning multiple sessions. These missions rely on utilities like `getUnitLoadout` to get the information they need. Unfortunately, programmers can not prepare for every case, we are human. These persistent missions do not always have the proper protections and systems in place to properly handle ID'd items such as those found in the popular ACRE, TFAR, and ACRE modifications. This can pose a threat to stability when duplicate instances of items can be present in a single session. This threat to stability can be easily neutralized by the implementation of loadout filtering at save and load.

## Loadout Management

Various popular modifications have implemented various forms of loadout management. Take ACE 3's *ACE Arsenal*, even such a well developed and managed community project suffers from the inability to properly handle ID'd items. Loadouts are saved, loaded, and exported with ID'd items causing potential instability and unexpected behavior. Filtered loadouts specified by mod makers could assure that these issues are avoided both during saving and loading, instantly resolving any previous issues as poisoned loadouts are repaired during load.

## Dynamic Role Selection

Game modes that implement dynamic role selection may rely on user-generated or config loadouts. These loadouts can change over time introducing potential mission breaking changes. Imagine a mission based on inserting, retrieving an item, and extracting back to the start. This could be a potential issue as future loadouts or config changes by additional modifications could result in missions ending instantly if players receive a loadout containing the item they are intended to retrieve. While this scenario is less likely that the two previously discussed, it is still an element that will be improved by the addition of filtered loadouts.

## Conclusion

Filtered loadouts may not as sexy as weapon optics or face wear randomization in the eyes of everyone, but they provide another tool in the toolbox that creators can use to increase dependability, durability, persistence, permanence, perseverance, stableness, cohesion, constancy, stability, and functionality. The simplicity of implementation for creators compared to current solutions being created will allow these benefits to be realized with ease.

Arma has a rich community of creators, builders, and dreamers. Adding this utility to the already vast abilities of CBA would unlock more possibilities than this amazing project already has. It would be impossible to cover the potential of this amazing community in such a brief analysis.

Historically CBA has stood for stability, functionality, and user experience. Filtered loadouts stand for the same three principals. CBA should stand with filtered loadouts.

mharis001:
This will definitely be a useful addition, however, I think this can be simplified to just two functions:

`CBA_fnc_addLoadoutFilter`

- Adds a new loadout filter, simply just adds the given code to the an array.
- I don't think removal is needed or ever makes sense, but this could adopt the [`CBA_fnc_addClassEventHandler`](https://github.com/CBATeam/CBA_A3/wiki/Adding-Event-Handlers-to-Classes-of-Objects#removing-an-event-handler) approach for that.

`CBA_fnc_filterLoadout`

- Filters the given loadout by applying all loadout filters to it.
- `CBA_fnc_getUnitLoadout` will not be needed because this function should handle receiving the different inputs for the `getUnitLoadout` command as well as a normal loadout array - `OBJECT` (unit), `STRING` (CfgVehicles class name), and `CONFIG` (loadout config).

commy2:
Totally agree with what **mharis001** wrote.

Removal makes no sense (here and I argue in general). If you need to disable it on the fly, turn it off by condition in the filter script.

`CBA_fnc_getUnitLoadout` currently is just input validation for `CBA_fnc_filterLoadout`. `CBA_fnc_filterLoadout` should do its own validation (if even necessary for this!), get is not needed.

2 functions are enough.

___

>Filtered loadouts may not be as sexy as weapon optics or face wear randomization in the eyes of everyone

lol nice dig

jonpas:
Let's just appreciate his post is exactly 500 words long. :D

commy2:
505 according to MS Word.

jonpas:
Don't count the title.

dedmen:
So in TFAR I iterate all BI Arsenal loadouts after you close arsenal, and remove the ID from TFAR radios,
Would this be essentially the same then? Would tihs automatically integrate with Arsenal or would I still need to do that myself?
I assume ACE Arsenal and BI Arsenal would need to implement this system into their loadout save/load code?
We know BI Arsenal will never do that, so who does that?
Or will TFAR/ACRE keep their loadout handlers the same but instead of patching directly, just invoke the CBA stuff? 
Such that we end up with a dozen mods all doing that and calling the loadout filtering a dozen times?

synixebrett:
I guess that could be added to CBA as part of this PR? Then that functionality would be part of every mod that would use this. That would potentially cause an issue though as mod defined filters should be used, but not mission defined filters

addons/common/fnc_addLoadoutFilter.sqf
```diff
@@ -0,0 +1,32 @@
+#include "script_component.hpp"
+/* ----------------------------------------------------------------------------
+Function: CBA_fnc_addLoadoutFilter
```
mharis001, line 3: Should add blank lines between different header sections.

addons/common/fnc_filterLoadout.sqf
```diff
@@ -0,0 +1,25 @@
+#include "script_component.hpp"
+/* ----------------------------------------------------------------------------
+Function: CBA_fnc_filterLoadout
```
mharis001, line 3: Should add blank lines between different header sections.

addons/common/init_filteredLoadout.sqf
```diff
@@ -0,0 +1,3 @@
+#include "script_component.hpp"
+
+GVAR(loadoutFilters) = [];
```
mharis001, line 3: Put this directly in `XEH_preInit.sqf`; one line does not need a separate file.

addons/common/fnc_addLoadoutFilter.sqf
```diff
@@ -0,0 +1,32 @@
...
+            _loadout
+        }] call CBA_fnc_addLoadoutFilter;
+    (end)
+Author:
+    SynixeBrett
+---------------------------------------------------------------------------- */
+
+params [["_function", {}, [{}]]];
+
+if (_function isEqualTo {}) exitWith {-1};
```
mharis001, line 30: Remove `-1` and add `nil` at the end of file since header says function returns nothing.

addons/common/fnc_filterLoadout.sqf
```diff
@@ -0,0 +1,25 @@
...
+        _loadout = [getUnitLoadout player] call CBA_fnc_filterLoadout;
+    (end)
+Returns:
+    filtered getUnitLoadout array <ARRAY>
+Author:
+    SynixeBrett
+---------------------------------------------------------------------------- */
+SCRIPT(filterLoadout);
+
+params [["_loadout", [], [[]]]];
```
mharis001, line 19: Should handle receiving the different inputs (OBJECT, STRING, CONFIG) for the `getUnitLoadout` command as well as a normal loadout array.

addons/common/fnc_addLoadoutFilter.sqf
commy2: It could return false, and true if it worked.

BrettMayson:
Close in favour of #1503

PR #1271, opened 2019-12-27
CBA_fnc_sortNestedArray - Preserve order of elements with same sorting value

mharis001:
- title

Example (before changing to default behaviour):

```sqf
_a = [
    ["a", 100],
    ["b", 200],
    ["e", 100],
    ["d", 150],
    ["c", 100],
    ["abc", 200]
];

[_a, 1, true, true] call CBA_fnc_sortNestedArray;
// [["a",100],["e",100],["c",100],["d",150],["b",200],["abc",200]]

_a pushBack ["x", 100];
// [["a",100],["e",100],["c",100],["d",150],["b",200],["abc",200],["x",100]]

[_a, 1, true, true] call CBA_fnc_sortNestedArray;
// [["a",100],["e",100],["c",100],["x",100],["d",150],["b",200],["abc",200]]

// above, ["x", 100] is sorted after all the elements that appeared before it
// however, if preserve is false, the ordering of elements relative to each other is not preserved
[_a, 1, true, false] call CBA_fnc_sortNestedArray;
// [["e",100],["c",100],["x",100],["a",100],["d",150],["abc",200],["b",200]]
```

### DIFF
--- a/addons/arrays/fnc_sortNestedArray.sqf
+++ b/addons/arrays/fnc_sortNestedArray.sqf
@@ -3,15 +3,19 @@
 Function: CBA_fnc_sortNestedArray
 
 Description:
-    Used to sort a nested array from lowest to highest using quick sort.
+    Sorts the given nested array in either ascending or descending order based on the
+    numerical value at specified index of sub arrays.
 
-    Sorting is based on the specified column, which must have numerical data.
+    Optionally, the current ordering (relative to each other) of array elements with
+    the same numerical sorting value can be preserved.
+
     Original array is modified.
 
 Parameters:
     _array - Nested array to be sorted <ARRAY>
     _index - Sub array item index to be sorted on <NUMBER>
     _order - true: ascending, false: descending (optional, default: true) <BOOLEAN>
+    _preserve - Preserve current order of items with the same sorting value (optional, default: false) <BOOLEAN>
 
 Example:
     (begin example)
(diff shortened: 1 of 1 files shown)


commy2:
I don't get it.

mharis001:
Updated the example to be more clear. My use case for this is an array of actions (each with a priority). This addition will allow actions to be sorted in terms of decreasing priority (like current), but also keep actions with the same priority in the order they were added - instead of performing additional unwanted sorting on the full "action sub-arrays".

commy2:
>keep actions with the same priority in the order they were added

Shouldn't that always be the case?

>instead of performing additional unwanted sorting on the full "action sub-arrays"

Seems like a bug/oversight to me.

I think this does not need additional syntax, but just always sort the way your alt syntax is now.

addons/arrays/fnc_sortNestedArray.sqf
```diff
@@ -22,20 +22,23 @@ Returns:
...
 
-private _helperArray = _array apply {
-    [_x select _index, _x]
-};
+private _helperArray = [];
+private _coefficient = [-1, 1] select _order;
+
+{
+    _helperArray pushBack [_x select _index, _coefficient * _forEachIndex, _x];
+} forEach _array;
```
commy2: Please keep the `apply`  to save N `pushBack` commands.
```sqf
private _helperArray = _array apply {
    [_x select _index, _coefficient * _forEachIndex, _x]
};
```

PabstMirror: no `_forEachIndex` in apply

commy2: Dang.

PR #1274, opened 2020-01-03
AI: Don't halt planned movement for empty group

Vdauphin:
`CBA_fnc_clearWaypoints` can be call for empty group. In this case, a waypoint is created at [0,0,0] because `getPosATL (leader _group)`. This waypoint is unnecessary in case of empty group because there is no reason to "halt all planned movement".

- This pull request avoid the creation of a waypoint for empty group.
- Each change in a separate line
  - Check if group is not empty
  - Create the waypoint if there is at least a unit in the group

### DIFF
--- a/addons/ai/fnc_clearWaypoints.sqf
+++ b/addons/ai/fnc_clearWaypoints.sqf
@@ -29,6 +29,8 @@ private _waypoints = waypoints _group;
     deleteWaypoint [_group, 0];
 } forEach _waypoints;
 
-// Create a self-deleting waypoint at the leader position to halt all planned movement (based on old waypoints)
-private _wp = _group addWaypoint [getPosATL (leader _group), 0];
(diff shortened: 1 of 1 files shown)


commy2:
Does the WP at [0,0,0] for the empty group cause any issues?

Vdauphin:
I am using `CBA_fnc_taskPatrol` which is calling `CBA_fnc_clearWaypoints` 
All groups look stuck and don't start the patrol

Vdauphin:
Here is the code I am using to identify the issue:

```sqf
private _group = createGroup side player;  
[_group, getpos player, 100, 2 + floor (random 4), "MOVE", "SAFE", "RED", "LIMITED", "STAG COLUMN", "", [5, 10, 20]] call CBA_fnc_taskPatrol;  

[_group] spawn { 
   _group = _this select 0;
   sleep 10;
   _group createUnit [typeOf player, getpos player, [], 0, "CARGO"];  
   btc_curator addCuratorEditableObjects [units _group, false];
}
```

PR #1275, opened 2020-01-04
Don't URLdecode when URL has no encoded characters

dedmen:
Ref to comment: https://github.com/CBATeam/CBA_A3/pull/1166#issuecomment-570826926

### DIFF
--- a/addons/strings/fnc_decodeURL.sqf
+++ b/addons/strings/fnc_decodeURL.sqf
@@ -34,7 +34,7 @@ if (isNil "_return") then {
         {
             _return = ([_return] + _x) call CBA_fnc_replace;
         } forEach UTF8_TABLE;
-    }
+    };
     if (isNull _cache) then {
         _cache = [] call CBA_fnc_createNamespace;
         missionNamespace setVariable [QGVAR(URLCache), _cache];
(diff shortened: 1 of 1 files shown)


addons/strings/fnc_decodeURL.sqf
```diff
@@ -28,11 +28,13 @@ private _return = _cache getVariable _string;
...
-        _return = ([_return] + _x) call CBA_fnc_replace;
-    } forEach UTF8_TABLE;
-
+    
+    //Only need to check if there is atleast one character to replace in the stirng
+    if ("%" in _return) then {
+        {
+            _return = ([_return] + _x) call CBA_fnc_replace;
+        } forEach UTF8_TABLE;
+    }
```
commy2:
```suggestion
    };
```
Too much C++

addons/strings/fnc_decodeURL.sqf
```diff
@@ -28,11 +28,13 @@ private _return = _cache getVariable _string;
 
 if (isNil "_return") then {
     _return = _string;
-
-    {
-        _return = ([_return] + _x) call CBA_fnc_replace;
-    } forEach UTF8_TABLE;
-
+    
+    //Only need to check if there is atleast one character to replace in the stirng
```
jonpas:
```suggestion
    // Only replace if there is at least one character to replace
```

addons/strings/fnc_decodeURL.sqf
dedmen: derp, yep indeed

PR #1281, opened 2020-01-12
restore selected tab in settings ui

commy2:
- make the settings UI remember which tab (SERVER/MISSION/CLIENT) you selected previously when reopening the UI.


jonpas:
What happens if previously selected tab isn't available (eg. no Server in Eden)?

commy2:
-_-
I'll try again tomorrow.

Issue #1190, opened 2019-07-27
Enoch/Contact weapons/mags need JAM


Drofseh:
**Arma 3 Version:** `1.94`

**Mods:**
```
- CBA_A3
```

**Description:**
This is just a reminder to myself to do CBA magwells for the new weapons /mags

commy2:
Enoch
```sqf
["75Rnd_762x39_Mag_F","Magazine","Bullet"],
["75Rnd_762x39_Mag_Tracer_F","Magazine","Bullet"],
["30rnd_762x39_AK12_Lush_Mag_F","Magazine","Bullet"],
["30rnd_762x39_AK12_Lush_Mag_Tracer_F","Magazine","Bullet"],
["30rnd_762x39_AK12_Arid_Mag_F","Magazine","Bullet"],
["30rnd_762x39_AK12_Arid_Mag_Tracer_F","Magazine","Bullet"],
["75rnd_762x39_AK12_Mag_F","Magazine","Bullet"],
["75rnd_762x39_AK12_Mag_Tracer_F","Magazine","Bullet"],
["75rnd_762x39_AK12_Lush_Mag_F","Magazine","Bullet"],
["75rnd_762x39_AK12_Lush_Mag_Tracer_F","Magazine","Bullet"],
["75rnd_762x39_AK12_Arid_Mag_F","Magazine","Bullet"],
["75rnd_762x39_AK12_Arid_Mag_Tracer_F","Magazine","Bullet"],
["10Rnd_Mk14_762x51_Mag","Magazine","Bullet"],
["30Rnd_65x39_caseless_msbs_mag","Magazine","Bullet"],
["30Rnd_65x39_caseless_msbs_mag_Tracer","Magazine","Bullet"],
["2Rnd_12Gauge_Pellets","Magazine","Bullet"],
["2Rnd_12Gauge_Slug","Magazine","Artillery"],
["6Rnd_12Gauge_Pellets","Magazine","Bullet"],
["6Rnd_12Gauge_Slug","Magazine","Artillery"],
["O_R_IR_Grenade","Magazine","UnknownMagazine"],
["I_E_IR_Grenade","Magazine","UnknownMagazine"]
```

Contact
```sqf
["30Rnd_556x45_Stanag_Blank","Magazine","Artillery"],
["30Rnd_65x39_caseless_blank_mag","Magazine","Artillery"],
["100Rnd_65x39_caseless_blank_mag","Magazine","Artillery"],
[...]

PR #1286, opened 2020-01-26
Add magwells for KSVK, AA-12, CZ-750, vz 59

Drofseh:
- See title


commy2:
So, when you'll do 1190 =?

Drofseh:
> So, when you'll do 1190 =?

I don't understand...

commy2:
#1190

PR #1293, opened 2020-02-09
Tools - Removed deprecated Mikero tools arguments make.py

Dahlgren:
- Make `make.py` compatible with latest Mikero Tools updates

### DIFF
--- a/tools/make.py
+++ b/tools/make.py
@@ -899,12 +899,6 @@ See the make.cfg file for additional build options.
         quiet = True
         argv.remove("quiet")
 
-    if "checkexternal" in argv:
-        argv.remove("checkexternal")
-        check_external = True
-    else:
-        check_external = False
-
     if "version" in argv:
         argv.remove("version")
         version_update = True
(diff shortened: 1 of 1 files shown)


Dahlgren:
Makepbo still fails on
https://github.com/CBATeam/CBA_A3/blob/0464ca4a143b92c71a0f8e04c0abac649841358f/addons/ee/config.cpp#L15

I've reached out to mikero about it

jonpas:
Why were external checks suddenly removed from Mikero tools?

Dahlgren:
It's no longer optional and will always be applied

jonpas:
Time to finally switch CBA and ACE to HEMTT, this is getting ridiculous.

Who wants this to not be an option? Sounds terrible for quick (eg. nightly) playable builds without extracting all the used Arma files.

commy2:
>I've reached out to mikero about it

If any questions arise for the purpose of this line; it is required and has no alternatives. However, I am not willing to participate in any discussions about this.

Dystopian:
> It's no longer optional and will always be applied

Why not just using old version of mikero tools until moving to HEMTT?

dedmen:
> Why not just using old version of mikero tools until moving to HEMTT?

We can certainly do that as devs, for builds and such.
But contributors who want to make a PR and test their addition, will not.

commy2:
Even the old version cannot properly handle the scope entry.

PR #1294, opened 2020-02-10
add CBA_fnc_switchableAttachments

commy2:
- Returns all attachments the item can eventually switch to.
- for https://github.com/acemod/ACE3/issues/7130

### DIFF
--- a/addons/accessory/fnc_switchableAttachments.sqf
+++ b/addons/accessory/fnc_switchableAttachments.sqf
@@ -31,6 +31,7 @@ while {
     isClass _config && {_forward pushBackUnique configName _config != -1}
 } do {};
 
+_config = _cfgWeapons >> _item;
 private _backward = [];
 while {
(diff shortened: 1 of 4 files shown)


addons/accessory/fnc_switchableAttachments.sqf
```diff
@@ -0,0 +1,41 @@
...
+
+private _cfgWeapons = configfile >> "CfgWeapons";
+private _config = _cfgWeapons >> _item;
+
+private _forward = [];
+while {
+    _config = _cfgWeapons >> getText (_config >> "MRT_SwitchItemNextClass");
+    isClass _config && {_forward pushBackUnique configName _config != -1}
+} do {};
+
```
Dystopian: if the last `_config` doesn't exist, backward loop will be empty.
```suggestion

_config = _cfgWeapons >> _item;
```

Dystopian:
maybe caching would be useful here

PR #1299, opened 2020-03-01
Return target exec result back to executor

veteran29:
- title

### DIFF
--- a/addons/diagnostic/fnc_initTargetDebugConsole.sqf
+++ b/addons/diagnostic/fnc_initTargetDebugConsole.sqf
@@ -14,7 +14,7 @@ Author:
 ---------------------------------------------------------------------------- */
 
 #define COUNT_WATCH_BOXES 8
-#define EXEC_RESULT ([nil] apply {[] call _this} param [0, text ""])
+#define EXEC_RESULT ([nil] apply {[] call _this} param [0, text "<any>"])
 #define EXEC_RESULT_CTRL (parsingNamespace getVariable ["BIS_RscDebugConsoleExpressionResultCtrl", controlNull])
 #define EXEC_SEND_RESULT {[EXEC_RESULT, {EXEC_RESULT_CTRL ctrlSetText str _this}] remoteExec ["call", remoteExecutedOwner]}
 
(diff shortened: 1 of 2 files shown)


addons/diagnostic/fnc_initTargetDebugConsole.sqf
```diff
@@ -13,7 +13,10 @@ Author:
     commy2
 ---------------------------------------------------------------------------- */
 
-#define COUNT_WATCH_BOXES 8
+#define COUNT_WATCH_BOXES   8
+#define EXEC_RESULT         ([nil] apply {[] call _this} param [0, text ""])
```
commy2: The watch fields report #NIL for nil return value and I like that you can see that something happened instead of void.

Vanilla reports:
```
nil -> 
[nil] -> [<any>]
```
So it only works for array elements there, but we can be better than that.

I'd wish for `text "<any>"` as default value, and also the same for the watch fields (could be part of this PR).

commy2: Also, don't do this multiple spaces inside a line formatting. Should there ever be another macro added with a longer name, all the other spacing will have to be adjusted resulting in a pink wall. Don't do that.

veteran29: Suggestions applied.

addons/diagnostic/fnc_initTargetDebugConsole.sqf
```diff
@@ -14,6 +14,9 @@ Author:
 ---------------------------------------------------------------------------- */
 
 #define COUNT_WATCH_BOXES 8
+#define EXEC_RESULT ([nil] apply {[] call _this} param [0, text "<any>"])
```
dedmen: is nil used here to be as cheap as possible? a number would be cheaper.

commy2: This is the debug console. It doesn't matter.

commy2: Why the eyes? I'd rather `_x` return <any> than a random number or whatever.

PR #1306, opened 2020-03-12
createPerFrameHandlerObject - remove pfeh if location deleted

PabstMirror:
If location doesn't exist, then you won't get a valid getVar from it

```
x = [{systemChat str time}, 1] call CBA_fnc_createPerFrameHandlerObject
```
```
deleteLocation x
```

`count cba_common_perFrameHandlerArray` does not decrease

Edit: This would be a rare bug, `CBA_fnc_deletePerFrameHandlerObject` was safe

### DIFF
--- a/addons/common/fnc_createPerFrameHandlerObject.sqf
+++ b/addons/common/fnc_createPerFrameHandlerObject.sqf
@@ -102,7 +102,7 @@ private _handle = [{
     params ["_logic"];
 
     if (isNil "_logic" || {isNull _logic}) exitWith {
-        (_logic getVariable "handle") call CBA_fnc_removePerFrameHandler;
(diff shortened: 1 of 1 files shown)


addons/common/fnc_createPerFrameHandlerObject.sqf
```diff
@@ -102,7 +102,7 @@ private _handle = [{
     params ["_logic"];
 
     if (isNil "_logic" || {isNull _logic}) exitWith {
-        (_logic getVariable "handle") call CBA_fnc_removePerFrameHandler;
+        [_this # 1] call CBA_fnc_removePerFrameHandler;
```
commy2:
```suggestion
         (_this select 1) call CBA_fnc_removePerFrameHandler;
```

PR #1307, opened 2020-03-12
Update make.py

PabstMirror:
Need this to compile with latest available free tools
Not sure if this conflicts with #1293 but these are the best tools that everyone has access to.
```
In File P:\x\cba\addons\ee\config.cpp: circa Line 15 var= is not an integer or out of range

Rapify error
"ee.pbo not produced due to error(s)" 
```

### DIFF
--- a/tools/make.py
+++ b/tools/make.py
@@ -1262,7 +1262,7 @@ See the make.cfg file for additional build options.
 
                     if os.path.isfile(nobinFilePath):
                         print_green("$NOBIN$ Found. Proceeding with non-binarizing!")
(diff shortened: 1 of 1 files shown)


PabstMirror:
@Dahlgren - Could you test this branch? 
Will it negatively effect the latest paid tools?
 Could you build cba_ee with/without this PR?

Dahlgren:
- With `-N` makepbo fails to launch (`-N wysiwig no longer supported`)
- Without `-N` it fails to build `cba_ee`

commy2:
Get the free version of the tools. Complain to him that the paid version does not work.

PR #1308, opened 2020-03-14
Added 762x39 SKS magazine wells

10Dozen:
- Add 762x39 SKS magazine well for clips (classic SKS) and for magazines (modernized SKS)


commy2:
You need these for CUP?

10Dozen:
> You need these for CUP?
Yes, we added SKS recently.

PR #1310, opened 2020-03-24
Tools - Set build number in HEMTT release build

jonpas:
- Set build number in pre-build step.
- Unset build number in release step (after zipping).
- Set build number in development to `000000`.

### DIFF
--- a/addons/main/script_version.hpp
+++ b/addons/main/script_version.hpp
@@ -1,4 +1,4 @@
 #define MAJOR 3
 #define MINOR 14
 #define PATCHLVL 0
-#define BUILD 200207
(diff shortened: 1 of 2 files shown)


jonpas:
@commy2 please test on Windows. Resulting build should have today's build number and everything should be reset at the end of the build process (no git diffs).

commy2:
After the fix, the only diff from build_release.bat are changed line endings in one file.


Idk if that is acceptable.

jonpas:
Interesting, those shouldn't be changed. `mod.cpp` doesn't do that, does it?

commy2:
Apparently not.

PR #1311, opened 2020-03-26
XEH - Update configs for 1.98 (Oldman)

PabstMirror:
For current RC (make sure to test after final release)
```
[CBA] (xeh) WARNING: O_UAV_01_Mine_F does not support Extended Event Handlers! Addon: expansion
[CBA] (xeh) WARNING: Land_MapBoard_01_Wall_Syndikat_Tanoa_F does not support Extended Event Handlers! Addon: expansion
[CBA] (xeh) WARNING: Land_MapBoard_01_Tripod_Syndikat_Tanoa_F does not support Extended Event Handlers! Addon: expansion
[CBA] (xeh) WARNING: Land_MapBoard_01_Map_Syndikat_Tanoa_F does not support Extended Event Handlers! Addon: expansion
```

### DIFF
--- a/addons/common/config.cpp
+++ b/addons/common/config.cpp
@@ -6,7 +6,7 @@ class CfgPatches {
         units[] = {};
         weapons[] = {};
         requiredVersion = REQUIRED_VERSION;
-        requiredAddons[] = {"A3_Data_F_Enoch_Loadorder", "A3_Data_F_Mod_Loadorder"};
+        requiredAddons[] = {"A3_Data_F_Oldman_Loadorder", "A3_Data_F_Mod_Loadorder"};
         author = "$STR_CBA_Author";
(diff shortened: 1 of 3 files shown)


commy2:
Is this bad if merged now? It may create duplicate classes if it inherits from something public (which this all seems to).

PabstMirror:
I think the base classes are all scope=1
so it should be fine, I can remove the 1.98 specific stuff

PabstMirror:
but I also remember XEH stuff changing between RC and release, so it might be better to hold off

commy2:
I'd say we ignore it for now and release it as update (3.5.1?) once Oldman is out.

jonpas:
3.5.1*

PabstMirror:
Ready to go, final seems same as RC

PR #1313, opened 2020-03-27
Events - Player events will report driver seat as turret [-1]

PabstMirror:
Replace #1312 
Reports driver/pilot as turret `[-1]`
This is mainly so it will also report pilot's weapon on "turretweapon" event

e.g. `["turretweapon", {x1 = +_this}, true] call CBA_fnc_addPlayerEventHandler;`
for AH-6 Littlebird will report `[player, "M134_minigun","missiles_DAR"]`

**This could cause problems for code that is not expecting driver's seat turret**
And many cars will report a weapon like `"TruckHorn2"` which might not be expected

### DIFF
--- a/addons/events/fnc_addPlayerEventHandler.sqf
+++ b/addons/events/fnc_addPlayerEventHandler.sqf
@@ -65,7 +65,7 @@ private _id = switch (_type) do {
     };
     case "turretweapon": {
         if (_applyRetroactively) then {
-            [GVAR(oldUnit), GVAR(oldUnit) currentWeaponTurret (GVAR(oldUnit) call CBA_fnc_turretPath), ""] call _function;
+            [GVAR(oldUnit), GVAR(oldTurretWeapon), ""] call _function;
         };
         [QGVAR(turretWeaponEvent), _function] call CBA_fnc_addEventHandler // return id
     };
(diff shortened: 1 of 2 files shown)


addons/events/fnc_addPlayerEventHandler.sqf
```diff
@@ -65,7 +65,7 @@ private _id = switch (_type) do {
     };
     case "turretweapon": {
         if (_applyRetroactively) then {
-            [GVAR(oldUnit), GVAR(oldUnit) currentWeaponTurret (GVAR(oldUnit) call CBA_fnc_turretPath), ""] call _function;
+            [GVAR(oldUnit), GVAR(oldTurretWeapon), ""] call _function;
```
commy2: Why?

addons/events/fnc_addPlayerEventHandler.sqf
```diff
@@ -95,7 +95,7 @@ private _id = switch (_type) do {
     };
     case "turret": {
         if (_applyRetroactively) then {
-            [GVAR(oldUnit), GVAR(oldUnit) call CBA_fnc_turretPath, []] call _function;
+            [GVAR(oldUnit), GVAR(oldTurret), []] call _function;
```
commy2: Same. Why?

addons/events/fnc_addPlayerEventHandler.sqf
PabstMirror: we have the value available in a variable, simplest and fastest way to return it

addons/events/fnc_addPlayerEventHandler.sqf
PabstMirror: we're already using `GVAR(oldUnit)`, why not do the same for other vars

commy2: All other besides player calculate the current value.

PabstMirror: If all the other ones drove off a cliff, should this one as well?

Issue #1241, opened 2019-10-13
CBA_fnc_isMusicPlaying Incorrect Returns


Ansible2:
**Mods:**
- **Arma 3:** `1.94 Stable`
- **CBA:** `3.12.2 Stable` 

**Description:**
`CBA_fnc_isMusicPlaying` returns `true` perpetually (after music has ended) when the music was played with `CBA_fnc_playMusic` and will not return true when used with the command `playMusic`.

**NOTE:** This also affects `CBA_fnc_playMusic`'s interrupt playing music capability. 

**Steps to reproduce:**
- Open editor in VR, place single player object, preview scenario in Singleplayer
- Open menu, paste the following lines into the debug console and click LOCAL EXEC button

```sqf
playmusic "eventTrack01_f_curator";

systemChat str (call cba_fnc_isMusicPlaying);

[
	{systemChat str (call cba_fnc_isMusicPlaying);},
	[],
	16
] call CBA_fnc_waitAndExecute;


[
	{

		["eventTrack01_f_curator",0,true] call CBA_fnc_playMusic;
		systemChat str (call cba_fnc_isMusicPlaying);
		[
			{systemChat str (call cba_fnc_isMusicPlaying);},
			[],
			16
		] call CBA_fnc_waitAndExecute;
	},
	[],
	18
] call CBA_fnc_waitAndExecute;
```

- Click CONTINUE button, listen for music and observe initial `systemChat`
- Wait 16 seconds, observe second `systemChat`
- Wait 2 seconds, listen for music and observe `systemChat`
- Wait 16 seconds for final `systemChat`

**Expected behavior:**
- Music plays & first message displays `true`
- Second message displays `false`
- Music plays & third message displays `true`
- Fourth message displays `false`

**Where did the issue occur?**
- Dedicated / Singleplayer / Editor (Singleplayer)

**Log Files:**
https://gist.github.com/Ansible2/ee0c44af9288745471f9fa058a7e99ff

**Additional context:**
- Used track length is approximately 14 seconds
- `CBA_fnc_stopMusic` will reset `CBA_fnc_isMusicPlaying`'s state back to `false`

PR #1315, opened 2020-03-30
Fix for CBA_fnc_isMusicPlaying not handling music stop properly

Eathox:
As you might know the CBA Music module isn't useful in its current state. This is because the function `CBA_fnc_isMusicPlaying` is unable to detect when a song ended and cant handle music that was started without calling `CBA_fnc_playMusic`.

This pull request fixes most of these issues by utilizing the vanilla music event handlers, however this leave's one special case. That being `playMusic ""` since that doesn't trigger any of the currently present music events, this issue is fixed by incorporating an extra check in `CBA_fnc_isMusicPlaying`.

When merged this pull request will:
- Fix `CBA_fnc_isMusicPlaying` incorrect return value ([#1241](https://github.com/CBATeam/CBA_A3/issues/1241), [#1024](https://github.com/CBATeam/CBA_A3/issues/1024))
- Make `CBA_fnc_getMusicPlaying` use `CBA_fnc_isMusicPlaying` instead of a `isNil` check
- Improve the accuracy of the `_trackTime` value returned from `CBA_fnc_getMusicPlaying`

### DIFF
--- a/addons/music/XEH_preInit.sqf
+++ b/addons/music/XEH_preInit.sqf
@@ -1,0 +1,5 @@
+#include "script_component.hpp"
+
+#include "XEH_PREP.sqf"
+
+[] call FUNC(initMusicHandler)
--- a/addons/music/fnc_isMusicPlaying.sqf
+++ b/addons/music/fnc_isMusicPlaying.sqf
@@ -20,4 +20,19 @@ Author:
     Dedmen, Commy2, Fishy
 ---------------------------------------------------------------------------- */
 
-!isNil QGVAR(track)
+#define MINIMUM_WAIT 0.1
+
+if (isNil QGVAR(track)) exitWith {false};
+
+GVAR(track) params ["_className", "_startTime", "_playPos", "_duration"];
+
+private _return = true;
+
+// Check if theres a track playing, since `playMusic ""` doesn't trigger an event
+private _trackTime = getMusicPlayedTime;
+if (CBA_missionTime > (_startTime + MINIMUM_WAIT) && {_trackTime == 0}) then {
+    GVAR(track) = nil;
+    _return = false;
+};
+
+_return
--- a/addons/music/CfgEventHandlers.hpp
+++ b/addons/music/CfgEventHandlers.hpp
@@ -1,0 +1,5 @@
+class Extended_PreInit_EventHandlers {
+    class ADDON {
+        init = QUOTE(call COMPILE_FILE(XEH_preInit));
+    };
+};
--- a/addons/music/XEH_PREP.sqf
+++ b/addons/music/XEH_PREP.sqf
@@ -1,0 +1,1 @@
+PREP(initMusicHandler);
--- a/addons/music/config.cpp
+++ b/addons/music/config.cpp
@@ -14,5 +14,6 @@ class CfgPatches {
     };
 };
 
+#include "CfgEventHandlers.hpp"
 #include "CfgFunctions.hpp"
 #include "CfgMusic.hpp"
--- a/addons/music/fnc_getMusicPlaying.sqf
+++ b/addons/music/fnc_getMusicPlaying.sqf
@@ -20,18 +20,11 @@ Author:
     Fishy, Dedmen
 ---------------------------------------------------------------------------- */
 
-if (isNil QGVAR(track)) exitWith {["", 0, 0]};
+if !(call CBA_fnc_isMusicPlaying) exitWith {["", 0, 0]};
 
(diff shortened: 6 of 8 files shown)


addons/music/fnc_isMusicPlaying.sqf
```diff
@@ -20,4 +20,19 @@ Author:
     Dedmen, Commy2, Fishy
 ---------------------------------------------------------------------------- */
 
-!isNil QGVAR(track)
+#define MINIMUM_WAIT 0.1
```
commy2: Explain.

addons/music/XEH_preInit.sqf
```diff
@@ -0,0 +1,5 @@
+#include "script_component.hpp"
+
+#include "XEH_PREP.sqf"
+
+[] call FUNC(initMusicHandler)
```
commy2: Don't think a function is needed here. Just put addMusicEH into the preInit script

addons/music/fnc_isMusicPlaying.sqf
Eathox: In order to check if `playMusic ""` was used to stop the music, the function checks if the current mission time and the mission time when the last track was started differ. If they do differ check if `getMusicPlayedTime` equals 0 if so that would mean that there is currently no music playing. However `getMusicPlayedTime` seems to be an estimation and can result in the same value over multiple frames, leading to false positives if the track just started. In order to prevent these false positives the function doesn't check if `getMusicPlayedTime` equals 0 if the current time and the starting time are withing `MINIMUM_WAIT` of each other.

commy2: So if getMusicPlayedTime returns garbo, how do we know 0.1s is enough?

Eathox: I have now done more in depth testing on both my desktop and laptop, in a locally hosted server with only one player. ([script I used for testing](https://pastebin.com/qG6qbaWB))

I tested a variety of FPSes taking 150 samples for each, these are my results:

FPS | Max Time | Average Time
--- | --- | ---
2 | 2.333010 | 1.155892
10 | 0.385986 | 0.300724
20 | 0.240967 | 0.191904
36 | 0.170898 | 0.157701
59 | 0.150009 | 0.126583
120 | 0.120972 | 0.106157
454 | 0.100006 | 0.090674

Based on this its safe to say 0.1s is definitely not enough.

This leaves me with a question: what would you like me to do?
Count for worst case scenario? Would have to test 1 FPS then.
Have a minimum delay of 2.5s, and switch that  to 0.5s if they have 10+ fps?
Or go even further and base the delay on a FPS - Minimum delay table?

Keep in mind that this only to account for `playMusic ""`, CBA_fnc_stopMusic works perfectly fine.

Eathox: I have gone with the FPS based switch since it pretty simple and still responsive.

addons/music/fnc_isMusicPlaying.sqf
```diff
@@ -20,19 +20,38 @@ Author:
...
+ * | --- | ------------------ | ---------------------- |
+ * | 2   | 2.333010           | 1.155892               |
+ * | 10  | 0.385986           | 0.300724               |
+ * | 20  | 0.240967           | 0.191904               |
+ * | 36  | 0.170898           | 0.157701               |
+ * | 59  | 0.150009           | 0.126583               |
+ * | 120 | 0.120972           | 0.106157               |
+ * | --- | ------------------ | ---------------------- |
+ * This data was collected from 2 machines on a locally hosted server with only one player */
+private _miniumWaitTime = [WAIT_HIGH_FPS, WAIT_LOW_FPS] select (diag_fpsMin < WAIT_SWITCH_FPS);
```
commy2, line 50: This is just weird. I hate it.

Eathox: Could you elaborate what it is that you hate about this? Its implementation or the the solution it self?

Also I just noticed an over sight on my part, this was meant to be `diag_fps` instead of `diag_fpsMin`

commy2: It just does not make any sense to have code run differently on different framerate.

Eathox: Got it, then I'll change it to accommodate for worst case scenario.

Eathox: Got it, I'll change it one minimum wait time that accommodates the worst case scenario.

PR #1317, opened 2020-04-03
Tools - Show version when installing HEMTT

PabstMirror:
Small QOL (also newest hemtt has lower version number?)

### DIFF
--- a/.github/workflows/arma.yml
+++ b/.github/workflows/arma.yml
@@ -30,18 +30,17 @@ jobs:
       uses: arma-actions/sqflint@master
       continue-on-error: true # No failure due to many false-positives
 
-  # Requires HEMTT 0.8.0-2cf6fb7
-  # build:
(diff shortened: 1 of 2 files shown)


jonpas:
Make sure it works with PowerShell 3.0 (not higher).

> Small QOL (also newest hemtt has lower version number?)

Yes, we were on a development version. There is no difference to the user right now.

jonpas:
Looks like HEMTT wasn't update on dockerhub yet, let's wait for that, I poked Brett.

jonpas:
Works fine with PowerShell 3.0.

commy2:
Merge?

jonpas:
No

> Looks like HEMTT wasn't update on dockerhub yet, let's wait for that, I poked Brett.

jonpas:
Builds going and commit build artifacts now available! :tada:

PR #1320, opened 2020-04-03
FIX: CBA_missionTime time leak

Vdauphin:
For `CBA_missionTime` above to 6e5s (7 days):
Addition with `CBA_missionTime` has some floating point precision error appearing randomly. This can be see thanks to this code mimicking the actual CBA behaviour:
```sqf
Vdofin_somme = 0;
Vdofin_start = 6e5;

Vdofin_missionTime = Vdofin_start;
Vdofin_lastTickTime = diag_tickTime;
v = addMissionEventHandler ["EachFrame", {
    private _additionCheckPrecision = _tickTime - Vdofin_lastTickTime;

    Vdofin_somme = Vdofin_somme + _additionCheckPrecision;

    Vdofin_missionTime = Vdofin_missionTime + _additionCheckPrecision;
    Vdofin_lastTickTime = _tickTime;
}];
```
Check in debug console the leak happening after a few second:
`Vdofin_somme + Vdofin_start - Vdofin_missionTime`

This leak is due to the problem that `_tickTime - GVAR(lastTickTime)` is around 0.01 to 0.1 which is between 1e6 to 1e7 power of magnitude smaller than `CBA_missionTime`. So when `_tickTime - GVAR(lastTickTime)` is added to 6e5 the result is the same as before the addition.
The idea is to wait until `_tickTime - GVAR(lastTickTime)` is higher. To do so, first we need to check if a floating point error occur and don't update `CBA_missionTime` AND `GVAR(lastTickTime)`. So the `_tickTime - GVAR(lastTickTime)` will get higher and higher until the floating point error do not occur any more.

For the previous code this mean this:
```sqf
Vdofin_somme = 0;
Vdofin_start = 6e5;

Vdofin_lastTickTime = diag_tickTime;
Vdofin_missionTime = Vdofin_start;
a = addMissionEventHandler ["EachFrame", {
    private _additionCheckPrecision = _tickTime - Vdofin_lastTickTime;
    Vdofin_somme = Vdofin_somme + _additionCheckPrecision;

    private _storeAddition = Vdofin_missionTime + _additionCheckPrecision;
    if (Vdofin_missionTime == _storeAddition) exitWith{};
    Vdofin_missionTime = _storeAddition;
    Vdofin_lastTickTime = _tickTime;
}];
```

For CBA, see commits.

This allow dedicated server with CBA work longer than 5e5s and lower than 1e6s without significant time leak.

- This PR avoid time leak when  `CBA_missionTime` above to 6e5s (7 days) and lower than 1e6s (11.5 days).
- Each time the CBA_missionTime is updated
  - Now checking if the addition : `CBA_missionTime + (_tickTime - GVAR(lastTickTime))` is equal to than `CBA_missionTime`
  - If is, this mean there is a floating point precision error
    - Skip the addition and the update of `GVAR(lastTickTime)`
  - If not, update the `CBA_missionTime` and `GVAR(lastTickTime)`

### DIFF
--- a/addons/common/init_perFrameHandler.sqf
+++ b/addons/common/init_perFrameHandler.sqf
@@ -46,9 +46,9 @@ GVAR(waitUntilAndExecArray) = [];
     };
     private _delete = false;
     {
-        private _timeDiff = [_x select [0, 3], CBA_missionTime];
+        private _timeDiff = [_x select [0, 3], CBA_missionTimeTriple];
         _timeDiff sort true;
-        if (_timeDiff#0 isEqualTo CBA_missionTime) exitWith {};
+        if (_timeDiff#0 isEqualTo CBA_missionTimeTriple) exitWith {};
 
         (_x select 4) call (_x select 3);
 
@@ -99,10 +99,10 @@ addMissionEventHandler ["Loaded", {
     GVAR(lastTickTime) = _tickTime;
 }];
 
-CBA_missionTime = [0, 0, 0]; // [999,999,999.99] = 999999999.99
+CBA_missionTimeTriple = [0, 0, 0]; // [999,999,999.99] = 999999999.99
 GVAR(lastTime) = time;
 
-// increase CBA_missionTime variable every frame
+// increase CBA_missionTimeTriple variable every frame
 if (isMultiplayer) then {
     // multiplayer - no accTime in MP
     if (isServer) then {
@@ -110,14 +110,15 @@ if (isMultiplayer) then {
         [QFUNC(missionTimePFH), {
             SCRIPT(missionTimePFH_server);
             if (time != GVAR(lastTime)) then {
-                CBA_missionTime = CBA_missionTime vectorAdd [0, 0, _tickTime - GVAR(lastTickTime)];
-                if (selectMax CBA_missionTime > 1000) then {
-                    private _limitReach = CBA_missionTime findIf {_x > 1000};
+                CBA_missionTimeTriple = CBA_missionTimeTriple vectorAdd [0, 0, _tickTime - GVAR(lastTickTime)];
+                if (selectMax CBA_missionTimeTriple > 1e3) then {
+                    private _limitReach = CBA_missionTimeTriple findIf {_x > 1e3};
                     private _diff = [0, 0, 0];
-                    _diff set [_limitReach, -1000];
+                    _diff set [_limitReach, -1e3];
                     _diff set [_limitReach - 1, 1];
-                    CBA_missionTime = CBA_missionTime vectorAdd _diff;
+                    CBA_missionTimeTriple = CBA_missionTimeTriple vectorAdd _diff;
                 };
+                CBA_missionTime = CBA_missionTimeTriple vectorDotProduct [1e6 , 1e3, 1];
                 GVAR(lastTime) = time; // used to detect paused game
             };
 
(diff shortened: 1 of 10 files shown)


addons/common/init_perFrameHandler.sqf
```diff
@@ -155,10 +161,13 @@ if (isMultiplayer) then {
...
+            private _additionCheckPrecision = CBA_missionTime + (_tickTime - GVAR(lastTickTime)) * accTime;
+            if (CBA_missionTime == _additionCheckPrecision) exitWith {}; // used to detect floating point precision error
+            CBA_missionTime = _additionCheckPrecision;
             GVAR(lastTime) = time; // used to detect paused game
+            GVAR(lastTickTime) = _tickTime;
+        } else {
+            GVAR(lastTickTime) = _tickTime;
         };
-
-        GVAR(lastTickTime) = _tickTime;
```
commy2, line 162: ???

commy2:
What does `_additionCheckPrecision` mean?

I don't understand what you mean by "leak", but I haven't tried your debug code for now.

Vdauphin:
> What does _additionCheckPrecision mean?

Don't know how to call it but in the PR it contain `CBA_missionTime + (_tickTime - GVAR(lastTickTime));` and is use later to check if the addition has been done as intended.

> I don't understand what you mean by "leak", but I haven't tried your debug code for now.

Basically `6e5 + 0.03` return `6e5` and not `600000.03`, you repeat this addition 100 times (in game CBA do it every frame) you still have `6e5` but you should have `600003.00`. So in this worst case you lost/leak 3 seconds.

addons/common/init_perFrameHandler.sqf
Vdauphin: We don't want to update `GVAR(lastTickTime) = _tickTime;` if `CBA_missionTime == _additionCheckPrecision` return true

commy2:
So instead of adding small deltas of time we want the delta to grow enough to make a difference?

Vdauphin:
Yes

commy2:
Hmm, that has its own imprecissions. I was thinking about incorporating 
https://community.bistudio.com/wiki/diag_deltaTime 
somehow.
And also storing the tickTime as string internally.

Vdauphin:
> Hmm, that has its own imprecissions. I was thinking about incorporating
> https://community.bistudio.com/wiki/diag_deltaTime
> somehow.

I saw it but didn't find a clever use of it or we need to design this thing way differently
We need to avoid an addition between a big and a small number, then we are safe.

>And also storing the tickTime as string internally.

I also think about it. Like you store `"60000.00"`, then select the `"00.00"`, add `00.00 + 0.03`, and concatenate in `"60000.03"` but it is may be over engineering. 

But computer scientist should already design something to deal with this, isn't it?

commy2:
>But computer scientist should already design something to deal with this, isn't it?

They'd just use doubles instead of single floats and be done with it.

We're stuck with singles. By using strings, I attempt to get the same benefit as doubles by changing data type.

Maybe we could also use some array [a,b], where b is the last 6 digits of the mission time and a is the number of millions.
[10,987654]
= 10,987,654

Vdauphin:
> Maybe we could also use some array [a,b], where b is the last 6 digits of the mission time and a is the number of millions.
> [10,987654]
> = 10,987,654

I like this idea but the imprecision start at 1e5 because we are summing to it 1e-2 number

commy2:
[10,000,000]
then.

Vdauphin:
If we use 3D array, we can use vector operation. 
`1201.09` will be define as `[0,1,201.09]`

commy2:
lol sure

Vdauphin:
Here is an example of addition without imprecision until 1e9 (theoretical).
```sqf
removeMissionEventHandler ["EachFrame", v];

Vdofin_somme = 0;
Vdofin_start = 6e5;

Vdofin_missionTime = [
    floor(Vdofin_start/1e6),
    floor(Vdofin_start/1e3) - floor(Vdofin_start/1e6) * 1e6,
    Vdofin_start - floor(Vdofin_start/1e3) * 1e3 - floor(Vdofin_start/1e6) * 1e6
];
Vdofin_lastTickTime = diag_tickTime;

v = addMissionEventHandler ["EachFrame", {

    Vdofin_somme = Vdofin_somme + _tickTime - Vdofin_lastTickTime;

    Vdofin_missionTime = Vdofin_missionTime vectorAdd [0, 0, _tickTime - Vdofin_lastTickTime];
    if (selectMax Vdofin_missionTime > 1000) then {
        private _limitReach = Vdofin_missionTime findIf {_x > 1000};
        private _diff = [0, 0, 0];
        _diff set [_limitReach, -1000];
        _diff set [_limitReach - 1, 1];
        Vdofin_missionTime = Vdofin_missionTime vectorAdd _diff;
    };
    Vdofin_lastTickTime = _tickTime;
}];
```
Watch in debug, the different is null:
`Vdofin_somme + Vdofin_start - (Vdofin_missionTime#0 * 1e6 + Vdofin_missionTime#1 * 1e3 + Vdofin_missionTime#2)`

The commit is a prof of concept. May be it is a too big hammer

Vdauphin:
So, if this concept is fine for the CBA Team (?)
I can provide backward compatibility by defining a  `CBA_missionTimeTriple` and update `CBA_missionTime` with `CBA_missionTimeTriple vectorDotProduct [1e6 , 1e3, 1]`

commy2:
Turning `CBA_missionTime` into an array breaks bwc.

This implementation feels goofy. Will have to think about this some more.

Vdauphin:
> Turning CBA_missionTime into an array breaks bwc.

I know, that why I said:

>I can provide backward compatibility by defining a CBA_missionTimeTriple and update CBA_missionTime with CBA_missionTimeTriple vectorDotProduct [1e6 , 1e3, 1]

commy2:
I'm still in favour of using a string to store this non-single float.

Vdauphin:
With vector representation we have more engine method: https://github.com/CBATeam/CBA_A3/pull/1320#issuecomment-609392524

commy2:
You'd have to change the string covering the >1000 digits only once every 1000 seconds=16,67 minutes. Performance is irrelevant.

Vdauphin:
> You'd have to change the string covering the >1000 digits only once every 1000 seconds=16,67 minutes. Performance is irrelevant.

I don't see a string approach without imprecision when doing addition, substation, comparison and `sort` will be useless

What is your idea?

PabstMirror:
see #1323

Vdauphin:
See https://github.com/CBATeam/CBA_A3/pull/1324

PR #1321, opened 2020-04-04
Settings - Show raw values in tooltip for list settings

PabstMirror:
image

### DIFF
--- a/addons/settings/fnc_gui_settingList.sqf
+++ b/addons/settings/fnc_gui_settingList.sqf
@@ -19,7 +19,11 @@ private _lbData = [];
         _tooltip = localize _tooltip;
     };
 
-    _tooltip =  format ["%1 - %2", _x, _tooltip];
+    if (_tooltip isEqualTo "") then {
+        _tooltip = str _x;
+    } else {
+        _tooltip = _tooltip + endl + str _x;
+    };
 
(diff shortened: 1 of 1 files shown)


addons/settings/fnc_gui_settingList.sqf
```diff
@@ -19,6 +19,8 @@ private _lbData = [];
         _tooltip = localize _tooltip;
     };
 
+    _tooltip =  format ["%1 - %2", _x, _tooltip];
```
Dystopian, line 22:
```suggestion
    _tooltip =  format [["%1 - %2", "%1"] select (_tooltip isEqualTo ""), _x, _tooltip];
```
to remove ` - ` when tooltip is empty

commy2:
```suggestion
    if (_tooltip isEqualTo "") then {
        _tooltip = str _x;
    } else {
        _tooltip = _tooltip + endl + str _x;
    };
```

Dystopian: commy2's suggestion looks more consistent with setting name tooltip

commy2:
Someone test if newlines work in listbox tooltips.

commy2:
Not sure about this string key being appriate.

PR #1323, opened 2020-04-05
Common - Add option to use vanilla game time (mission config)

PabstMirror:
ref #1320
simpler fix
just add option to use vanilla `time`
might not be perfectly synced, but should handle extended plays much better

### DIFF
--- a/addons/common/init_perFrameHandler.sqf
+++ b/addons/common/init_perFrameHandler.sqf
@@ -98,8 +98,16 @@ addMissionEventHandler ["Loaded", {
 }];
 
 CBA_missionTime = 0;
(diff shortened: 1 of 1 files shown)


Vdauphin:
Way better!
Why not make it the default behavior?

PabstMirror:
ref 
https://github.com/acemod/ACE3/pull/1092
https://github.com/CBATeam/CBA_A3/pull/273

`time` may not be synced perfectly between clients/jips
ignores `accTime`
so there is a reason it exists, but these might be less important than extended play missions

commy2:
No.

jonpas:
Status?

PR #1324, opened 2020-04-07
fix CBA_missionTime in long missions (above 7 days)

commy2:
- alt #1320

I have no tested this yet, so there may be typos. It should all be as simple as this though, unless I am missing something.

### DIFF
--- a/addons/common/init_perFrameHandler.sqf
+++ b/addons/common/init_perFrameHandler.sqf
@@ -5,7 +5,6 @@
 
 GVAR(perFrameHandlerArray) = [];
 GVAR(perFrameHandlersToRemove) = [];
-GVAR(lastTickTime) = diag_tickTime;
 
 GVAR(waitAndExecArray) = [];
 GVAR(waitAndExecArrayIsSorted) = false;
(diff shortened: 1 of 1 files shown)


Vdauphin:
Definitely this PR simplify the CBA_missionTime update and resolve CBA_missionTime being stuck due to tiny add of value to a big number.

But, I feel it missing the opportunity to give support to do addition/subtraction/comparison of two big number without lost of precision or unintended results. I prefer the vector representation of the time for precision (math operation give us the necessary to deal with) and the classic number representation for backward compatibility. Also this representation give the possibility for mod to choose the time representation they need.

addons/common/init_perFrameHandler.sqf
```diff
@@ -14,11 +13,34 @@ GVAR(nextFrameBufferA) = [];
...
+
+        while {GVAR(missionTimePrecise) >= 1000} do {
+            GVAR(missionTimePrecise) = GVAR(missionTimePrecise) - 1000;
+            GVAR(missionTimeThousands) = GVAR(missionTimeThousands) + 1;
+        };
+
+        CBA_missionTimeStr = format [
+            "%1%2",
+            GVAR(missionTimeThousands) toFixed 0,
+            (1000 + GVAR(missionTimePrecise)) toFixed 6 select [1],
```
Vdauphin, line 39:
```suggestion
            (1000 + GVAR(missionTimePrecise)) toFixed 6 select [1]
```

commy2:
@Vdauphin 
```sqf
[my_longAgo, CBA_fnc_missionTimeStr] call CBA_fnc_missionTimeDelta;
```

Vdauphin:
Nice!
Do you plan to migrate CBA_fnc_waitAndExecute and PFH to this?

commy2:
WAE and WUAE sure, but PFH uses tickTime, not missionTime (does not suspend or accelerate in SP, starts from game start, not mission start, is not synchronized).

addons/common/init_perFrameHandler.sqf
```diff
@@ -14,11 +13,34 @@ GVAR(nextFrameBufferA) = [];
...
+CBA_missionTimeStr = "0000.000000";
+
 // per frame handler system
 [QFUNC(onFrame), {
     SCRIPT(onFrame);
     private _tickTime = diag_tickTime;
-    call FUNC(missionTimePFH);
+
+    // Update CBA_missionTime when synchronized, time has started, and game is not paused.
+    if ([GVAR(missionTimeSynchronized), time > 0, isGamePaused] isEqualTo [true, true, false]) then {
```
mharis001, line 28: I think this is better written as:
```suggestion
    if (GVAR(missionTimeSynchronized) && {time > 0 && {!isGamePaused}}) then {
```

commy2: Isn't that slower though? That shit runs every frame.

mharis001: It is slightly faster.
```
Result:
0.0028 ms

Cycles:
10000/10000

Code:
[cba_common_missionTimeSynchronized, time > 0, isGamePaused] isEqualTo [true, true, false]
```
```
Result:
0.0021 ms

Cycles:
10000/10000

Code:
cba_common_missionTimeSynchronized && {time > 0 && {!isGamePaused}}
```

commy2: What was `cba_common_missionTimeSynchronized` in your example?` It will be `true` all mission long.

mharis001: It was `true`; the overall result for both statements was also `true`.

dedmen: hows the speed without the lazy eval?
creating array, calling "true" command 3 times, creating another array, calling isEqualTo command surely is a bit expensive.
But lazy eval is also expensive but in this case as commy said, we don't expect neither missionTimeSynchronized or the time check to ever be false, so why lazy eval?

mharis001: Yeah, no lazy eval is the fastest.

```
Result:
0.0017 ms

Cycles:
10000/10000

Code:
cba_common_missionTimeSynchronized && time > 0 && !isGamePaused
```

Updated the suggestion.

jonpas:
Status?

jaj22:
If this PR is stuck in limbo, I think a simpler option that fixes the main issue would be to replace the `time != GVAR(lastTime)` lines with `!isGamePaused`.

We're seeing players with significant ACE progress bar slowdown in under two days now, I guess because players have much higher frame rates than they used to (thx to Dedmen and X3D).

PR #1325, opened 2020-04-13
Add CBA_Init attribute to editor

veteran29:
This attribute holds the code of Init box as an string

- title
- editor attribute with content of `Init` box as string

ToDo:
- [x] Objects
- [x] Groups
- [x] Logics

I was unsure how to name the controls.

### DIFF
--- a/addons/common/Cfg3DEN.hpp
+++ b/addons/common/Cfg3DEN.hpp
@@ -142,7 +142,7 @@ class Cfg3DEN {
                     // the usage of local variables and return values.
                     onKillFocus = QUOTE(\
                         private _code = ctrlText (_this select 0);\
-                        (uiNamespace getVariable [ARR_2('GVAR(InitAttributeValue)', controlNull)]) ctrlSetText _code;\
+                        (ctrlParent (_this select 0) getVariable [ARR_2('GVAR(InitAttributeValue)', controlNull)]) ctrlSetText _code;\
                         if (_code != '') then {\
                             _code = 'call{' + _code + '}';\
                         };\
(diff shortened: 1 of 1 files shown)


addons/common/Cfg3DEN.hpp
```diff
@@ -129,14 +129,56 @@ class Cfg3DEN {
...
+                    // variant and into CBA_Init attribute.
+                    // Automatically adds call-block wrapper to enable
+                    // the usage of local variables and return values.
+                    onKillFocus = QUOTE(\
+                        private _code = ctrlText (_this select 0);\
+                        (uiNamespace getVariable [ARR_2('GVAR(InitAttributeValue)', controlNull)]) ctrlSetText _code;\
+                        if (_code != '') then {\
+                            _code = 'call{' + _code + '}';\
+                        };\
+                        (ctrlParentControlsGroup (_this select 0) controlsGroupCtrl 100) ctrlSetText _code;\
```
commy2, line 149: Don't use the UI name space, but a parent control / display instead. Whatever is more appropriate.
I don't see there part where it strips the `call wrapper` ?

veteran29: The onLoad code is inherited.

veteran29:
I could not find a way to get parent control of group of attributes so this got a little bit more complicated and I had to use "specialized" controls for object, group and logic to prevent issues when mixed types are edited at once.

veteran29:
Found hopefully last issue:
image
when this is checked on Init attribute I need to enable CBA_Init attribute too. (when editing multiple items of same type)

addons/common/Cfg3DEN.hpp
```diff
@@ -129,14 +129,127 @@ class Cfg3DEN {
...
+                    );
+                };
+            };
+        };
+
+        class GVAR(InitAttribute_Object): GVAR(EditCodeMulti3) {
```
mharis001, line 199: You shouldn't need to use an edit control if this just a dummy that saves a copy of the init's value - can use an empty controls group.

veteran29:
After discussion on slack this seems use case specific and too dirty for CBA. Closing.

PR #1327, opened 2020-04-25
Settings - Fix localized category duplicating

Dystopian:
- title.

Settings support both localized and non-localized category. If add one setting with localized string and other with non-localized (but same when both are localized), there will be 2 same categories in addon list. This PR fixes it.

### DIFF
--- a/addons/settings/fnc_export.sqf
+++ b/addons/settings/fnc_export.sqf
@@ -38,10 +38,6 @@ private _temp = [];
 
     (GVAR(default) getVariable _setting) params ["_defaultValue", "", "", "", "_category"];
 
-    if (isLocalized _category) then {
-        _category = localize _category;
-    };
-
     if (_exportDefault || {
         !(_value isEqualTo _defaultValue) ||
         SANITIZE_PRIORITY(_setting,_priority,_source) > SANITIZE_PRIORITY(_setting,0,_source)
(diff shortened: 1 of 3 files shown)


commy2:
Are you sure you can use Chinese letters etc. in variable names (setVariable)?

Dystopian:
I'm not sure but I would avoid such variable names. Anyway I didn't find any `setVariable` with category as name.

commy2:
https://github.com/CBATeam/CBA_A3/blob/master/addons/settings/fnc_gui_addonChanged.sqf#L15
https://github.com/CBATeam/CBA_A3/blob/master/addons/keybinding/fnc_initDisplayConfigure.sqf

Dystopian:
Then close. Should be documented btw.

Dystopian:
> Not to use `localize` should be documented?

yeah

commy2:
Why not fix the issue another way?

commy2:
I mean, it does appear to be a problem (it should also happen in keybinding, as the logic is copied from there).

Dystopian:
> Why not fix the issue another way?

because it would require to dig into CBA settings code. For me it's easier to fix on my side, see https://github.com/acemod/ACE3/pull/7613/commits/f842d229441d5a69fd27386c809f003d28208dc9.
Maybe better solution would be documenting using non-localized or English-only strings for category. And we could fix it in all ACE settings in one PR.

> I mean, it does appear to be a problem (it should also happen in keybinding, as the logic is copied from there).

Yes, problem can happen (as it happened to me) when different addons/mods create settings in one category. Or when settings in one addon are created in different times. And error can be seen not right away.

PR #1328, opened 2020-04-25
Inventory - Add context menu options for pointers and optics

veteran29:
- title

https://i.imgur.com/1X6U2Xq.mp4


commy2:
Appears to work.

commy2:
>This branch has not been deployed

dedmen:
How about "next state" instead just add a action for each specific state?

veteran29:
I don't understand. Can you elaborate?

PabstMirror:
If an attachment can be 3 different things [light, IR laser, green laser]
then list all possibilities, instead of "next/previous"

dedmen:
Yes that.
Instead of just "Next State"
"Switch to Red Laser"

But we don't have nested context menu? So fear is that it might clutter the menu with too many entries?

jonpas:
> But we don't have nested context menu? So fear is that it might clutter the menu with too many entries?

It might for attachments with many states, but less so if only available once attached.

veteran29:
Feel free to do another PR then. I think it's just fluff, attachments rarely have more than 3 states, you can click through them really quickly as the menu is set to not close automatically.

PR #1329, opened 2020-04-27
Item Rename Framework

BrettMayson:
- Add a way to rename items during the mission
- Intended for unique items with IDs, but not required (Dogtags, TFAR, ACRE, etc)

**Before**
image

**After**
image

**Known Issues**:

Like the Item Context Menu, this does not work on corpses.


PabstMirror:
- maybe mouseMoving/mouseHolding instead of pfeh? either is probably fine

- there might be a way to do this and keep localization? (ie multi-language clients)
broadcast name as "%1 [commy]"
and then clients do `format [_name, getText (classname>>dispalyName]`
so you can optionally include `%1` to get original name and append custom stuff how you want

commy2:
>there might be a way to do this and keep localization?

renamedItems namespace and function should be local.

synixebrett:
The reason I used the PFH instead of onDrag / MouseMoving is because if you are in a container, and someone else drops the item in it, you will not see the updated name until you do something that causes it to update. The PFH handles both dragging and items being added externally.

commy2:
MouseHolding + MouseMoving fire every frame.

synixebrett:
Every frame? Even when you aren't doing anything?

commy2:
One of them fires when you move your mouse, the other when you hold your mouse. This covers every frame.

commy2:
https://community.bistudio.com/wiki/User_Interface_Event_Handlers#onMouseHolding

synixebrett:
What if I am not moving my mouse? That is the part I don't get how it covers every frame. Does that event fire if the mouse is not moving?

I'll try it out tonight anyway

PabstMirror:
you add event to **both** mouse moving and mouse holding
1 of the 2 is guaranteed to run each frame

synixebrett:
Yeah I get to add both, I read commy's link too fast as I'm at work right now. In my head `mouseHolding` was holding the mouse button down... Which is why I was so confused, I see clearly now

synixebrett:
I have been unable to get this working with display event handlers on MouseHolding and MouseMoving, not entirely sure what is going wrong but I may need someone else to take a crack at it

commy2:
There. That is how we mean.

synixebrett:
Yeah I did something similar, but the `getInventoryClassname` stopped working and I couldn't figure out why

SzwedzikPL:
@commy2 please review current version.

addons/ui/fnc_renameInventoryItem.sqf
```diff
@@ -0,0 +1,37 @@
...
+Returns:
+    Nothing
+
+Authors:
+    SynixeBrett
+---------------------------------------------------------------------------- */
+
+params [
+  ["_class", "", [""]],
+  ["_name", "", [""]]
```
commy2, line 26: 4 indents

commy2:
Should be able to set the picture as well.

SzwedzikPL:
@commy2 Done. I tried earlier my luck with changing the tooltip but so far i've been unable to find a way to do that. It looks like it comes directly from the engine.

I've been also thinking about supporting `nil` as a value in `CBA_fnc_renameInventoryItem` so `""` can be used to remove item name or icon. What do you think?

commy2:
Yeah, couldn't get the tooltips to work either. Probably some hard coded stuff managing the inventory ui tooltips that prevents scripted tooltips from working, sadly. Could always try to create custom tooltip controls though.

nil already converts to "" due to params command. Since empty stings make no sense (and for picture you could use the empty_ca.paa) it should treat "" as "revert to default".

SzwedzikPL:
Ok, for now i think vanilla tooltips are fine, we can always add some workaround for tooltips later.

Any other suggestions?
Also, is there a chance for this to release in 3.15.2 ?

commy2:
There're no plans for anything lol

commy2:
Is this done?

synixebrett:
Afaik, yes.

The only outstanding issue is the bug with corpses, the same issue as the item context menu framework.

SzwedzikPL:
@commy2 do you have any idea how to properly support items inside ground/cargo backpacks? `InventoryOpened` EH doesn't work in such case, nor `ContainerOpened`. I found some ugly way using loops and [everyBackpack](https://community.bistudio.com/wiki/everyBackpack) but it doesn't look like good/proper solution.

commy2:
Why does `InventoryOpened` not work?

commy2:
The correct event name is "ContainerOpened". You can get which one is opened by tracking double click on the container list.

addons/ui/fnc_getInventoryItemData.sqf
```diff
@@ -0,0 +1,58 @@
...
+    };
+    case "VEST_CONTAINER": {
+        _container = vestContainer _unit;
+    };
+    case "BACKPACK_CONTAINER": {
+        _container = backpackContainer _unit;
+    };
+};
+
+// Reports classname, but only for magazines.
```
johnb432, line 48: When I redid the inventory filtering in https://github.com/acemod/ACE3/pull/9706, I found all item classnames are contained in `lbData` except for backpacks:
https://github.com/acemod/ACE3/blob/b714c8bce2628ff8d4795da8a7f2fc05bb82f474/addons/inventory/functions/fnc_forceItemListUpdate.sqf#L33

Workaround for backpacks is to get them using the display name and picture.

addons/ui/fnc_renameInventoryItem.sqf
```diff
@@ -0,0 +1,40 @@
...
+    ["_name", "", [""]],
+    ["_picture", "", [""]]
+];
+
+// Exit if no interface or empty class
+if (!hasInterface || {_class isEqualTo ""}) exitWith {};
+
+// Create local namespace on first use
+if (isNil QGVAR(renamedItems)) then {
+    GVAR(renamedItems) = false call CBA_fnc_createNamespace;
```
johnb432, line 37: Update to hashmap. It means that the input would have to be made config-case, but if `configName (_class call CBA_fnc_getItemConfig)` is run at the start, it would allow us to detect invalid classes, as it would return `""`.

addons/ui/fnc_initDisplayInventory.sqf
```diff
@@ -228,44 +231,54 @@ _uniformItems setVariable [QGVAR(containerType), "UNIFORM_CONTAINER"];
...
+            if (ctrlShown _control) then {
+                for [{_i = 0}, {_i < (lbSize _control)}, {_i = _i + 1}] do {
+                    private _classname = ([_control, _i] call FUNC(getInventoryItemData)) select 0;
+                    if !(isNil "_classname") then {
+                        private _renameParams = GVAR(renamedItems) getVariable [_classname, []];
+
+                        if !(_renameParams isEqualTo []) then {
+                            _renameParams params ["_name", "_picture"];
+
+                            if !(_name isEqualTo "") then {
```
johnb432, line 268: I prefer `!=`, but at least update to `isNotEqualTo`:
```suggestion
                            if (_name != "") then {
```

Update where necessary.

johnb432 (commented):
@BrettMayson Can you resolve the merge conflicts?

- You can't rename backpacks atm. It might be possible to add it without game engine changes, but I'm not certain.
- https://github.com/acemod/ACE3/pull/10102#discussion_r1662176263
   Which method is better - PFH, mouse EH or map EH? We should try to decide on a single method and stick with that imo, to make it more uniform.

johnb432:
> It might be possible to add it without game engine changes, but I'm not certain.

Not possible atm, made a ticket: https://feedback.bistudio.com/T183096

PabstMirror (approved):

PabstMirror:
I'd like to merge now even if backpacks aren't working, 
thoughts?

johnb432:
It might be possible to just add `backpackCargo` to https://github.com/CBATeam/CBA_A3/pull/1329/files#diff-32e5060e1f3d15868d5a7eb7582a521ea990945ef5d516056f0c38bf16799fa5R49-R56 and have it work like that. If so, it would be a more robust solution than what ACE atm.

addons/ui/fnc_getInventoryItemData.sqf
```diff
@@ -0,0 +1,58 @@
...
+private _classname = _control lbData _index;
+if (_classname isEqualTo "") then {
+    // For weapons, items and glasses, use the lb index and compare with cargo item list.
+    private _cargoItems = weaponCargo _container + itemCargo _container + magazineCargo _container;
```
johnb432, line 52:
```suggestion
    private _cargoItems = weaponCargo _container + itemCargo _container + magazineCargo _container + backpackCargo _container;
```

johnb432:
> Can DLC or other atypical items pose issues?

Yes, atypical items can pose issues. Repro: I loaded ACE, CUP Weapons and RHS and put some mines from each mod into my backpack in the ACE arsenal (I had 13 items total in backpack, 1 base game training mine and 12 from CUP/RHS, I took 5 mines from 1 category, rest were 1 each). Restarted the game without CUP Weapons and RHS and went into the ACE arsenal. There, I still had 13 items, but 12 of them were `""`. In the backpack, 8 rows were filled with the "Premium item" thing that you get when you have non-existing item.
I added a backpack, renamed to `"test"` and added the backpack in my backpack. When I opened my backpack, the 3rd row was renamed from `""` to `"test"`.

```sqf
_container = backpackContainer player;
weaponCargo _container + itemCargo _container + magazineCargo _container + backpackCargo _container;
```
returns `["TrainingMine_Mag","","","","","","","","","","","","","B_AssaultPack_rgr"]`, so it makes sense.

Any item renamed through the framework is affected by this: I tested it by trying to rename the training mine I had - this time it was the top most (again, makes sense).

There might be a way around this very specific issue though: `getMagazineCargo` is able to tell how many of each invalid classes there are. I'll experiment with the `getXCargo` commands to see if I can get it to work (would avoid us having to `arrayIntersect` the result).

johnb432:
The `getXCargo` commands resolve the issue.

However, there are still 2 issues:
1) I renamed the training mag to `"test 2"`. Because `_control lbData _index` returns `""`, it also renames the first invalid item to `"test 2"`.
2) if you have 2 weapons of the same class, but they don't have the same attachments, then they won't stack in the inventory, but they are reported as the same class (`getWeaponCargo` returns `2`, `weaponCargo` returns it twice).
I don't think using inventory commands to determine the order of the UI is the way forward. Imo we should not support backpacks atm, remove the empty classname stuff and hope the BI ticket will be resolved.

addons/ui/fnc_getInventoryItemData.sqf
```diff
@@ -0,0 +1,58 @@
...
+    _cargoItems = _cargoItems arrayIntersect _cargoItems;
+
+    _classname = _cargoItems param [_index, ""];
+};
```
johnb432, line 56:
```suggestion
```

addons/ui/fnc_renameInventoryItem.sqf
```diff
@@ -0,0 +1,44 @@
...
+// Create local namespace on first use
+if (isNil QGVAR(renamedItems)) then {
+    GVAR(renamedItems) = createHashMap;
+};
+
+private _config = _class call CBA_fnc_getItemConfig;
+if (isNull _config) exitWith { ERROR_1("Class [%1] does not exists",_class) };
+_class = configName _config; // Ensure class name is in config case
+
+GVAR(renamedItems) set [_class, [_name, _picture]];
```
johnb432, line 44: It otherwise returns `false`.
```suggestion
GVAR(renamedItems) set [_class, [_name, _picture]];

nil
```

addons/ui/fnc_getInventoryItemData.sqf
johnb432, line 48: // Reports classname for every item type except backpacks (https://feedback.bistudio.com/T183096)

johnb432 (changes requested):

addons/ui/fnc_getInventoryItemData.sqf
johnb432, line 48:
```suggestion
// Reports classname for every item type except backpacks (https://feedback.bistudio.com/T183096)
```

johnb432: > Workaround for backpacks is to get them using the display name and picture.

That would likely cause ambiguities for backpacks with linked items, as I doubt they would change the picture and display name for each variant.

PR #1330, opened 2020-05-01
Add CBA_fnc_canAddItem

Dystopian:
- add function for checking if unit has enough free space in inventory to store item.

`canAdd*` commands can return `false` even if there is free space in unit inventory (see [this ticket](https://feedback.bistudio.com/T120627) and other about `canAdd`). These commands take current unit load into account. Sometimes we don't need to check unit load e.g. when taking out ACE Earplugs or weapon accessories.

As usually function and variables names are subject to discuss.

### DIFF
--- a/addons/common/fnc_canAddItem.sqf
+++ b/addons/common/fnc_canAddItem.sqf
@@ -38,7 +38,7 @@ params [
     ["_checkBackpack", true, [false]]
 ];
 
-if (_unit isEqualTo objNull || {_item isEqualTo ""}) exitWith {false};
+if (isNull _unit || {_item isEqualTo ""}) exitWith {false};
 
 #define TYPE_VEST 701
 #define TYPE_UNIFORM 801
--- a/addons/common/fnc_canAddItem.sqf
+++ b/addons/common/fnc_canAddItem.sqf
@@ -100,9 +100,12 @@ if (
     _checkUniform
     && {TYPE_UNIFORM in _allowedSlots}
     && {
-        // each time subtract whole number of items which can be put in container
-        _count = _count - floor (getContainerMaxLoad uniform _unit * (1 - loadUniform _unit) / _mass);
-        _count <= 0
+        _mass == 0
+        || {
+            // each time subtract whole number of items which can be put in container
+            _count = _count - floor (getContainerMaxLoad uniform _unit * (1 - loadUniform _unit) / _mass);
+            _count <= 0
+        }
     }
 ) exitWith {true};
 
@@ -110,8 +113,11 @@ if (
     _checkVest
     && {TYPE_VEST in _allowedSlots}
     && {
-        _count = _count - floor (getContainerMaxLoad vest _unit * (1 - loadVest _unit) / _mass);
-        _count <= 0
+        _mass == 0
+        || {
+            _count = _count - floor (getContainerMaxLoad vest _unit * (1 - loadVest _unit) / _mass);
+            _count <= 0
+        }
     }
 ) exitWith {true};
 
@@ -119,8 +125,11 @@ if (
     _checkBackpack
     && {TYPE_BACKPACK in _allowedSlots}
     && {
-        _count = _count - floor (getContainerMaxLoad backpack _unit * (1 - loadBackpack _unit) / _mass);
-        _count <= 0
+        _mass == 0
+        || {
+            _count = _count - floor (getContainerMaxLoad backpack _unit * (1 - loadBackpack _unit) / _mass);
+            _count <= 0
+        }
     }
 ) exitWith {true};
 
--- a/addons/common/fnc_canAddItem.sqf
+++ b/addons/common/fnc_canAddItem.sqf
@@ -100,8 +100,9 @@ if (
     _checkUniform
     && {TYPE_UNIFORM in _allowedSlots}
     && {
-        private _maxLoad = getContainerMaxLoad uniform _unit;
-        _maxLoad >= _mass * _count + _maxLoad * loadUniform _unit
+        // each time subtract whole number of items which can be put in container
+        _count = _count - floor (getContainerMaxLoad uniform _unit * (1 - loadUniform _unit) / _mass);
+        _count <= 0
     }
 ) exitWith {true};
 
@@ -109,8 +110,8 @@ if (
     _checkVest
     && {TYPE_VEST in _allowedSlots}
     && {
-        private _maxLoad = getContainerMaxLoad vest _unit;
-        _maxLoad >= _mass * _count + _maxLoad * loadVest _unit
+        _count = _count - floor (getContainerMaxLoad vest _unit * (1 - loadVest _unit) / _mass);
+        _count <= 0
     }
 ) exitWith {true};
 
@@ -118,8 +119,8 @@ if (
     _checkBackpack
     && {TYPE_BACKPACK in _allowedSlots}
     && {
-        private _maxLoad = getContainerMaxLoad backpack _unit;
-        _maxLoad >= _mass * _count + _maxLoad * loadBackpack _unit
+        _count = _count - floor (getContainerMaxLoad backpack _unit * (1 - loadBackpack _unit) / _mass);
+        _count <= 0
     }
 ) exitWith {true};
 
--- a/addons/common/fnc_canAddItem.sqf
+++ b/addons/common/fnc_canAddItem.sqf
@@ -1,6 +1,6 @@
 #include "script_component.hpp"
 /* ----------------------------------------------------------------------------
-Function: CBA_fnc_unitHasSpace
+Function: CBA_fnc_canAddItem
 
 Description:
     Checks if unit has enough free space in inventory to store item.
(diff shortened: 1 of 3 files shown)


commy2:
CBA_fnc_canAddItem?

addons/common/fnc_unitHasSpace.sqf
```diff
@@ -0,0 +1,129 @@
...
+};
+
+private ["_mass", "_allowedSlots"];
+private _itemMassAllowedSlots = GVAR(itemMassAllowedSlots) getVariable _item;
+
+if (!isNil "_itemMassAllowedSlots") then {
+    _mass = _itemMassAllowedSlots # 0;
+    _allowedSlots = _itemMassAllowedSlots # 1;
+    TRACE_3("cache found",_item,_mass,_allowedSlots);
+} else {
```
commy2, line 57:
```sqf
GVAR(itemMassAllowedSlots) getVariable [_item, []] params ["_mass", "_allowedSlots"];

if (isNil "_mass") then {
```

addons/common/fnc_unitHasSpace.sqf
```diff
@@ -0,0 +1,129 @@
...
+            };
+        } else {
+            private _cfgGlassesItem = configFile >> "CfgGlasses" >> _item;
+            if (isClass _cfgGlassesItem) then {
+                _mass = getNumber (_cfgGlassesItem >> "mass");
+            } else {
+                _mass = -1;
+                _allowedSlots = [];
+            }
+        };
```
commy2, line 94: Instead of nested else if, why not exitWith or switch?

addons/common/fnc_unitHasSpace.sqf
```diff
@@ -0,0 +1,129 @@
...
+    };
+    TRACE_3("caching",_item,_mass,_allowedSlots);
+    GVAR(itemMassAllowedSlots) setVariable [_item, [_mass, _allowedSlots]];
+};
+
+if (_mass == -1) exitWith {false};
```
commy2, line 100: This means item doesn't exist, right? Should add as comment.

addons/common/fnc_unitHasSpace.sqf
```diff
@@ -0,0 +1,129 @@
...
+
+if (
+    _checkUniform
+    && {
+        private _maxLoad = getContainerMaxLoad uniform _unit;
+        _maxLoad >= _mass * _count + _maxLoad * loadUniform _unit
+    }
+    && {TYPE_UNIFORM in _allowedSlots}
+) exitWith {true};
```
commy2, line 109:
```sqf
if (_checkUniform && {TYPE_UNIFORM in _allowedSlots && {
    private _maxLoad = getContainerMaxLoad uniform _unit;
    _maxLoad >= _mass * _count + _maxLoad * loadUniform _unit
}}) exitWith {true};
```
etc.
?

addons/common/fnc_unitHasSpace.sqf
Dystopian: Can't exitWith because setVariable to cache. Just tested switch and it looks quite ugly because of `_cfgWeaponsItem _cfgMagazinesItem_cfgGlassesItem` vars assignment.

addons/common/fnc_unitHasSpace.sqf
Dystopian: looks not regular and less maintainable and IIRC from your tests nested condition doesn't give really much performance over just lazy one

addons/common/fnc_unitHasSpace.sqf
Dystopian: this was my first version but without `, []` part so I got error and had to rewrite it. Thanks :)

Dystopian:
> CBA_fnc_canAddItem?

I thought about it or even just `CBA_fnc_canAdd` but I didn't want to confuse it with `canAdd` command. Anyway for me naming here is not important so I accept any suggestions.

commy2:
The purpose is the same as the canAdd command though, and it is obvious that they behave differently, since one is a command, the other a function from CBA.

addons/common/fnc_unitHasSpace.sqf
commy2: I still think the in _allowedSlot check should be second.

addons/common/fnc_unitHasSpace.sqf
commy2: Nested else if looks ugly too.

addons/common/fnc_unitHasSpace.sqf
```diff
@@ -0,0 +1,124 @@
...
+        [player, "acc_flashlight"] call CBA_fnc_unitHasSpace
+        [player, "30Rnd_556x45_Stanag", 7, false, true, false] call CBA_fnc_unitHasSpace
+    (end)
+
+Author:
+    Dystopian
+---------------------------------------------------------------------------- */
+SCRIPT(unitHasSpace);
+
+params [
```
mharis001, line 32: Should add type checking to all parameters since this is a public function.

addons/common/fnc_canAddItem.sqf
```diff
@@ -0,0 +1,126 @@
...
+params [
+    ["_unit", objNull, [objNull]],
+    ["_item", "", [""]],
+    ["_count", 1, [0]],
+    ["_checkUniform", true, [false]],
+    ["_checkVest", true, [false]],
+    ["_checkBackpack", true, [false]]
+];
+
+if (_unit isEqualTo objNull || {_item isEqualTo ""}) exitWith {false};
```
mharis001, line 41:
```suggestion
if (isNull _unit || {_item isEqualTo ""}) exitWith {false};
```

dedmen:
This has the same issue as canAdd command.
Lets say you add 30 items with 1 space each.
In uniform, you have 12 spaces free, in vest you have 20 spaces free, in backpack you have 29 spaces free.

Yes you can add the 30 items, but canAdd returns false.

My idea for the solution would be to not check if count of items fits, but instead for uniform/vest/backpack instead count how many of the items could fit into each of them, and the sum the number up and check if the possible number of items that can fit is bigger than the requested count.

Dystopian:
`count` made function a little bit more complex. Maybe function should return item count which can be put to unit? It would make function more universal.

dedmen:
I think instead of all the seperate exitWiths and stuff, its easier to just count up the total that fit and do one check at the end.
Though less efficient of course 🙃

commy2:
lgtm.
Could add another abstraction layer by putting those container checks into a function and passing vest, vestLoad, checkVest and allowSlotVest etc. as arguments.
Would look nicer for sure.

Dystopian:
I thought about it. But for 3 containers I decided it's not worth it.

addons/common/fnc_canAddItem.sqf
```diff
@@ -0,0 +1,127 @@
...
+            _mass = -1;
+            _allowedSlots = [];
+        };
+    };
+    TRACE_3("caching",_item,_mass,_allowedSlots);
+    GVAR(itemMassAllowedSlots) setVariable [_item, [_mass, _allowedSlots]];
+};
+
+if (_mass == -1) exitWith {false}; // item doesn't exist
+
```
Dystopian, line 98: note for me: add `mass==0` check

addons/common/fnc_canAddItem.sqf
```diff
@@ -100,27 +100,36 @@ if (
...
     _checkBackpack
     && {TYPE_BACKPACK in _allowedSlots}
     && {
-        _count = _count - floor (getContainerMaxLoad backpack _unit * (1 - loadBackpack _unit) / _mass);
-        _count <= 0
+        _mass == 0
+        || {
+            _count = _count - floor (getContainerMaxLoad backpack _unit * (1 - loadBackpack _unit) / _mass);
+            _count <= 0
+        }
```
commy2, line 132: Can't you check this once before checking the 3 containers?

Dystopian: no, it can have mass==0 and allowedSlots not with all containers

commy2: I see.

commy2:
```sqf
#define CAN_ADD_TO(CONTAINER, CONTAINER_TYPE) (CONTAINER_TYPE in _allowedSlots && {\
    _mass == 0 || {\
        _count = _count - floor (getContainerMaxLoad CONTAINER * (1 - load##CONTAINER) / _mass);\
        _count <= 0\
    }\
})\


_checkUniform && {CAN_ADD_TO(uniform _unit, TYPE_UNIFORM)} ||
{_checkVest && {CAN_ADD_TO(vest _unit, TYPE_VEST)}} ||
{_checkBackpack && {CAN_ADD_TO(backpack _unit, TYPE_BACKPACK)}}
```
🤔

commy2:
Even if you hate it, I am having fun.
```sqf
#define CAN_ADD_TO(CONTAINER) _check##CONTAINER && {
    DOUBLES(TYPE,CONTAINER) in _allowedSlots && {\
        _mass == 0 || {\
            _count = _count - floor (getContainerMaxLoad CONTAINER _unit * (1 - load##CONTAINER _unit) / _mass);\
          _count <= 0\
        }\
    }\
}


CAN_ADD_TO(UNIFORM) || {CAN_ADD_TO(VEST) || {CAN_ADD_TO(BACKPACK)}}
```

PabstMirror (approved):

Freddo3000:
This has already been merged, but it seems to be tied to this: https://feedback.bistudio.com/T154476

`canAdd` returning false doesn't have to do with having over a certain amount of items, it has to do with having more than 1000 `loadAbs`, or 1.00 `load` on a unit, corresponding to `maxSoldierLoad`. It seems to possibly be a game bug that you can go over this limit.

commy2:
Sure, but this does not change anything in this PR, does it?

Freddo3000:
Not really, just adding additional details. Depending on whether the ticket is fixed later, this function may become redundant or invalid, though I wouldn't hold my breath at this stage of the game.

PR #1333, opened 2020-05-08
add extension to write to cba_settings.sqf file

commy2:
- title

It would be super convenient if settings could be written by the game to the settings files directly. So I wanted to have an extension for that. Pardon my poor C++, I have no idea what's going on, lol

`READ` -> reads setting value & force lvl
`WRITE` -> overwrites setting with given value & force lvl
`CLEAR` -> empties cba_settings.sqf
`PARSE` -> gives a `parseSimpleArray`'able super array of all settings, value, force lvl triplets. Since extension callback is limited by like 17k chars, this comes in batches of 10k. The idea is to append the 10k batches in SQF before using PSA.

Currently only works for the main game userconfig, not for missions (need to figure out the system path).


extensions/cba_settings/cba_settings.cpp
```diff
@@ -0,0 +1,335 @@
...
+    string result = "";
+
+    ifstream file;
+    file.open(SETTINGS_FILE, ios::in);
+
+    string line;
+    while (file.good()) {
+        getline(file, line);
+        if (file.good() || line.length() != 0) // Don't explode file by appending duplicate additional newlines.
+            result = result + trim(line) + "\n";
```
dedmen, line 50:
```suggestion
            result.append(trim(line) + "\n");
```

extensions/cba_settings/cba_settings.cpp
```diff
@@ -0,0 +1,335 @@
...
+
+string clear() {
+    ofstream file;
+    file.open(SETTINGS_FILE, ios::out);
+    file.close();
+    return "true";
+}
+
+string load(int strlen) {
+    string result = "";
```
dedmen, line 41: reserve.

`result.reserve(std::filesystem::size(SETTINGS_FILE));`

Otherwise you resize/reallocate/copy almost every time you append a line.

extensions/cba_settings/cba_settings.cpp
```diff
@@ -0,0 +1,335 @@
...
+}
+
+string clear() {
+    ofstream file;
+    file.open(SETTINGS_FILE, ios::out);
+    file.close();
+    return "true";
+}
+
+string load(int strlen) {
```
dedmen, line 40: strlen is the name of a function, please don't use that as variable name.

extensions/cba_settings/cba_settings.cpp
```diff
@@ -0,0 +1,335 @@
...
+    ifstream file;
+    file.open(SETTINGS_FILE, ios::in);
+
+    string line;
+    while (file.good()) {
+        getline(file, line);
+        line = trim_left(line);
+
+        // Ignore commented out lines.
+        if (line.substr(0, 2) == "//")
```
dedmen, line 74: This copies the string every time. Would be WAY more efficient to use a string_view here, as that is all that you need. For trim_left too.

dedmen: Also why trim left and right seperately? why not immediately trim both sides?

extensions/cba_settings/trim.h
```diff
@@ -0,0 +1,33 @@
...
+string trim_right(string str) {
+    size_t last = str.find_last_not_of(WHITESPACE);
+
+    if (last == string::npos)
+        return "";
+    else
+        return str.substr(0, last + 1);
+}
+
+string trim(string str) {
```
dedmen, line 25: Example for a more efficient trim that doesn't copy the string
https://github.com/dedmen/armake/blob/cpp/src/utils.cpp#L240

extensions/cba_settings/cba_settings.cpp
```diff
@@ -0,0 +1,335 @@
...
+
+    string value = "";
+    if (argsCnt >= 2)
+        value = args[1];
+
+    int force = 0;
+    if (argsCnt >= 3)
+        force = min(stoi(args[2]), 2);
+
+    if (_stricmp(function, "version") == 0) {
```
dedmen, line 297: why even check the whole string?
c
l
p
r
a

the first character would be sufficient for what you need

extensions/cba_settings/cba_settings.cpp
```diff
@@ -0,0 +1,335 @@
...
+        // Get setting.
+        size_t i = line.find("=");
+        if (i == string::npos) // Remove ill-formed line.
+            continue;
+
+        string item = line.substr(0, i);
+        item = trim_right(item);
+
+        if (_stricmp(item.c_str(), setting.c_str()) == 0) {
+            // Upade matching lines.
```
dedmen, line 225:
```suggestion
            // Updade matching lines.
```

extensions/cba_settings/cba_settings.cpp
```diff
@@ -0,0 +1,335 @@
...
+    file << line << endl;
+    file.close();
+    return "true";
+}
+
+string write(string setting, string value, int force) {
+    string result = "false";
+
+    ofstream temp;
+    temp.open(string(TEMP_FILE).c_str(), ios::out);
```
dedmen, line 186: what is the temp file used for here?

dedmen, line 186: you are reloading, reparsing, rewriting the whole file for every setting property write?
You could just parse it once, and store it in memory. Your extension won't get unloaded

extensions/cba_settings/cba_settings.cpp
```diff
@@ -0,0 +1,335 @@
...
+        getline(file, line);
+        line = trim_left(line);
+
+        // Ignore commented out lines.
+        if (line.substr(0, 2) == "//")
+            continue;
+
+        // Count force tags.
+        int force = 0;
+        while (line.substr(0, 6) == "force ") {
```
dedmen, line 134: you are copying the substring on every iteration here. If you use the string_view trim, it will only move a pointer around

dedmen:
I would

struct settingEntry {
uint8_t force;
std::string settingValue;
}

std::unordered_map<std::string, settingEntry> settings;


on parse then clear map and parse entries into it.
on update you just update the entry, and write out all others. No need to reparse file.

extensions/cba_settings/cba_settings.cpp
commy2: >store it in memory
How?

commy2:
>is it? I thought callbacks were unlimited

It only gave me that long a string (17 something k characters) when I tried how large a settings file I can get.

extensions/cba_settings/cba_settings.cpp
commy2: True I guess, but what does it matter? This is not something that needs to be particularly fast.

extensions/cba_settings/cba_settings.cpp
commy2: >why not immediately trim both sides?

```
    force force   setting    =    123;
```
is something one could write. If I process this, I need multiple trims, but not on both sides all the time.

extensions/cba_settings/cba_settings.cpp
commy2: Haha.

extensions/cba_settings/cba_settings.cpp
commy2: Cool, didn't know `filestream::size` works is a thing. Though, is it reasonable to allocate ~10k characters (bytes?) at once?

commy2:
>I would
>unordered_map

But what if I edit the file while the game is running?

extensions/cba_settings/cba_settings.cpp
dedmen: Well more reasonable than first allocating 16, then deleting and reallocating 32, then deleting and reallocating 64, then deleting and reallocating 128, then deleting and reallocating 512, then deleting and reallocating 2048, then deleting and reallocating 8192, then...

extensions/cba_settings/cba_settings.cpp
dedmen: Comment below, just a global variable. Array or map or some container

dedmen:
> But what if I edit the file while the game is running?
store file change date of last parse/write.
Then when reading check file change date, if it was edited you need to reload. If not then not.

extensions/cba_settings/cba_settings.cpp
```diff
@@ -0,0 +1,335 @@
...
+            result = "[\"" + item + "\"," + value + "," + to_string(force) + "]";
+            break;
+        }
+    }
+
+    file.close();
+    return result;
+}
+
+string append(string setting, string value, int force) {
```
TheMagnetar, line 167: Passing by reference is advisable with classes. It  avoids copying the whole class. The same can be applied to the other functions.

```suggestion
string append(const string &setting, const string &value, int force) {
```

extensions/cba_settings/cba_settings.cpp
```diff
@@ -0,0 +1,335 @@
...
+    ofstream file;
+    file.open(SETTINGS_FILE, ios::app);
+
+    string line = "";
+    while (force-- > 0)
+        line = line + "force ";
```
TheMagnetar, line 173:
```suggestion
        line.append("force ");
```

extensions/cba_settings/cba_settings.cpp
```diff
@@ -0,0 +1,335 @@
...
+
+#define RET \
+strncpy_s(output, outputSize, result.c_str(), _TRUNCATE);\
+return
+
+#define RETURN(val)\
+strncpy_s(output, outputSize, result.c_str(), _TRUNCATE);\
+return (val)
+
+#define SYNTAX_ERROR_WRONG_PARAMS_SIZE 101
```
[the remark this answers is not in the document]
commy2: What's the benefit over
```cpp
const int SYNTAX_ERROR_WRONG_PARAMS_SIZE = 101;
```
?

extensions/cba_settings/cba_settings.cpp
commy2: Cool. Didn't know it copies, and that this is how one prevents it.

extensions/cba_settings/cba_settings.cpp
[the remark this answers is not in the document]
TheMagnetar: constexpr is evaluated at compile time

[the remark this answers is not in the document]
dedmen: which won't make any difference here as the end result.

extensions/cba_settings/cba_settings.cpp
dedmen: Doesn't really matter, just my "every nanosecond counts" bullshit.

commy2: I understand.

PR #1335, opened 2020-05-13
JR - Ensure class names returned by compatibleItems are in config case

mharis001:
- title, previously items were not guaranteed to match `CfgWeapons` config case

For example (before this change), `optic_Aco` was returned as `optic_aco`:

```sqf
currentWeapon player; // "arifle_MX_F"
primaryWeaponItems player; // ["","acc_pointer_IR","optic_Aco",""]
primaryWeaponItems player arrayIntersect ([currentWeapon player] call CBA_fnc_compatibleItems); // ["acc_pointer_IR"]
"optic_aco" in ([currentWeapon player] call CBA_fnc_compatibleItems); // true
"optic_Aco" in ([currentWeapon player] call CBA_fnc_compatibleItems); // false
```

### DIFF
--- a/addons/jr/fnc_compatibleItems.sqf
+++ b/addons/jr/fnc_compatibleItems.sqf
@@ -35,9 +35,10 @@ private _compatibleItems = GVAR(namespace) getVariable _weapon;
 
 if (isNil "_compatibleItems") then {
     _compatibleItems = [];
-    private _cfgWeapon = configFile >> "CfgWeapons" >> _weapon;
+    private _cfgWeapons = configFile >> "CfgWeapons";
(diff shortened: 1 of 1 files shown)


commy2:
bwc though

mharis001:
I don't think that is an issue because any case-insensitive tests remain unaffected but case-sensitive tests get fixed to work as expected (config case would be expected since all(?) commands use that).

commy2:
Some functions are all lower case, some are config case. It is a mess and not thought of when this was ported from ASDG to CBA.

PabstMirror (approved):

Issue #1336, opened 2020-05-14
Functions are inaccessible in Eden if RECOMPILE is defined on game start


Dystopian:
**Mods (complete and add to the following information):**
- **Arma 3:** `1.98 stable`
- **CBA:** `3.15.1 dev 5c478a099b0954baf1c0b085861978e52847ce9b`

**Description:**
Sometimes you need CBA functions in Eden editor (e.g. in Debug console). Also sometimes RECOMPILE is needed to edit functions while game is running. But when RECOMPILE is defined in `x\cba\addons\main\script_mod.hpp` on game start the functions are not compiled to missionNamespace and can't be used in console by name. They are compiled to uiNamespace only.
Also this prevents ACE Arsenal working in Eden editor.

**Steps to reproduce:**
- define RECOMPILE in `x\cba\addons\main\script_mod.hpp`;
- start game with filepatching;
- open Eden editor, open console;
- put `CBA_fnc_leftTrim`, `missionNamespace getVariable "CBA_fnc_leftTrim"`, `uiNamespace getVariable "CBA_fnc_leftTrim"` to watch fields. Function exists in uiNamespace and doesn't exist in missionNamespace or by name.

**Expected behavior:**
Functions should exist in missionNamespace. 

**Where did the issue occur?**
- Editor (Singleplayer)

**Log Files:**
- https://gist.github.com/Dystopian/9ea398e6240814ee1bfc846912c75434

commy2:
> Also this prevents ACE Arsenal working in Eden editor.

But only if RECOMPILE is set?

Dystopian:
> But only if RECOMPILE is set?

yes, and only on game start

commy2:
But all RECOMPILE does is set the recompile token to 1 in CfgFunctions. How, when and if a CfgFunctions function like `CBA_fnc_leftTrim` is compiled is up to BI and `\a3\functions_f\initFunctions.sqf`. Idk what to do about that. I'd say you should use PREP().

PR #1337, opened 2020-05-15
XEH - Fix CBA functions not available in editor when using RECOMPILE …

PabstMirror:
…(DEV Only)

Ref #1336

- Describe what this pull request will do
- Each change in a separate line

### DIFF
--- a/addons/xeh/fnc_preInit.sqf
+++ b/addons/xeh/fnc_preInit.sqf
@@ -25,6 +25,18 @@ SLX_XEH_DisableLogging = uiNamespace getVariable ["SLX_XEH_DisableLogging", fals
 
 XEH_LOG("PreInit started. v" + getText (configFile >> "CfgPatches" >> "cba_common" >> "versionStr"));
 
+#ifdef RECOMPILE
+if (isNil "CBA_fnc_compileFunction") then {
+    INFO("Recompiling all CBA Functions [for use in threeden]");
+    { 
+        { 
+            private _fncName = format ["CBA_fnc_%1", configName _x]; 
+            _fncName call BIS_fnc_recompile; 
+        } forEach (configProperties [_x]); 
+    } forEach (configProperties [configFile >> "CfgFunctions" >> "CBA"]); 
+};
+#endif
+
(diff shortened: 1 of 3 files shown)


addons/xeh/fnc_preInit.sqf
```diff
@@ -25,6 +25,18 @@ SLX_XEH_DisableLogging = uiNamespace getVariable ["SLX_XEH_DisableLogging", fals
...
 XEH_LOG("PreInit started. v" + getText (configFile >> "CfgPatches" >> "cba_common" >> "versionStr"));
 
+#ifdef RECOMPILE
+if (isNil "CBA_fnc_compileFunction") then {
+    INFO("Recompiling all CBA Functions [for use in threeden]");
+    { 
+        { 
+            private _fncName = format ["CBA_fnc_%1", configName _x]; 
+            _fncName call BIS_fnc_recompile; 
+        } forEach (configProperties [_x]);
```
commy2, line 35: Before you change this, can this config or the super config contain any other tokens besides classes? Looks like it according to https://community.bistudio.com/wiki/Arma_3_Functions_Library. Though we likely not make use of them.

commy2:
But this only works when setting the macro in the actual CBA header files and then building cba_main, not when editing some addon? I really don't see the point of this.

Dystopian:
Just tested. Works as expected, CBA functions present, ACE Arsenal works.
And at least `SKIP_SCRIPT_NAME` definition is ignored in missionNamespace functions.

commy2:
Why do you uncomment RECOMPILE in cba_main in the first place though?

Dystopian:
Because I want to patch some CBA function and if I understand correctly RECOMPILE works only on game start.

commy2:
Patch for what?

Dystopian:
Patch to modify function, add something, maybe fix some bug, maybe just test/debug. Usual development.

PabstMirror:
`recompile` is pretty much equivalent to `disable_compile_cache`
it does force add scriptname header stuff, but that's only a problem for ultra precise benchmark-ing

PR #1338, opened 2020-05-15
CBA - use PREPMAIN() instead of CfgFunctions

commy2:
- title
- allows recompiling with the PREP stuff, not the BI CfgFunctions stuff

### DIFF
--- a/addons/accessory/CfgFunctions.hpp
+++ b/addons/accessory/CfgFunctions.hpp
@@ -1,7 +1,0 @@
-class CfgFunctions {
-    class CBA {
-        class Inventory {
-            PATHTO_FNC(switchableAttachments);
-        };
-    };
-};
--- a/addons/accessory/XEH_PREP.sqf
+++ b/addons/accessory/XEH_PREP.sqf
@@ -1,2 +1,5 @@
+PREPMAIN(switchableAttachments);
 
-PREP(switchAttachment);
+if (hasInterface) then {
+    PREP(switchAttachment);
+};
--- a/addons/accessory/XEH_preInit.sqf
+++ b/addons/accessory/XEH_preInit.sqf
@@ -1,39 +1,39 @@
 #include "script_component.hpp"
 
-if (!hasInterface) exitWith {};
-
 #include "XEH_PREP.sqf"
 
-[ELSTRING(common,WeaponsCategory), "MRT_SwitchItemNextClass_R", [LSTRING(railNext), LSTRING(railNext_tooltip)], {
-    [1, "next"] call FUNC(switchAttachment) // return
-}, {}, [DIK_L, [false, true, false]]] call CBA_fnc_addKeybind;
-
-[ELSTRING(common,WeaponsCategory), "MRT_SwitchItemPrevClass_R", [LSTRING(railPrev), LSTRING(railPrev_tooltip)], {
-    [1, "prev"] call FUNC(switchAttachment) // return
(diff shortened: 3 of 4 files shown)


commy2:
Upside will be that we can trim the header residue of both PREP and PREPMAIN after this is done and merged.

PabstMirror:
>CBA_fnc_addSettings is defined by CfgFunctions, cba_settings_fnc_init by PREP.
This means that cba_settings_fnc_init will be undefined when used by an addon in preInit that does not require cba_settings and is therefore likely loaded before the settings XEH preInit can define the function.

ref https://github.com/CBATeam/CBA_A3/pull/1111

commy2:
What Pabst said is a true concern and I don't know how to go about it (I will do the rest anyway for now).

PabstMirror:
but it's true for all cba funcs
e.g. addon using addPlayerEventHandler in preInit

commy2:
I know and I am thinking about it.

PabstMirror:
with current setup, I can start ace and then then disable compile cache as I need for each addon
with this PR once the game is started and the PREP_list is generated then it will always be there and compileFinaled, 

so I find this more restrictive
preInit with cba and ace takes less than a second, so I don't know if this kind of optimization is worth the drawbacks

commy2:
This is not done, pabst. The recompiling stuff is not implemented.

commy2:
@PabstMirror You may test again now.

PabstMirror:
```
11:31:36 Attempt to override final function - ace_attach_fnc_attach
11:31:36 Attempt to override final function - ace_attach_fnc_canattach
```
for just about every ace func, `DISABLE_COMPILE_CACHE` has no effect

commy2:
>DISABLE_COMPILE_CACHE has no effect

OK -.-

PabstMirror:
ace's prep is changed from cba
```
#ifdef DISABLE_COMPILE_CACHE
    #undef PREP
    #define PREP(fncName) DFUNC(fncName) = compile preprocessFileLineNumbers QPATHTOF(functions\DOUBLES(fnc,fncName).sqf)
#else
    #undef PREP
    #define PREP(fncName) [QPATHTOF(functions\DOUBLES(fnc,fncName).sqf), QFUNC(fncName)] call CBA_fnc_compileFunction
#endif
```

commy2:
`Attempt to override final function` is caused by CfgFunction init being executed multiple times due to some vanilla bug or whatever. I can work around that, but it should not affect functionality for now.

>ace's prep is changed from cba

Okay, that explains it. I will change it in ACE.

commy2:
>Attempt to override final function

spam should be fixed now.

Adds `_autoPrep` flag to CBA_fnc_compileFunction. Function is only added to PREP_List if this flag is set to true. PREPMAIN uses it, PREP does not, which means you can use DISABLED_COMPILE_CACHE in the component, but it also means you still need the PREP in preInit.

Any other issues with this?

commy2:
Test `CBA_fnc_addKeybind` without cba_main in requiredAddons.

veteran29:
> Test CBA_fnc_addKeybind without cba_main in requiredAddons.

Tested with following addons:
[required.zip](https://github.com/CBATeam/CBA_A3/files/5012510/required.zip)
[no_required.zip](https://github.com/CBATeam/CBA_A3/files/5012511/no_required.zip)

When `afm_pointing` does not have anything in required addons then the keybind is not added when `addKeybind `function is called in preInit. When `cba_main` is required then it works.

PR #1341, opened 2020-05-26
Add time getter to wait and waitUntil

Freddo3000:
This allows for other time commands to be checked, such as diag_frameNo, diag_tickTime, dateToNumber, etc.

The main benefit to this is for environments where CBA_missionTime does not tick, such as in editor or while the game is paused where a perFrameHandler would otherwise have to be used.

<details>
<summary>Provided Examples</summary>

```sqf
[{false},{},[],5,{
    systemChat "This will timeout in 5 frames"
},{diag_frameNo}] call CBA_fnc_waitUntilAndExecute
```

```sqf
[{player sideChat "5 frames later!"}, [], 5, {diag_frameNo}] call CBA_fnc_waitAndExecute;

// Note, will not work for new years
[{player sideChat "A month later!"}, [], 0.1, {dateToNumber date}] call CBA_fnc_waitAndExecute;
```

</details>

- Allow diag_frameNo, diag_tickTime, dateToNumber and custom implementations to be checked for CBA_fnc_waitAndExecute/waitUntilAndExecute.
- Remove sorting of waitAndExecuteArray, as it is incompatible with the above changes. This means all entries will be checked, not only the first one. A simple condition is used, so it isn't too bad.

### DIFF
--- a/addons/disposable/XEH_preInit.sqf
+++ b/addons/disposable/XEH_preInit.sqf
@@ -67,7 +67,7 @@ private _cfgMagazines = configFile >> "CfgMagazines";
 } forEach configProperties [configFile >> "CBA_DisposableLaunchers", "isArray _x"];
 
 ["CBA_settingsInitialized", {
-    ["All", "InitPost", FUNC(replaceMagazineCargo), nil, nil, true] call CBA_fnc_addClassEventHandler;
+    ["All", "InitPost", {call FUNC(replaceMagazineCargo)}, nil, nil, true] call CBA_fnc_addClassEventHandler;
 }] call CBA_fnc_addEventHandler;
 
 ADDON = true;


Freddo3000:
Marked as draft as for some reason, it does not execute it editor. Manually executing the queue however appears to function, I'll have to look into it.

commy2:
Slow and also deletes the sorting, completely destroying the point of this implementation.

Freddo3000:
Would it be better to put it in a separate queue, or should you have to do something similar to 
```sqf
myFunc = {systemChat "5 frames later"};

[{
	diag_frameNo > _this # 0
},{
systemChat "5 frames later";
(_this # 1) call myFunc;
},[diag_frameNo + 5, [myArgs]]] call CBA_fnc_waitUntilAndExecute;

// 5 frames later
```

The goal here was to make one able to do something similar to 
```sqf
myFunc = {systemChat "5 frames later"};

[myFunc,[myArgs],5,{diag_frameNo}] call CBA_fnc_waitAndExecute;
// 5 frames later
```

Freddo3000:
```sqf
params [["_function", {}, [{}]], ["_args", []], ["_delay", 0, [0]],["_getter","",[""]]];

if (_getter isEqualTo "") then {
    GVAR(waitAndExecArray) pushBack [CBA_missionTime + _delay, _function, _args];
    GVAR(waitAndExecArrayIsSorted) = false;
} else {
    [compile format ["%1 > %2", _getter, ([] call compile _getter) + _delay]] call CBA_fnc_waitUntilAndExecute;
};
```

Would this be acceptable, or should it be moved to a separate function?

commy2:
If the problem is that CBA_missionTime does not update/work in the editor, or the main menu, the lobby or wherever, then imo the proper fix would be to get CBA_missionTime to work there.

Freddo3000:
There's that though I'd like an implementation allowing it to be used with other script commands, such as 
```sqf
daytime
diag_tickTime
estimatedTimeLeft
getMusicPlayedTime
playerRespawnTime
serverTime
time
dateToNumber
```
and whatever custom implementations people might cook up.

commy2:
Why not add a pfh for these? Whenever needed I mean.

Freddo3000:
You mean something like this?

```sqf
if (_getter != "default") exitWith {
	GVAR(waitAndExecDayTimeArray) pushback [entry];
	if (isNil pfhDayTimeHandle) then {
		pfhDayTimeHandle = [{
			/*
			https://github.com/CBATeam/CBA_A3/blob/5c478a099b0954baf1c0b085861978e52847ce9b/addons/common/init_perFrameHandler.sqf#L43-L60
			*/
			if (GVAR(waitAndExecDayTimeArray) isEqualTo []) exitWith {
				[pfhDayTimeHandle] call CBA_fnc_removePerFrameHandler;
				pfhDayTimeHandle = nil:
			};
		}] call CBA_fnc_addPerFrameHandler;
	};
};
```

commy2:
Yeah. (Minus the removing part.)

veteran29:
Can't you use WUAE for these?

Freddo3000:
@veteran29 You *can*, but it isn't all that elegant
```sqf
myFunc = {systemChat "5 frames later"};

[{
	diag_frameNo > _this # 0
},{
systemChat "5 frames later";
(_this # 1) call myFunc;
},[diag_frameNo + 5, [myArgs]]] call CBA_fnc_waitUntilAndExecute;

// 5 frames later
```

vs

```sqf
myFunc = {systemChat "5 frames later"};

[myFunc,[myArgs],5,{diag_frameNo}] call CBA_fnc_waitAndExecute;
// 5 frames later
```

PabstMirror:
The `sort` is key to how sort and exec works; ace medical may add hundreds of entries, but only a single time comparison is needed each frame because it is sorted

You could add wrappers to waitUntilAndExec
```
CBA_fnc_waitFrames = { // public
    params ["_code", "_data", "_frames"];
    [CBA_fnc_waitFrames_condition, CBA_fnc_waitFrames_statement, [diag_frameNo+_frames, _code, _data]] call CBA_fnc_waitUntilAndExecute
};
CBA_fnc_waitFrames_condition = { // internal
    diag_frameNo > (_this # 0) # 0
};
CBA_fnc_waitFrames_statement = { // internal
    (_this # 2) call (_this # 1)
};
```

Freddo3000:
Unable to reopen this pull request, new one here https://github.com/CBATeam/CBA_A3/pull/1342

PR #1344, opened 2020-05-27
Unlock malaria-infected civs

Kexanone:
#### Preface:
The Old Man update introduced malaria-infected civilians, however these classes are currently not visible in Eden and Zeus. The way they work is that they change the wound textures of the base civilian classes, which means that the infection is only visible when the unit is injured.

#### When merged this pull request will:
- Unlock malaria-infected civilian classes.
- Set the proper identities for malaria-infected civilian (fixes the faces).
- Add damage to malaria-infected civilian to make the infection visible.


addons/characters/CfgVehicles.hpp
```diff
@@ -0,0 +1,69 @@
...
+            'BIS_Lucas_sick', \
+            'BIS_Renly_sick' \
+        ]; \
+        (_this select 0) setDamage 0.45"; \
+    }
+
+class CfgVehicles {
+    class C_Man_casual_1_F_afro;
+    class C_Man_casual_1_F_afro_sick: C_Man_casual_1_F_afro {
+       UNLOCK_MALARIA_INFECTED_CIVILIAN;
```
jonpas, line 22: 3-space indentation? :D

addons/characters/CfgVehicles.hpp
```diff
@@ -0,0 +1,69 @@
...
+            'BIS_Ambient01_sick', \
+            'BIS_Ambient02_sick', \
+            'BIS_Ambient03_sick', \
+            'BIS_Arthur_sick', \
+            'BIS_Howard_sick', \
+            'BIS_John_sick', \
+            'BIS_Lucas_sick', \
+            'BIS_Renly_sick' \
+        ]; \
+        (_this select 0) setDamage 0.45"; \
```
PabstMirror, line 16: `setDammage` has global effects, every time someone connects to server this will run

PabstMirror:
Is there anyway to make these non-playable? (can not be slotted by a player)

addons/characters/CfgVehicles.hpp
commy2: It also overwrites XEH for these classes, somewhat ironically.

addons/characters/stringtable.xml
```diff
@@ -0,0 +1,15 @@
...
+<Project name="CBA_A3">
+    <Package name="Characters">
+        <Key ID="STR_CBA_Characters_Component">
+            <English>Community Base Addons - Characters</English>
+            <German>Community Base Addons - Charaktere</German>
+            <French>Community Base Addons - Caractères</French>
+        </Key>
+        <Key ID="STR_CBA_Characters_MenMalariaInfected">
+            <English>Men (Malaria-Infected)</English>
+            <German>Männer (Malariainfiziert)</German>
```
commy2, line 11: I could find "Malariainfektion" as compound noun, but not this. Pretty sure it should say "Malaria infiziert" as short hand for "mit Malaria infiziert".

commy2:
>Is there anyway to make these non-playable? (can not be slotted by a player)

I don't think so. Well, you could always script the editor the remove the attribute when the selection contains at least one of these, but that would be rather clumsy.

addons/characters/CfgVehicles.hpp
commy2: Also, while `setIdentity` is a local effects command, `selectRandom` will lead to every client picking a different, not synchronized identity :D

addons/characters/CfgVehicles.hpp
commy2: Actually 7.

addons/characters/CfgVehicles.hpp
```diff
@@ -0,0 +1,69 @@
...
+
+class CfgVehicles {
+    class C_Man_casual_1_F_afro;
+    class C_Man_casual_1_F_afro_sick: C_Man_casual_1_F_afro {
+       UNLOCK_MALARIA_INFECTED_CIVILIAN;
+    };
+
+    class C_Man_casual_3_F_afro;
+    class C_Man_casual_3_F_afro_sick: C_Man_casual_3_F_afro {
+       UNLOCK_MALARIA_INFECTED_CIVILIAN;
```
commy2, line 27: The class is missing an `author` entry. Keep in mind that `author` can not be inherited (is ignored when inherited), so these classes will currently display as "unknown community author" (or something similar), which is funny, because it's super generic, yet still completely wrong.

I'd add to the UNLOCK_xxx macro:
```cpp
        author = "$STR_A3_Bohemia_Interactive";
```

addons/characters/CfgVehicles.hpp
commy2: Just me messing around, but a global init script could work if seeds are used for randomization, e.g. the second part of the object net id:
```sqf
params ['_unit'];

if (isServer) then {
    _unit setDamage 0.45;
};

private _identities = ['BIS_Ambient01_sick', 'BIS_Ambient02_sick', 'BIS_Ambient03_sick', 'BIS_Arthur_sick', 'BIS_Howard_sick', 'BIS_John_sick', 'BIS_Lucas_sick', 'BIS_Renly_sick'];
private _identity = _identities select floor (parseNumber (_unit call BIS_fnc_netId splitString ":" param [1, '0']) random count _identities);
_unit setIdentity _identity;
```

I do not know if `setIdentity` even works in init. Yesterday I helped a poor soul on reddit with setSpeaker and it wouldn't work even in InitPost until a 5 second delay was added.

addons/characters/CfgVehicles.hpp
oOKexOo: lol

addons/characters/CfgVehicles.hpp
oOKexOo: I'll give another thought about selecting the identity.
`setIdentity` most certainly work in the init. Miller's identity is set that way.

commy2: It's a campaign unit though. Tried Miller in MP?

addons/characters/CfgVehicles.hpp
```diff
@@ -1,19 +1,26 @@
...
+                'BIS_Ambient03_sick', \
+                'BIS_Arthur_sick', \
+                'BIS_Howard_sick', \
+                'BIS_John_sick', \
+                'BIS_Lucas_sick', \
+                'BIS_Renly_sick' \
+            )]; \
+            [ARR_2( \
+                QGVAR(broadcastIdentity), \
+                [ARR_2(_unit, _identity)] \
```
commy2, line 20: I thought the identity is copied from the server on JIP by default?

commy2:
```cpp
class C_man_p_beggar_F_afro_sick: C_man_p_beggar_F_afro {};
class C_Man_casual_1_F_afro_sick: C_Man_casual_1_F_afro {};
class C_Man_casual_3_F_afro_sick: C_Man_casual_3_F_afro {};
class C_Man_casual_4_F_afro_sick: C_Man_casual_4_F_afro {};
class C_Man_casual_5_F_afro_sick: C_Man_casual_5_F_afro {};
class C_Man_casual_6_F_afro_sick: C_Man_casual_6_F_afro {};
class C_man_polo_1_F_afro_sick: C_man_polo_1_F_afro {};
class C_man_polo_2_F_afro_sick: C_man_polo_2_F_afro {};
class C_man_polo_3_F_afro_sick: C_man_polo_3_F_afro {};
class C_man_polo_6_F_afro_sick: C_man_polo_6_F_afro {};
class C_man_sport_2_F_afro_sick: C_man_sport_2_F_afro {};
```

addons/characters/CfgVehicles.hpp
oOKexOo: Doesn't seem to be the case. The face got lost for JIPs when I used `cba_fnc_globalEvent`.

addons/characters/CfgVehicles.hpp
oOKexOo: Funny enough Miller works fine in MP even though `setIdentity` in the init line is not postponed by a frame. Dunno how that works...

commy2:
This is how I imagined it.

commy2:
`C_man_p_beggar_F_afro_sick` has a LOD issue. In first person, the sick material is used, but not in third person (this includes looking at other entities regardless of camera mode).

<details><summary>pics</summary>




</details>

Seems to be a non-matching material in both lods compared to what is listed in class Damage. Not going to fix this BI bug though. Unit is excluded.

commy2:
And this is how you do it without the scripts (apart from the setDamage part).

Game apparently has faces for different races:
```
QGVAR(Head_NATO_sick)
QGVAR(Head_African_sick)
QGVAR(Head_Asian_sick)
QGVAR(Head_Greek_sick)
QGVAR(Head_TK_sick)
QGVAR(Head_Tanoan_sick)
QGVAR(Head_Euro_sick)
```
But all classes are African. What do about that?

What's the rationale behind 0.475 vs 0.45? 0.45 seems to be enought to trigger head damage material.

addons/characters/CfgVehicles.hpp
commy2: obsolete

oOKexOo:
> C_man_p_beggar_F_afro_sick has a LOD issue.

Yeah, I noticed that something was off with that class, that's why I excluded it.

> And this is how you do it without the scripts (apart from the setDamage part).

Agreed, much nicer solution when one knows how to configure characters properly.

> Game apparently has faces for different races:

Yeah, I noticed that, but I'm not sure they are functional.

> What's the rationale behind 0.475 vs 0.45? 0.45 seems to be enough to trigger head damage material.

I thought so too, but when I tested it on dedicated yesterday, the results were inconsistent with `0.45`. With `0.50` they start limping.

commy2:
>Yeah, I noticed that, but I'm not sure they are functional.

I will try. If they are, they could still be used with setFace, which should then also change the skin color of the body.

>I thought so too, but when I tested it on dedicated yesterday, the results were inconsistent with 0.45

May have been a rounding thing? Can't really imagine that though.

commy2:
The other faces all work in that they don't show bloddy scratches as a wounded person would. Only some of them show the skin disease though. I'd say we make the existing classes public, and give the other faces identity tokens and that's it. So this PR is done.

oOKexOo:
Works nicely on dedicated.

oOKexOo:
> May have been a rounding thing? Can't really imagine that though.

My earlier comment was wrong. It is actually consistent. `0.45` doesn't work for JIPs, but `0.46` seems to be fine.

jonpas (approved):

PR #1346, opened 2020-05-30
Invisible Target improvements

### LINKED ISSUE #1331: Using invisible targets on dedicated server results in server crash
**Mods (complete and add to the following information):** Custom repo, but can be repro'd with just cba
- **Arma 3:** `x.xx` [e.g. 1.00 stable, rc,  dev] Stable 
- **CBA:** `3.x.x` [e.g. 3.0.0 stable, commit hash] 3.15.1 stable from workshop
_Make sure to reproduce the issue with only CBA on a newly created mission!_

**Description:**
Running a mission containing a cba invisible target crashes the server, requiring it to reboot its client, and returns the following error in its log:

18:11:34 Error: Neither Geometry nor LandContact level is present in model x\cba\addons\ai\invisibletarget.p3d

Although I haven't checked the logs to see if this error happens on singleplayer, I know it runs fine without crashing.

**Steps to reproduce:**
- Run a mission on a dedicated server that contains a pre-placed invisible target object

**Expected behavior:**
N/A

- Dedicated server only

**Log Files:**
- https://pastebin.com/G6D6NK7a

**Additional context:**
N/A

**Screenshots:**
N/A

### LINKED ISSUE #1340: Invisible vehicle targets
Making AI use grenades, rockets, tank cannons or other explosives against any object that is not considered a vehicle can be rather frustrating.
Since CBA already has the invisible infantry target solider, it would be useful to add an invisible vehicle as well.

An invisible vehicle target would be an easy solution to this problem, that everyone could easily use.

How i usually do this:
https://community.bistudio.com/wiki/fireAtTarget
https://community.bistudio.com/wiki/forceWeaponFire
However using these requires more scripting than a simple task like this should. Also can be quite performance intensive in busy missions or when trying to create sustained fire.
And the "destroy waypoint" seems to be rather useless and inconsistent.

commy2:
- Adds LandContact and Geometry LODs to InvisibleTarget.p3d, which _should_
- fix #1331
- Adds InvisibleTargetVehicle.p3d, which is the same as the other model, except with FireGeometry, so the object can be used as
- Vehicle Target & Air Target classes for all 3 sides, which
- closes #1340
- The InvisibleTargetVehicle.p3d P3D uses a cloth material, so the hit effect is minimal (almost unnoticable)

Please test if the crash is resolved.


PabstMirror:
before  dedi would hardlock for ~2 minutes, with client just showing "Wait for host" - see timecodes on server rpt
```
11:37:58 [CBA] (xeh) INFO: [4513,54.209,0] PreInit finished.
11:39:50 [CBA] (xeh) INFO: [4520,166.826,0] PostInit started. MISSIONINIT: missionName=testCBA-invis, missionVersion=53, worldName=Stratis, isMultiplayer=true, isServer=true, isDedicated=true, CBA_isHeadlessClient=false, hasInterface=false, didJIP=false
```
after: - almost no delay
```
11:41:09 [CBA] (xeh) INFO: [2796,31.631,0] PreInit finished.
11:41:09 [CBA] (xeh) INFO: [2803,32.227,0] PostInit started. MISSIONINIT: missionName=testCBA-invis, missionVersion=53, worldName=Stratis, isMultiplayer=true, isServer=true, isDedicated=true, CBA_isHeadlessClient=false, hasInterface=false, didJIP=false


```

PabstMirror (approved):

commy2:
So #1331 is not a crash, but a temporary freeze?

PR #1350, opened 2020-07-05
add CBA_fnc_hashValues

Fusselwurm:
There is a `hashKeys` function  - but the corresponding function `hashValues` is missing

### DIFF
--- a/addons/hashes/fnc_hashValues.sqf
+++ b/addons/hashes/fnc_hashValues.sqf
@@ -1,0 +1,25 @@
+#include "script_component.hpp"
+#include "script_hashes.hpp"
+/* ----------------------------------------------------------------------------
+Function: CBA_fnc_hashValues
+
+Description:
+    Returns all values in a Hash.
+
+    See <CBA_fnc_hashCreate>.
+
+Parameters:
+    _hash - Hash to get values for [Array which is a Hash structure]
+
+Returns:
+    Array of all values [Array]
+
+Author:
+    Fusselwurm
+---------------------------------------------------------------------------- */
+SCRIPT(hashValues);
+
+// -----------------------------------------------------------------------------
+params [["_hash", [[], []], [[]]]];
+
++(_hash select HASH_VALUES)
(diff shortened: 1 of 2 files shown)


addons/hashes/fnc_hashValues.sqf
```diff
@@ -0,0 +1,25 @@
...
+
+Author:
+    Fusselwurm
+---------------------------------------------------------------------------- */
+SCRIPT(hashValues);
+
+// -----------------------------------------------------------------------------
+params [["_hash", [[], []], [[]]]];
+
++(_hash select HASH_VALUES)
```
PabstMirror, line 25:
```suggestion
_hash select HASH_VALUES
```
`select` always creates a copy

PabstMirror (approved):

addons/hashes/fnc_hashValues.sqf
mharis001: `ARRAY select NUMBER` returns a reference. For example:
```sqf
private _a = [[1], []]; 
private _b = _a select 0; 
_b pushBack 2; 
_a // -> [[1,2],[]]
```

Fusselwurm: ok, I removed the changes again :sweat_smile:  `+(x select y)` it is!

commy2: Consider a shallow copy:

``` 

[] + (_hash select HASH_VALUES

```
which will keep value array references as is what I would expect.

PabstMirror: oh no I'm an idiot; I was thinking of `select {}`

commy2, line 25:
```sqf
[] + (_hash select HASH_VALUES)
```

as described.

commy2:
Now it does a deep copy.


```sqf
private _names = [] call CBA_fnc_hashCreate; 
[_names, "super", [1,2,3]] call CBA_fnc_hashSet; 
 
{ 
    _x pushBack 4; 
} forEach (_hash call CBA_fnc_hashValues); 
 
[_names, "super"] call CBA_fnc_hashGet // got [1,2,3], expected [1,2,3,4]
```

while:
```sqf
private _names = [] call CBA_fnc_hashCreate; 
[_names, "super", [1,2,3]] call CBA_fnc_hashSet; 
 
([_names, "super"] call CBA_fnc_hashGet) pushBack 4; 
 
[_names, "super"] call CBA_fnc_hashGet // got [1,2,3]
```

PR #1354, opened 2020-07-08
Object cache for CBA_fnc_getObjectConfig

kuloodporny:
- Make CBA_fnc_getObjectConfig cache config class pointer as object variable "CBA_config"

### DIFF
--- a/addons/common/fnc_getObjectConfig.sqf
+++ b/addons/common/fnc_getObjectConfig.sqf
@@ -42,7 +42,7 @@ if _isObject then {
 };
 
 // Invalid class
-if (count _className == 0) exitWith {_result};
+if (_className isEqualTo "") exitWith {_result};
 
 {
     private _config = configFile >> _x >> _className;
--- a/addons/common/fnc_getObjectConfig.sqf
+++ b/addons/common/fnc_getObjectConfig.sqf
@@ -10,7 +10,7 @@ Parameters:
     _object - Any kind of object or class name <STRING, OBJECT>
 
 Returns:
-    _config - Objects config. <CONFIG>
+    _result - Objects config. <CONFIG>
 
 Example:
     (begin example)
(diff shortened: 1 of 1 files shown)


commy2:
Benchmarks please.

dedmen:
I created a internal ticket for https://sqf.bisimulations.com/display/SQF/configOf which can probably just replace this then

kuloodporny:
CBA v3.15.1.200418
```Result:
0.0086 ms

Cycles:
10000/10000

Code:
[player] call cba_fnc_getObjectConfig```
CBA with Artifact from this task:
```Result:
0.0062 ms

Cycles:
10000/10000

Code:
[player] call cba_fnc_getObjectConfig```

**Comment by commy2** [comment]
Now do
```sqf
typeOf player call cba_fnc_getObjectConfig
```

addons/common/fnc_getObjectConfig.sqf
```diff
@@ -19,24 +20,39 @@ Example:
...
+if (isClass _result) exitWith {_result};
+
+// Convert object to classname to be found if needed
+private _className = _object;
+if _isObject then {
+    _className = typeOf _object;
+};
+
+// Invalid class
+if (count _className == 0) exitWith {_result};
```
PabstMirror, line 45:
```suggestion
if (_className == "") exitWith {_result};
```

kuloodporny: I used ` if (_className isEqualTo "") exitWith {_result};` as it seems to have tiny performance advantage even in this case.
```
Result:
3.69373 ms

Cycles:
271/10000

Code:
for "_i" from 0 to 10000 do {"" isEqualTo ""};
```
vs
```
Result:
3.7594 ms

Cycles:
266/10000

Code:
for "_i" from 0 to 10000 do {"" == ""};
```
I did about 20 measurements and these are the best ones I obtained, with `==` consistently slower than `isEqualTo`.

kuloodporny:
In case of string parameter, the changed function is significantly **slower**:
Steam CBA_A3:
```
Result:
0.0074 ms

Cycles:
10000/10000

Code:
typeOf player call cba_fnc_getObjectConfig
```
This branch :
```
Result:
0.0103 ms

Cycles:
10000/10000

Code:
typeOf player call cba_fnc_getObjectConfig
```

kuloodporny:
I have rewritten the solution a bit to make the script simpler, the results in console tests are promising - below 80 ms for text and below 50 ms for object, but I need to check it with new artifact.

kuloodporny:
and indeed, the code with less changes is not that much worse:
```
Result:
0.0079 ms

Cycles:
10000/10000

Code:
typeOf player call cba_fnc_getObjectConfig
```
and object parameter is significantly faster:
```
Result:
0.005 ms

Cycles:
10000/10000

Code:
player call cba_fnc_getObjectConfig
```
I did over 30 tests each, the fastest ones are presented and the average results are not significantly worse.

commy2:
Now let's benchmark different simulation types:
```sqf
mine1 = createMine ["APERSMine", ASLToAGL getPosASL player, [], 0];
```
once, then code performance:
```
mine1 call cba_fnc_getObjectConfig
```

kuloodporny:
**Getting config of `mine1` is significantly slower with the caching enabled - over 50% overhead is present.**
Caching on:
```
Result:
0.0157 ms

Cycles:
10000/10000

Code:
mine1 call cba_fnc_getObjectConfig
```
Caching off:
```
Result:
0.01 ms

Cycles:
10000/10000

Code:
mine1 call cba_fnc_getObjectConfig
```

**Getting config for `typeOf mine1` is pretty much the same.**
Caching on:
```
Result:
0.0098 ms

Cycles:
10000/10000

Code:
typeOf mine1 call cba_fnc_getObjectConfig```
Caching off:
```
Result:
0.0097 ms

Cycles:
10000/10000

Code:
typeOf mine1 call cba_fnc_getObjectConfig```

commy2:
It seems that the change gives an advantage for some object classes, but disfavours classnames and a few object classes. It is not clear to me that it is an all around improvement to have this cached.

kuloodporny:
Agree, I'd say let's scratch it and wait for configOf to be implemented.

PR #1359, opened 2020-07-29
state machines: configurable number of ticks per frame

Fusselwurm:
CBA state machines are very polite beasts: they each run a single tick on a single element each frame. 
This is nice for performance considerations, but it makes them slooow to react if you have a lot of entities.

This PR creates an optional additional parameter for `CBA_statemachines_fnc_create` that allows users to specify how many ticks the given state machine should run on each frame, defaulting to 1.

To not add yet another indentation level to the clockwork function, I extract the tick logic into `CBA_statemachine_fnc_tick`.


addons/statemachine/fnc_createFromConfig.sqf
```diff
@@ -28,7 +28,9 @@ if (isNull _config) exitWith {};
 
 private _list = compile getText (_config >> "list");
 private _skipNull = (getNumber (_config >> "skipNull")) > 0;
-private _stateMachine = [_list, _skipNull] call FUNC(create);
+private _ticksPerFrame = getNumber (_config >> "ticksPerFrame");
+_ticksPerFrame = if (_ticksPerFrame == 0) then {1} else {_ticksPerFrame};
```
jonpas, line 32: select?

addons/statemachine/fnc_createFromConfig.sqf
```diff
@@ -28,7 +28,9 @@ if (isNull _config) exitWith {};
 
 private _list = compile getText (_config >> "list");
 private _skipNull = (getNumber (_config >> "skipNull")) > 0;
-private _stateMachine = [_list, _skipNull] call FUNC(create);
+private _ticksPerFrame = getNumber (_config >> "ticksPerFrame");
+_ticksPerFrame = [_ticksPerFrame, 1] select (_ticksPerFrame == 0);
```
mharis001, line 32: Maybe use `max 1` here.

PiZZAD0X:
https://github.com/PiZZAD0X/Olsen-Framework-Arma-3/blob/732eae4efc878f75e40ddaacddd1cd15d9c39511/modules/headless_ai/cfgFunctions/statemachine/fn_clockwork.sqf#L105

Yup, been using a statemachine with tick system for a while, works well.

jonpas:
@BaerMitUmlaut

PR #1364, opened 2020-08-22
Add parameter to force target debug console from server's config

TheShinriel:
- Add category "CBA Tools" for addon's options
- Add addon's options "forceTargetDebug"
  - Will start the target debug console even if there is the mission's parameter is not selected, allowing to force this tool from server's configuration
  - The console will appear directly, but mission must be restarted to take into account the variables
- Fix CBA_fn_addSetting parameter _isGlobal, based on the param in the script

### DIFF
--- a/addons/diagnostic/XEH_preInit.sqf
+++ b/addons/diagnostic/XEH_preInit.sqf
@@ -25,60 +25,62 @@ ADDON = true;
     GVAR(settingsInitialized) = true;
 }] call CBA_fnc_addEventHandler;
 
-waitUntil{!isNil QGVAR(settingsInitialized)};
-if (
-    getMissionConfigValue ["EnableTargetDebug", 0] == 1
-    || {getNumber (configFile >> "EnableTargetDebug") == 1}
-    || {GVAR(enableTargetDebug)}
-) then {
-    INFO("EnableTargetDebug is enabled.");
+0 spawn {
+    waitUntil{!(isNil QGVAR(settingsInitialized))};
+    if (
+        getMissionConfigValue ["EnableTargetDebug", 0] == 1
+        || {getNumber (configFile >> "EnableTargetDebug") == 1}
+        || {GVAR(enableTargetDebug)}
+    ) then {
+        INFO("EnableTargetDebug is enabled.");
 
-    [QGVAR(watchVariable), {
-        params ["_clientID", "_varIndex", "_statementText"];
-        TRACE_3("targetWatchVariable",_clientID,_varIndex,_statementText);
+        [QGVAR(watchVariable), {
+            params ["_clientID", "_varIndex", "_statementText"];
+            TRACE_3("targetWatchVariable",_clientID,_varIndex,_statementText);
 
-        private _timeStart = diag_tickTime;
-        private _returnString = _statementText call {
-            private ["_clientID", "_statementText", "_varName", "_timeStart", "_x"]; // prevent these variables from being overwritten
-            _this = ([nil] apply compile _this) select 0;
-            if (isNil "_this") exitWith {"<any>"};
-            str _this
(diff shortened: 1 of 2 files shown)


addons/diagnostic/stringtable.xml
```diff
@@ -92,5 +92,9 @@
             <Italian>[CBA] Abilita il debug remoto. Richiede la console di debug.</Italian>
             <Czech>[CBA] Povoluje ladění vzdáleného cíle. Vyžaduje ladící konzoli.</Czech>
         </Key>
+        <Key ID="STR_cba_diagnostic_ForceTargetDebug">
+            <English>Force target debugging for all missions</English>
+            <French>Forcer le débogage de la cible pour toutes les missions</French>
```
commy2, line 97: Periods after sentences.

commy2:
- Why "force" though? Why not EnableTargetDebug like the other switches? They are disjunct with none having any priority whatsoever.

- Category could be Debug or probably better yet Diagnositic, to correspond to the component (less invented terminology).

addons/diagnostic/XEH_preInit.sqf
```diff
@@ -18,7 +20,7 @@ GVAR(projectileTrackedUnits) = [];
 
 ADDON = true;
 
-if (getMissionConfigValue ["EnableTargetDebug", 0] == 1 || {getNumber (configFile >> "EnableTargetDebug") == 1}) then {
+if (getMissionConfigValue ["EnableTargetDebug", 0] == 1 || {getNumber (configFile >> "EnableTargetDebug") == 1} || {GVAR(forceTargetDebug)}) then {
```
PabstMirror, line 23: this happens before settings are initialized

addons/diagnostic/stringtable.xml
TheShinriel: I guess you mean dots? Fix soon

TheShinriel:
> * Why "force" though? Why not EnableTargetDebug like the other switches? They are disjunct with none having any priority whatsoever.
> * Category could be Debug or probably better yet Diagnositic, to correspond to the component (less invented terminology).

* Force, was because of the existance 
  * "force" because of the existing, that option will ignore the configuration of the mission. But indeed, by looking the option, could have sens alone.
* I will rename for "Diagnositic"

addons/diagnostic/XEH_preInit.sqf
TheShinriel: `#include "initSettings.sqf"`
Is called before, to init the settings.
I have tested solo and multi with and without the option on mission without the mission's config selected, that works great

addons/diagnostic/stringtable.xml
commy2: Period as in full stop
`.`

addons/diagnostic/XEH_preInit.sqf
commy2: The global variable is definitely not synchronized from the server at preInit. This is a global setting. Try on a server as dedicated client.

TheShinriel: I will test more with a dedicated client and a dedicated server
I could add a waitUntil in the condition, not executed if mission is set to true, however, it will wait the synchronizatio... still have to check what variable I can watch to know when it's done

Thx for the explanation.

addons/diagnostic/stringtable.xml
TheShinriel: Reviewed, to use the same text as the option in Eden

addons/diagnostic/XEH_preInit.sqf
TheShinriel: Oki added a commit, that wait the event CBA_settingsInitialized event

PR #1376, opened 2020-10-03
Utilize isNotEqualTo

Neviothr:
- title

### DIFF
--- a/addons/settings/fnc_export.sqf
+++ b/addons/settings/fnc_export.sqf
@@ -43,7 +43,7 @@ private _temp = [];
     };
 
     if (_exportDefault || {
-        (_value isNotEqualTo _defaultValue) ||
+        _value isNotEqualTo _defaultValue ||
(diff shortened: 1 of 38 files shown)


PabstMirror:
looks good, but I don't think `isNotEqualTo` will make it into 2.00, so it may be a while until this can be merged

addons/settings/fnc_export.sqf
```diff
@@ -43,7 +43,7 @@ private _temp = [];
     };
 
     if (_exportDefault || {
-        !(_value isEqualTo _defaultValue) ||
+        (_value isNotEqualTo _defaultValue) ||
```
commy2, line 46: You can get rid of these parentheses

commy2 (approved):

commy2:
do not merge before isNotEqualTo is in

PabstMirror (approved):
lgtm

PR #1377, opened 2020-10-13
hashValues - don't copy arrays

PabstMirror:
Co-Authored-By: commy2 <6576312+commy2@users.noreply.github.com>

Without fix it will fail this test case
```
_data pushBack [7];
TEST_OP(_values,isEqualTo,[[ARR_2(3,[7])]],"hashValues - deep array copy");
```

```
[CBA] (hashes) ERROR: Test FAIL x\cba\addons\hashes\test_hashes.sqf:154
             (_values isEqualTo [[3, [7]]])
```

### DIFF
--- a/addons/hashes/fnc_hashValues.sqf
+++ b/addons/hashes/fnc_hashValues.sqf
@@ -22,4 +22,4 @@ SCRIPT(hashValues);
 // -----------------------------------------------------------------------------
 params [["_hash", [[], []], [[]]]];
 
-[]+(_hash select HASH_VALUES)
+ [] + (_hash select HASH_VALUES) // flat-copy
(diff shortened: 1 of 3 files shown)


addons/hashes/fnc_hashValues.sqf
```diff
@@ -22,4 +22,4 @@ SCRIPT(hashValues);
 // -----------------------------------------------------------------------------
 params [["_hash", [[], []], [[]]]];
 
-+(_hash select HASH_VALUES)
+[]+(_hash select HASH_VALUES)
```
commy2, line 25:
```suggestion
 [] + (_hash select HASH_VALUES) // flat-copy
```
Spaces around command, and doc for obscure syntax.

commy2 (approved):

jokoho48:
hashKeys should also changed to have the same behaviour

PR #1379, opened 2020-10-14
Fixed CBA_fnc_addKeybind and custom controls

### LINKED ISSUE #1378: Cannot register keybind with a default custom control
**Mods**
- **Arma 3:** `1.98 146373`
- **CBA:** `3.1.5`
**Description:**

Using `CBA_fnc_addKeybind` you can map custom user actions 1-5 on to a control, but if you try 6-20 it will fail due to it being incorrectly verified as out of bounds for DIK codes (not taking into account the extra ones added for the user actions for joystick compatibility.)

The issue is with [this](https://github.com/CBATeam/CBA_A3/blob/79c04b62f67c926962fe079c49258e646fac0365/addons/keybinding/fnc_addKeybind.sqf#L106) line

It may also be worth documenting those custom controls somewhere too in existing documentation. 

**Steps to reproduce:**
```sqf
["testMod"
    , "workingBind"
    , "This one is mapped"
    , {hint "works!"}
    , {}
    , [0xFE, [false, false, false]] //Custom user action 5
] call CBA_fnc_addKeybind;

["testMod"
    , "brokenBind"
    , "This one is not mapped"
    , {hint "Doesn't work"}
    , {}, [0xFF, [false, false, false]] //Custom user action 6
] call CBA_fnc_addKeybind;
```

**Expected behaviour:**
Keybind `"This one is mapped"` and `"This one is not mapped"` should be mapped to custom user actions 5 and 6 respectively.

- Any location where you try to map the keybind

**Log Files:**
[https://gist.github.com/mattysmith22/a49f9060c2334f7214c8d9575bc16dbf](https://gist.github.com/mattysmith22/a49f9060c2334f7214c8d9575bc16dbf)
Specific focus on the following line:
```
12:05:10 [CBA] (keybinding) WARNING: Keybind testmod$brokenbind's default keybind is invalid [DIK: 255]
```
**Additional context:**
This was found while migrating a mod (AH-64D apache) from old custom controls to new CBA keybinds due to some people having issues with conflicts. This was intended to provide seamless backwards compatibility.

mattysmith22:
*When merged this pull request will:**
- Fix #1378

### DIFF
--- a/addons/keybinding/fnc_addKeybind.sqf
+++ b/addons/keybinding/fnc_addKeybind.sqf
@@ -21,6 +21,27 @@ Description:
         0xF8: Mouse wheel up
         0xF9: Mouse wheel down
 
+        0xFA: Custom user action 1
+        0xFB: Custom user action 2
+        0xFC: Custom user action 3
+        0xFD: Custom user action 4
+        0xFE: Custom user action 5
+        0xFF: Custom user action 6
+        0x100: Custom user action 7
+        0x101: Custom user action 8
+        0x102: Custom user action 9
+        0x103: Custom user action 10
+        0x104: Custom user action 11
+        0x105: Custom user action 12
+        0x106: Custom user action 13
+        0x107: Custom user action 14
+        0x108: Custom user action 15
+        0x109: Custom user action 16
+        0x10A: Custom user action 17
+        0x10B: Custom user action 18
+        0x10C: Custom user action 19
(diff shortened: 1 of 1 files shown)


mattysmith22:
I had a look around and that is the only place I can see this limitation. It still may be worth getting someone who is more knowledgable on the addon to double check

commy2 (approved):

PR #1385, opened 2020-10-25
XEH - Add support for backpack containers

commy2:
- add XEH support for backpack containers <details><summary>spoiler</summary>"title"</details>

```cpp
class Extended_InitPost_EventHandlers {
    class Bag_Base {
        Mission_Test = "systemChat str _this; diag_log _this";
    };
};
```

Note however, that the retroactive flag does not seem to pick up already existing backpacks:
```sqf
["Bag_Base", "InitPost", {
    systemChat str _this;
}, nil, nil, true] call CBA_fnc_addClassEventHandler;
```

### DIFF
--- a/addons/xeh/CfgVehicles.hpp
+++ b/addons/xeh/CfgVehicles.hpp
@@ -233,6 +233,9 @@ class CfgVehicles {
     class WeaponHolder: ReammoBox {
         XEH_ENABLED;
     };
+    class Bag_Base: ReammoBox {
+        XEH_ENABLED;
+    };
 
     class ThingX;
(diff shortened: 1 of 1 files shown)


commy2:
For the problem above, I don't see a simple work around.

```sqf
allMissionObjects "" apply {everyBackpack _x}
```
would be very slow.

PabstMirror:
Doesn't fire for backpacks added to cargo: `car addBackpackCargo ["B_AssaultPack_mcamo_Ammo", 1]`
or backpacks from disassembled static weapons

PR #1387, opened 2020-11-03
Translation - Add French Translations (Ai)

Elgin675:
- Add french translations in `ai` module.

### DIFF
--- a/addons/ai/stringtable.xml
+++ b/addons/ai/stringtable.xml
@@ -28,10 +28,12 @@
         <Key ID="STR_CBA_AI_InvisibleTargetVehicle">
             <English>Invisible Target Vehicle</English>
             <German>Unsichtbares Ziel - Fahrzeug</German>
+            <French>Véhicule cible invisible</French>
         </Key>
         <Key ID="STR_CBA_AI_InvisibleTargetAir">
             <English>Invisible Target Airplane</English>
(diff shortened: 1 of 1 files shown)


commy2:
By using these names, the objects don't appear grouped in the editor I believe.

Elgin675:
@commy2 Indeed.
But I had no choice, to maintain consistency with the third invisible target, whose name I cannot change because it is inherited from vanilla.

If I could, I would have done this :
- Cible invisible - Soldat
- Cible invisible - Véhicule
- Cible invisible - Avion

But `CBA_B_InvisibleTarget` is inherited from vanilla (_"Soldat cible invisible"_), so I have no choice.
However, if we could rename `CBA_B_InvisibleTarget` via stringtable, I would be more than happy to change that.

commy2:
Ah, I see. Yeah, that sucks. I would've went with the systematic name anyway, but your call.

Maybe we could just copy paste the token and edit the French entry? Not sure if it's worth that for one language, though maybe there is a point if multiple languages have that oversight.

Elgin675:
> 
> 
> Maybe we could just copy paste the token and edit the French entry? Not sure if it's worth that for one language, though maybe there is a point if multiple languages have that oversight.

@commy2 If we could do that, that would be great :) !
I think it could also be useful for some other languages.

PabstMirror (approved):

PR #1393, opened 2020-11-20
AI - Put Invisible Target Soldier String Key Into CBA

commy2:
- The translation already is in the base game and used on these objects by inheritance.
- Therefore all these translations are presumed to be correct.
- This allows translators to change the names to fix the lexicographic order of this object with the other two targets (e.g. French)

### DIFF
--- a/addons/ai/stringtable.xml
+++ b/addons/ai/stringtable.xml
@@ -29,7 +29,7 @@
             <English>Invisible Target Soldier</English>
             <German>Unsichtbares Ziel - Soldat</German>
             <Chinese>隱形目標士兵</Chinese>
-            <French>Soldat cible invisible</French>
+            <French>Cible invisible - Soldat</French>
             <Spanish>Soldado objetivo invisible</Spanish>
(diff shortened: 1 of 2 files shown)


addons/ai/stringtable.xml
```diff
@@ -25,6 +25,22 @@
             <Russian>ИИ Место строительства</Russian>
             <Polish>Pozycja AI w budynku</Polish>
         </Key>
+        <Key ID="STR_CBA_AI_InvisibleTargetSoldier">
+            <English>Invisible Target Soldier</English>
+            <German>Unsichtbares Ziel - Soldat</German>
+            <Chinese>隱形目標士兵</Chinese>
+            <French>Soldat cible invisible</French>
```
Elgin675, line 32:
```suggestion
            <French>Cible invisible - Soldat</French>
```

Elgin675 (changes requested):

PabstMirror (approved):

PR #1406, opened 2021-01-11
Add naked unit bug workaround

Kexanone:
- Fix acemod/ACEX#180

<details>
<summary>Video for ZEN</summary>

https://user-images.githubusercontent.com/18646707/104236594-1704a380-5457-11eb-9965-78bd3949cd54.mp4

</details>

### DIFF
--- a/addons/network/XEH_preInit.sqf
+++ b/addons/network/XEH_preInit.sqf
@@ -8,7 +8,7 @@ ADDON = false;
 [QGVAR(validateLoadout), {
     params ["_unit", "_loadout"];
 
-    if (GVAR(LoadoutValidation) > 0 && !(getUnitLoadout _unit isEqualTo _loadout)) then {
+    if (GVAR(loadoutValidation) > 0 && {(getUnitLoadout _unit) isNotEqualTo _loadout}) then {
         _unit setUnitLoadout _loadout;
     };
 }] call CBA_fnc_addEventHandler;
--- a/addons/network/XEH_preInit.sqf
+++ b/addons/network/XEH_preInit.sqf
@@ -8,7 +8,7 @@ ADDON = false;
 [QGVAR(validateLoadout), {
     params ["_unit", "_loadout"];
 
-    if (GVAR(loadoutValidation) > 0 && {(getUnitLoadout _unit) isNotEqualTo _loadout}) then {
+    if (GVAR(loadoutValidation) > 0 && {getUnitLoadout _unit isNotEqualTo _loadout}) then {
         _unit setUnitLoadout _loadout;
     };
 }] call CBA_fnc_addEventHandler;
--- a/addons/network/XEH_preInit.sqf
+++ b/addons/network/XEH_preInit.sqf
@@ -18,7 +18,7 @@ ADDON = false;
     if !(_local) then {
         // Broadcast loadout to new owner if unit was once local on this machine
         if (GVAR(loadoutValidation) > 0 && {_unit getVariable [QGVAR(wasLocal), false]}) then {
-            if (GVAR(loadoutValidation) == 1 && {playableUnits findIf {_x == _unit} == -1}) exitWith {};
+            if (GVAR(loadoutValidation) == 1 && {!(_unit in playableUnits)}) exitWith {};
             private _loadout = getUnitLoadout _unit;
             [QGVAR(validateLoadout), [_unit, _loadout], _unit] call CBA_fnc_targetEvent;
         };
--- a/addons/network/XEH_preInit.sqf
+++ b/addons/network/XEH_preInit.sqf
@@ -15,15 +15,14 @@ ADDON = false;
 
 ["CAManBase", "Local", {
     params ["_unit", "_local"];
-    if (_local) then {
-        _unit setVariable [QGVAR(wasLocal), true];
-    } else {
+    if !(_local) then {
         // Broadcast loadout to new owner if unit was once local on this machine
         if (GVAR(enableLoadoutValidation) && {_unit getVariable [QGVAR(wasLocal), false]}) then {
             private _loadout = getUnitLoadout _unit;
             [QGVAR(validateLoadout), [_unit, _loadout], _unit] call CBA_fnc_targetEvent;
         };
     };
+    _unit setVariable [QGVAR(wasLocal), _local];
 }] call CBA_fnc_addClassEventHandler;
 
 ["CAManBase", "Init", {
--- a/addons/network/XEH_preInit.sqf
+++ b/addons/network/XEH_preInit.sqf
@@ -38,9 +38,7 @@ ADDON = false;
         _unit setVariable [QGVAR(enableRestoreLoadout), true];
     } else {
         // Broadcast loadout to new owner if local was not triggered by init
-        if !(_unit getVariable [QGVAR(enableRestoreLoadout), false]) then {
-            _unit setVariable [QGVAR(enableRestoreLoadout), true];
-        } else {
+        if (_unit getVariable [QGVAR(enableRestoreLoadout), false]) then {
             private _loadout = getUnitLoadout _unit;
             [QGVAR(restoreLoadout), [_unit, _loadout], _unit] call CBA_fnc_targetEvent;
         };
--- a/addons/network/XEH_preInit.sqf
+++ b/addons/network/XEH_preInit.sqf
@@ -8,7 +8,7 @@ ADDON = false;
 [QGVAR(validateLoadout), {
     params ["_unit", "_loadout"];
 
-    if (GVAR(enableLoadoutValidation) && !(getUnitLoadout _unit isEqualTo _loadout)) then {
+    if (GVAR(LoadoutValidation) > 0 && !(getUnitLoadout _unit isEqualTo _loadout)) then {
         _unit setUnitLoadout _loadout;
     };
 }] call CBA_fnc_addEventHandler;
@@ -17,7 +17,8 @@ ADDON = false;
     params ["_unit", "_local"];
     if !(_local) then {
         // Broadcast loadout to new owner if unit was once local on this machine
-        if (GVAR(enableLoadoutValidation) && {_unit getVariable [QGVAR(wasLocal), false]}) then {
+        if (GVAR(LoadoutValidation) > 0 && {_unit getVariable [QGVAR(wasLocal), false]}) then {
+            if (GVAR(LoadoutValidation) == 1 && {playableUnits findIf {_x == _unit} == -1}) exitWith {};
             private _loadout = getUnitLoadout _unit;
             [QGVAR(validateLoadout), [_unit, _loadout], _unit] call CBA_fnc_targetEvent;
         };
--- a/addons/network/XEH_preInit.sqf
+++ b/addons/network/XEH_preInit.sqf
@@ -8,7 +8,7 @@ ADDON = false;
 [QGVAR(validateLoadout), {
     params ["_unit", "_loadout"];
 
-    if ((GVAR(enableLoadoutValidation) && !(getUnitLoadout _unit isEqualTo _loadout)) then {
+    if (GVAR(enableLoadoutValidation) && !(getUnitLoadout _unit isEqualTo _loadout)) then {
         _unit setUnitLoadout _loadout;
     };
 }] call CBA_fnc_addEventHandler;
--- a/addons/network/XEH_preInit.sqf
+++ b/addons/network/XEH_preInit.sqf
@@ -8,7 +8,7 @@ ADDON = false;
 [QGVAR(validateLoadout), {
     params ["_unit", "_loadout"];
 
-    if !(getUnitLoadout _unit isEqualTo _loadout) then {
+    if ((GVAR(enableLoadoutValidation) && !(getUnitLoadout _unit isEqualTo _loadout)) then {
         _unit setUnitLoadout _loadout;
     };
 }] call CBA_fnc_addEventHandler;
--- a/addons/network/XEH_preInit.sqf
+++ b/addons/network/XEH_preInit.sqf
@@ -33,9 +33,25 @@ ADDON = false;
 
 ["CAManBase", "Local", {
     params ["_unit", "_local"];
-    if !(_local) then {
-        private _loadout = getUnitLoadout _unit;
-        [QGVAR(restoreLoadout), [_unit, _loadout], _unit] call CBA_fnc_targetEvent;
+    if (_local) then {
+        // Enable loadout restoration for units transferred to the machine
+        _unit setVariable [QGVAR(enableRestoreLoadout), true];
+    } else {
+        // Broadcast loadout to new owner if local was not triggered by init
+        if !(_unit getVariable [QGVAR(enableRestoreLoadout), false]) then {
+            _unit setVariable [QGVAR(enableRestoreLoadout), true];
+        } else {
+            private _loadout = getUnitLoadout _unit;
+            [QGVAR(restoreLoadout), [_unit, _loadout], _unit] call CBA_fnc_targetEvent;
+        };
+    };
+}] call CBA_fnc_addClassEventHandler;
+
+// Enable loadout restoration for units spawned on the machine
+["CAManBase", "Init", {
+    params ["_unit"];
+    if (local _unit) then {
+        _unit setVariable [QGVAR(enableRestoreLoadout), true];
     };
 }] call CBA_fnc_addClassEventHandler;
 
--- a/addons/network/XEH_preInit.sqf
+++ b/addons/network/XEH_preInit.sqf
@@ -2,6 +2,8 @@
 
 ADDON = false;
 
+#include "initSettings.sqf"
+
 // Restore loadouts lost by the naked unit bug
 [QGVAR(validateLoadout), {
     params ["_unit", "_loadout"];
@@ -9,7 +11,6 @@ ADDON = false;
     if !(getUnitLoadout _unit isEqualTo _loadout) then {
         _unit setUnitLoadout _loadout;
     };
-
 }] call CBA_fnc_addEventHandler;
 
 ["CAManBase", "Local", {
@@ -18,7 +19,7 @@ ADDON = false;
         _unit setVariable [QGVAR(wasLocal), true];
     } else {
         // Broadcast loadout to new owner if unit was once local on this machine
-        if (_unit getVariable [QGVAR(wasLocal), false]) then {
+        if (GVAR(enableLoadoutValidation) && {_unit getVariable [QGVAR(wasLocal), false]}) then {
             private _loadout = getUnitLoadout _unit;
             [QGVAR(validateLoadout), [_unit, _loadout], _unit] call CBA_fnc_targetEvent;
         };
--- a/addons/network/XEH_preInit.sqf
+++ b/addons/network/XEH_preInit.sqf
@@ -1,0 +1,42 @@
+#include "script_component.hpp"
+
+ADDON = false;
+
+// Restore loadouts lost by the naked unit bug
+[QGVAR(restoreLoadout), {
+    params ["_unit", "_loadout"];
+    if (count _loadout < 7) exitWith {};
+
+    // Check if loadout got lost
+    private _loadoutLost = False;
+    {
+        (_loadout select _forEachIndex) params [["_backup", "", [""]]];
+        // Reject backup if it contains less information
+        if (_backup isEqualTo "" && !(_x isEqualTo "")) exitWith {};
+        if !(_x isEqualTo _backup) exitWith {
+            _loadoutLost = True;
+        };
+    } forEach [
+        primaryWeapon _unit,
+        secondaryWeapon _unit,
+        handgunWeapon _unit,
+        uniform _unit,
+        vest _unit,
+        backpack _unit,
+        headgear _unit
+    ];
+
+    if (_loadoutLost) then {
+        _unit setUnitLoadout _loadout;
+    };
+}] call CBA_fnc_addEventHandler;
+
+["CAManBase", "Local", {
+    params ["_unit", "_local"];
+    if !(_local) then {
+        private _loadout = getUnitLoadout _unit;
+        [QGVAR(restoreLoadout), [_unit, _loadout], _unit] call CBA_fnc_targetEvent;
+    };
+}] call CBA_fnc_addClassEventHandler;
+
+ADDON = true;
--- a/addons/network/XEH_preInit.sqf
+++ b/addons/network/XEH_preInit.sqf
@@ -17,7 +17,7 @@ ADDON = false;
     params ["_unit", "_local"];
     if !(_local) then {
         // Broadcast loadout to new owner if unit was once local on this machine
-        if (GVAR(LoadoutValidation) > 0 && {_unit getVariable [QGVAR(wasLocal), false]}) then {
+        if (GVAR(loadoutValidation) > 0 && {_unit getVariable [QGVAR(wasLocal), false]}) then {
             if (GVAR(loadoutValidation) == 1 && {playableUnits findIf {_x == _unit} == -1}) exitWith {};
             private _loadout = getUnitLoadout _unit;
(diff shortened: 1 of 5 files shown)


jonpas:
This contains commits from Malaria-infected civilians.

Kexanone:
> This contains commits from Malaria-infected civilians. Intentional?

Yeah, I forgot to mention that this should be merged after #1344, since it uses `characters` addon introduced in that commit.

> Possibly, I still don't think it will fix it 100% under some special conditions (indicating other problems though).

It has at least all the parts we suggested

Kexanone:
Moved it to `network`

addons/network/XEH_preInit.sqf
```diff
@@ -0,0 +1,42 @@
...
+        backpack _unit,
+        headgear _unit
+    ];
+
+    if (_loadoutLost) then {
+        _unit setUnitLoadout _loadout;
+    };
+}] call CBA_fnc_addEventHandler;
+
+["CAManBase", "Local", {
```
PabstMirror, line 34: when someone jips this is triggered for every unit in the game
sending a flood of 150+ events that each include a full loadout array is not good for network traffic

maybe just wait 10 seconds and then add the local EH?

commy2:
Okay, but why do I have to trash my performance every time someone connects to the server just because someone doesn't know how to use `setUnitLoadout` correctly?

addons/network/XEH_preInit.sqf
jonpas: This should be improved yeah.

jonpas:
> just because someone doesn't know how to use `setUnitLoadout` correctly?

This doesn't have anything to do with the use of `setUnitLoadout`. This is an Arma bug with transferring units to different machines, ref. https://github.com/acemod/ACEX/pull/163 and https://github.com/acemod/ACEX/issues/180.

jonpas:
Could also add an activation gate, similar to player EH's - don't run if nothing is using it. If it can't be made fast enough in other ways.

addons/network/XEH_preInit.sqf
```diff
@@ -0,0 +1,42 @@
...
+        };
+    } forEach [
+        primaryWeapon _unit,
+        secondaryWeapon _unit,
+        handgunWeapon _unit,
+        uniform _unit,
+        vest _unit,
+        backpack _unit,
+        headgear _unit
+    ];
```
commy2, line 27: If the unit is supposed to be naked, why check all this stuff?

commy2:
>This is an Arma bug with transferring units to different machines

Why not let the one that transfers the units check this then? Maybe a function that does the transfering and does sanity checks for loadouts and whatever afterwards instead of this.

jonpas:
CBA shouldn't do any kind of transferring, that's too implementation specific. That also makes function implementation more convoluted. Ideally, we catch all bug states and fix them automagically.

addons/network/XEH_preInit.sqf
jonpas: They aren't guaranteed to be naked that's the bug, it doesn't always happen. Most notable cases are loss of everything, but sometimes they lose only vest, or weapon etc.

We should avoid setting loadout if unnecessary I think.

Kexanone:
> CBA shouldn't do any kind of transferring, that's too implementation specific. That also makes function implementation more convoluted. Ideally, we catch all bug states and fix them automagically.

I second that. In addition the bug also happens when Zeus disconnects, since the units get transferred to the server automatically. This can only be captured with a `local` EH.

addons/network/XEH_preInit.sqf
Kexanone: Hmm, figuring out a good timing not to mess up jips and still supporting HC transfer scripts is going to be tricky.
Alternatively, we only trigger it when `_local` is `true`. The problem is to figure out on, which machine the units were previously though.

Kexanone: It's even worse. Every time a unit gets created, the other machines trigger local as well.
9ca6b6a fixes this.

addons/network/XEH_preInit.sqf
```diff
@@ -0,0 +1,56 @@
...
+
+ADDON = false;
+
+// Restore loadouts lost by the naked unit bug
+[QGVAR(restoreLoadout), {
+    params ["_unit", "_loadout"];
+    if (count _loadout < 7) exitWith {};
+
+    // Check if loadout got lost
+    private _loadoutLost = False;
```
commy2, line 11: false

addons/network/XEH_preInit.sqf
```diff
@@ -0,0 +1,56 @@
...
+    {
+        (_loadout select _forEachIndex) params [["_backup", "", [""]]];
+        // Reject backup if it contains less information
+        if (_backup isEqualTo "" && !(_x isEqualTo "")) exitWith {};
+        if !(_x isEqualTo _backup) exitWith {
+            _loadoutLost = True;
```
commy2, line 17: true

addons/network/XEH_preInit.sqf
```diff
@@ -0,0 +1,56 @@
...
+[QGVAR(restoreLoadout), {
+    params ["_unit", "_loadout"];
+    if (count _loadout < 7) exitWith {};
+
+    // Check if loadout got lost
+    private _loadoutLost = False;
+    {
+        (_loadout select _forEachIndex) params [["_backup", "", [""]]];
+        // Reject backup if it contains less information
+        if (_backup isEqualTo "" && !(_x isEqualTo "")) exitWith {};
```
commy2, line 15: is isNotEqualTo in current Arma?

addons/network/XEH_preInit.sqf
```diff
@@ -0,0 +1,56 @@
...
+        };
+    };
+}] call CBA_fnc_addClassEventHandler;
+
+// Enable loadout restoration for units spawned on the machine
+["CAManBase", "Init", {
+    params ["_unit"];
+    if (local _unit) then {
+        _unit setVariable [QGVAR(enableRestoreLoadout), true];
+    };
```
commy2, line 53: explain this

addons/network/XEH_preInit.sqf
Kexanone: Unfortunately, not yet.

addons/network/XEH_preInit.sqf
Kexanone: I typed too much Python recently xD

addons/network/XEH_preInit.sqf
Kexanone: The comment is just above. The restoration will be enabled once the unit becomes local to the machine, which can happen in two cases:
- Unit is placed on the machine
- Unit is transferred to the machine

The rogue trigger only happens on the machines where the unit is not local.

addons/network/XEH_preInit.sqf
commy2: I still don't get this part at all. Shouldn't it literally just be:
```sqf
[QGVAR(restoreLoadout), {
    params ["_unit", "_loadout"];

    if !(getUnitLoadout _unit isEqualTo _loadoutLost) then {
        _unit setUnitLoadout _loadout;
    };
```
???

addons/network/XEH_preInit.sqf
```diff
@@ -0,0 +1,56 @@
...
+["CAManBase", "Local", {
+    params ["_unit", "_local"];
+    if (_local) then {
+        // Enable loadout restoration for units transferred to the machine
+        _unit setVariable [QGVAR(enableRestoreLoadout), true];
+    } else {
+        // Broadcast loadout to new owner if local was not triggered by init
+        if (_unit getVariable [QGVAR(enableRestoreLoadout), false]) then {
+            private _loadout = getUnitLoadout _unit;
+            [QGVAR(restoreLoadout), [_unit, _loadout], _unit] call CBA_fnc_targetEvent;
```
commy2, line 43: Shouldn't `enableRestoreLoadout` be reset to `false` here as well? Isn't the meaning of the variable just `isLocal` / `wasLocal`?

addons/network/XEH_preInit.sqf
commy2: The name of the event is `restoreLoadout`, but we don't know if it actually restores a loadout. What it actually does is check the loadout and repair it if broken. "validateLoadout"?

commy2:
Still should be opt-in.

jonpas:
> Still should be opt-in.

Agreed, ACEX Headless also requires specific enabling it, because not everyone is experiencing the bug in large amounts (most likely to do with amount of players).

addons/network/XEH_preInit.sqf
Kexanone: Maybe I haven't described the problem well. When a unit is created, the `local` EH gets triggered on all machines where the unit is not local. This trigger is also added to the jip queue, which explains the issue that @PabstMirror has described. So we essentially only need to suppress the `local` EH on all machines where the unit is not local at the beginning.

addons/network/XEH_preInit.sqf
Kexanone: > The name of the event is `restoreLoadout`, but we don't know if it actually restores a loadout. What it actually does is check the loadout and repair it if broken. "validateLoadout"?

Agreed, `validateLoadout` makes more sense.

> I still don't get this part at all. Shouldn't it literally just be:
> 
> ```sqf
> [QGVAR(restoreLoadout), {
>     params ["_unit", "_loadout"];
> 
>     if !(getUnitLoadout _unit isEqualTo _loadout) then {
>         _unit setUnitLoadout _loadout;
>     };
> }] call CBA_fnc_addEventHandler;
> ```
> 
> ???

Yeah, this will already to. I initially though that we should just check a subset of items like it was done for the ACEX implementation. Moreover, the "Reject backup if it contains less information" part is actually no longer needed, since I noticed that this exception no longer occurs, since the rogue trigger was dealt with.

addons/network/XEH_preInit.sqf
```diff
@@ -0,0 +1,35 @@
...
+["CAManBase", "Local", {
+    params ["_unit", "_local"];
+    if (_local) then {
+        _unit setVariable [QGVAR(wasLocal), true];
+    } else {
+        // Broadcast loadout to new owner if unit was once local on this machine
+        if (_unit getVariable [QGVAR(wasLocal), false]) then {
+            private _loadout = getUnitLoadout _unit;
+            [QGVAR(validateLoadout), [_unit, _loadout], _unit] call CBA_fnc_targetEvent;
+        };
```
commy2, line 24: Why not reset the variable?

commy2 (dismissed):
make it opt in

Kexanone:
I turned it off by default.

commy2:
User setting is a terrible way to make this opt in? Should be a toggle that can be used by the ACEX thing that does the switching, or anyone else that does the switching. I see no point to let the end user decide whether this runs or not.

The toggle can be a simple main config entry.

Kexanone:
> User setting is a terrible way to make this opt in? Should be a toggle that can be used by the ACEX thing that does the switching, or anyone else that does the switching. I see no point to let the end user decide whether this runs or not.
> 
> The toggle can be a simple main config entry.

Hmm, ACEX also had it as a user setting. Is there any convention/example for adding such settings as config entry for CBA?

jonpas:
> I see no point to let the end user decide whether this runs or not.

> because not everyone is experiencing the bug in large amounts (most likely to do with amount of players).

But I would still keep it on an activation by another mod, let's say ACEX calls a CBA function to activate it. That would also make more sense to the user, to not have to search for relevant setting in a whole other category. Config would not be ideal as you can't change that mid-game.

commy2:
Not sure if function works. You have to start recording the was/isLocal states from the start of the game.

Kexanone:
> Not sure if function works. You have to start recording the was/isLocal states from the start of the game.

`wasLocal` are recorded from the start independent of the settings, so definitely doable.

addons/network/XEH_preInit.sqf
Kexanone: Fixed in 145733b

addons/network/XEH_preInit.sqf
```diff
@@ -0,0 +1,35 @@
...
+#include "initSettings.sqf"
+
+// Restore loadouts lost by the naked unit bug
+[QGVAR(validateLoadout), {
+    params ["_unit", "_loadout"];
+
+    if !(getUnitLoadout _unit isEqualTo _loadout) then {
+        _unit setUnitLoadout _loadout;
+    };
+}] call CBA_fnc_addEventHandler;
```
commy2, line 14: Event should do nothing if this stuff is not enabled, for security reasons.

Cyruz143:
As per Slack CBA chats. It would be nice for me to have this only run on units whos locality changes from server to client, e.g. players connecting in to AI slots that were previously owned by the server. This will also sometimes trigger the naked bug.

Kexanone:
> As per Slack CBA chats. It would be nice for me to have this only run on units whos locality changes from server to client, e.g. players connecting in to AI slots that were previously owned by the server. This will also sometimes trigger the naked bug.

Such a unidirectional limitation makes little sense. Units can also get naked when they are transferred to the server when a player disconnects.

On a different note, seeing that there is interest in this workaround even beyond mod makers, keeping it as a CBA setting maybe isn't that stupid after all...

Cyruz143:
> 
> 
> > As per Slack CBA chats. It would be nice for me to have this only run on units whos locality changes from server to client, e.g. players connecting in to AI slots that were previously owned by the server. This will also sometimes trigger the naked bug.
> 
> Such a unidirectional limitation makes little sense. Units can also get naked when they are transferred to the server when a player disconnects.
> 
> Edit: Actually, with the workaround as a CBA setting, you could already do this by only enabling it on the server.
> 
> On a different note, seeing that there is interest in this workaround even beyond mod makers, keeping it as a CBA setting maybe isn't that stupid after all...

You've misunderstood, as per Slack chat with commy, 

> enabled for all units / enabled for playable units / disabled

Kexanone:
> enabled for all units / enabled for playable units / disabled

Ah, I see.  Yes, this could be an additional option ofc.

addons/network/XEH_preInit.sqf
Kexanone: What kind of argument is that even? Anyone that can call a CBA EH could do a lot more harm than calling this one...

commy2: Except that you can use EG setUnitLoadout on the server without using the command on client.

commy2:
I also wrote that I don't want this to be a setting. It should be enabled by config / mission config / preInit global flag really.

addons/network/XEH_preInit.sqf
Kexanone: > I also wrote that I don't want this to be a setting. It should be enabled by config / mission config / preInit global flag really.

Why not? What is so good about denying anyone this workaround that isn't a mod or mission maker?

Kexanone:
> I also wrote that I don't want this to be a setting. It should be enabled by config / mission config / preInit global flag really.

Why not? What is so good about denying anyone this workaround that isn't a mod or mission maker?

commy2:
Because people are dumb and check everything even if they don't understand what it means. This fix is not free and won't help in 99% of played missions at all.

commy2:
Also, settings themselves are associated with an overhead.

Kexanone:
Alright, at this point I'll just go with the config option and be done with it...

jonpas:
I also don't exactly agree with a setting in CBA itself. It should be activated externally.

For that however, config option is not enough, there must be something that can activate it on runtime, for use by ZEN/ACEX.

Kexanone:
> I also don't exactly agree with a setting in CBA itself. It should be activated externally.
> 
> For that however, config option is not enough, there must be something that can activate it on runtime, for use by ZEN/ACEX.

So what? We are kicking it out as a setting from CBA only to reintroduce it as a setting in our other mods? Dunno what your plan for ACEX is. Zeus is quite vulnerable to this bug. If it's not an user option, I would just activate it for ZEN by config.

jonpas:
In ACEX, it is a user option, and I am not sure I would want to activate it full time, as not everyone is experiencing the bug. But maybe I should just do that...

Kexanone:
> ACEX Headless needs to be enabled. Which means I need to be able to turn that on or off based on the user enabling ACEX Headless. It is pointless to run this if there is no transferring going on at all.

If you don't need the headless component at all, wouldn't you just throw it out?

> I am not throwing CBA Setting out directly, but I think it is very unfriendly to the user if they have to search for something unrelated. Maybe I am just thinking of this as a too specific thing to ZEN/ACEX.

I think it's very ACEX specific. It makes more sense to me to have it as an option in CBA network settings than somewhere as a Zeus setting.

Kexanone:
Oh well, how about putting the entire thing into a function that has to be execute during preinit by the mod that wants to use the fix? Is it really necessary to have it a setting that can be changed on the fly?

jonpas:
> Oh well, how about putting the entire thing into a function that has to be execute during preinit by the mod that wants to use the fix? Is it really necessary to have it a setting that can be changed on the fly?

Function executed during preInit is about the same as config option. Neither can be changed later. No, runtime is not necessary for ACEX (because it was never implemented when CBA Settings became a thing), but postInit would be required, as that's when it determines if it should activate Headless transfers or not.

> Edit: If so, we just remove the setting from CBA and leave `CBA_network_enableLoadoutValidation` as the switch that can be changed on the fly.

What do you mean a switch? Boolean global variable? Then mods/missions would overwrite each other.

---

I am leaning more and more towards a CBA Network setting, looking at it from a wider PoV.

Kexanone:
> > Oh well, how about putting the entire thing into a function that has to be execute during preinit by the mod that wants to use the fix? Is it really necessary to have it a setting that can be changed on the fly?
> 
> Function executed during preInit is about the same as config option. Neither can be changed later. No, runtime is not necessary for ACEX (because it was never implemented when CBA Settings became a thing), but postInit would be required, as that's when it determines if it should activate Headless transfers or not.

Yeah, you are absolutely right. It's not a better approach...

> > Edit: If so, we just remove the setting from CBA and leave `CBA_network_enableLoadoutValidation` as the switch that can be changed on the fly.
> 
> What do you mean a switch? Boolean global variable? Then mods/missions would overwrite each other.
> 

Yeah, that would have been the idea, but you have the said competition, which is so nicely handled by an already existing framework called CBA settings^^

jonpas:
Ok, I vote CBA Settings, it's the best way to go.

> This fix ... won't help in 99% of played missions at all.

That's not true, it will help every Zeus mission and every session where a headless client with some kind of transferring is used. Those form a large percentage of players (Zeus use specifically, remember, Zeus units spawn on client, if client leaves or commands transfer, it is the same type of transfer as to a headless client).

> Because people are dumb and check everything even if they don't understand what it means.

I don't know how this is relevant, we can't fix dumb people. You can only ever avoid dumb people by preventing them from using your product.

Kexanone:
If you are worried about stupid users trashing your server, just force disable it. As requested, 37449af prevents now the validation to have any effect on your machine when you disable the option.

jonpas:
It is inherently a global setting, so that should be `1`.

Kexanone:
> It is inherently a global setting, so that should be `1`.

Done

jonpas:
I forgot about this though

> Edit: Actually, with the workaround as a CBA setting, you could already do this by only enabling it on the server.

But that isn't very clear at all, so if we add this it should probably be a drop-down setting with 3 options as described above.

Kexanone:
> I forgot about this though
> 
> > Edit: Actually, with the workaround as a CBA setting, you could already do this by only enabling it on the server.
> 
> But that isn't very clear at all, so if we add this it should probably be a drop-down setting with 3 options as described above.

As I mentioned in the same comment, it doesn't make much sense to have it one-directional. I just pointed out that you could have different setting on different machines before, not that it makes sense to do so.

Kexanone:
Oh wait, yeah. I almost forgot. The additional option to make it only applicable to playable units I suppose. Is it really needed?

jonpas:
That's the one I meant. I think it would be nice, that way you don't get any overhead on AI units? I didn't look at it enough to know if it would actually be any better.

Kexanone:
> That's the one I meant. I think it would be nice, that way you don't get any overhead on AI units? I didn't look at it enough to know if it would actually be any better.

Alright, I added it as a third option.

addons/network/initSettings.sqf
```diff
@@ -0,0 +1,16 @@
+[
+    QGVAR(LoadoutValidation),
```
jonpas, line 2:
```suggestion
    QGVAR(loadoutValidation),
```

addons/network/XEH_preInit.sqf
```diff
@@ -0,0 +1,36 @@
...
+
+ADDON = false;
+
+#include "initSettings.sqf"
+
+// Restore loadouts lost by the naked unit bug
+[QGVAR(validateLoadout), {
+    params ["_unit", "_loadout"];
+
+    if (GVAR(LoadoutValidation) > 0 && !(getUnitLoadout _unit isEqualTo _loadout)) then {
```
jonpas, line 11:
```suggestion
    if (GVAR(loadoutValidation) > 0 && {(getUnitLoadout _unit) isNotEqualTo _loadout}) then {
```

addons/network/XEH_preInit.sqf
```diff
@@ -0,0 +1,36 @@
...
+        _unit setUnitLoadout _loadout;
+    };
+}] call CBA_fnc_addEventHandler;
+
+["CAManBase", "Local", {
+    params ["_unit", "_local"];
+    if !(_local) then {
+        // Broadcast loadout to new owner if unit was once local on this machine
+        if (GVAR(LoadoutValidation) > 0 && {_unit getVariable [QGVAR(wasLocal), false]}) then {
```
jonpas, line 20:
```suggestion
        if (GVAR(loadoutValidation) > 0 && {_unit getVariable [QGVAR(wasLocal), false]}) then {
```

addons/network/XEH_preInit.sqf
```diff
@@ -0,0 +1,36 @@
...
+            if (GVAR(LoadoutValidation) == 1 && {playableUnits findIf {_x == _unit} == -1}) exitWith {};
```
jonpas, line 21:
```suggestion
            if (GVAR(loadoutValidation) == 1 && {playableUnits findIf {_x == _unit} == -1}) exitWith {};
```

jonpas (changes requested):
Minor settings casing.

addons/network/XEH_preInit.sqf
jonpas: This condition can actually be merged with the first condition.

addons/network/XEH_preInit.sqf
commy2, line 21:
```sqf
             if (GVAR(LoadoutValidation) == 1 && {playableUnits findIf {_x == _unit} == -1}) exitWith {};
```
Why?????????????????????

```suggestion
            // Exit if unit is not playable and that option is enabled.
            if (GVAR(LoadoutValidation) == 1 && {!(_unit in playableUnits)}) exitWith {};
```

commy2: See below.

Kexanone: I second that "why?". No idea what I was thinking back then😅

jonpas (approved):

addons/network/XEH_preInit.sqf
```diff
@@ -8,7 +8,7 @@ ADDON = false;
 [QGVAR(validateLoadout), {
     params ["_unit", "_loadout"];
 
-    if (GVAR(LoadoutValidation) > 0 && !(getUnitLoadout _unit isEqualTo _loadout)) then {
+    if (GVAR(loadoutValidation) > 0 && {(getUnitLoadout _unit) isNotEqualTo _loadout}) then {
```
commy2, line 11:
```suggestion
    if (GVAR(loadoutValidation) > 0 && {getUnitLoadout _unit isNotEqualTo _loadout}) then {
```

jonpas: No, that is less clear.

PabstMirror (approved):

PR #1408, opened 2021-01-16
Remove outdated unit tests from arrays

Kexanone:
- Remove unit tests in `arrays` component that pass invalide arguments.

### DIFF
--- a/addons/arrays/test_findMax.sqf
+++ b/addons/arrays/test_findMax.sqf
@@ -56,10 +56,6 @@ TEST_TRUE(isNil "_result",_fn);
 _result = [1, "not a number", 3, 4, 5] call CBA_fnc_findMax;
 TEST_TRUE(isNil "_result",_fn);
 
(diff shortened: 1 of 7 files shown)


commy2:
None of these are outdated though. They also all pass, except `x\cba\addons\arrays\test_shuffle.sqf:26`, though I can't make out why that expects the count to be 1 either. Just ignore.

This pull request should be closed.

Kexanone:
> None of these are outdated though. They also all pass, except `x\cba\addons\arrays\test_shuffle.sqf:26`, though I can't make out why that expects the count to be 1 either. Just ignore.
> 
> This pull request should be closed.

My bad. You are right, it's only that `x\cba\addons\arrays\test_shuffle.sqf:26` that breaks the test. The wrong argument types only show a script error, but pass the test.

commy2:
Yeah, sanity is checked even when script errors are raised.

PR #1414, opened 2021-02-05
Remote-Promise system

X39:
Just noticed that i have had this "in the pipe" since 2018 ... finally creating the PR for it

- Add a promise system, allowing to poll the server, calling a local method afterwards

### DIFF
--- a/addons/promise/fnc_create.sqf
+++ b/addons/promise/fnc_create.sqf
@@ -1,0 +1,51 @@
+/* ----------------------------------------------------------------------------
+Function: CBA_fnc_promise_create
+
+Description:
+	Creates and executes a new Promise.
+
+Parameters:
+    _objects - Object or array of objects that perform Say <OBJECT, ARRAY>
+    _params  - [sound, maxTitlesDistance,speed] or "sound" <ARRAY, STRING>
+
+	_rcv - remoteExecCall target.
+	_mthd - either string (methodname) or code to execute on target.
+	_mthdArgs - custom arguments to pass to the target.
+	_args - args to pass to the callback (see next param)
+	_cb - callback for the promise (executed on current machine).
+          Will receive an array, with index 0 being what is passed include
+		  _args and index 1 being whatever is returned by the target invokation
+		  [_args, _mthdResult]
+
+Returns:
+    Identifier that allows to check if a promise is yet done.
+	See CBA_fnc_promise_done.
+
+Example:
+    (begin example)
+        [
+			random_player, "random_method", [],
+			[_someLocalVariable], {
+				params ["_args", "_result"];
+				_args params  ["_someLocalVariable"];
+				diag_log _result;
+				diag_log _someLocalVariable;
+			}
+		] call CBA_fnc_promise_create;
+    (end)
+
+Author:
+    X39
+---------------------------------------------------------------------------- */
+#include "script_component.hpp"
+params ["_rcv", "_mthd", "_mthdArgs", "_args", "_cb"];
+private _stamp = diag_tickTime;
+private _index = 0;
+isNil {
+	_index = GVAR(requests) find 0;
+	if (_index == -1) then
+	{ _index = GVAR(requests) pushBack [_args, _cb, _stamp]; }
+	else { GVAR(requests) set [_index, [_args, _cb, _stamp]]; }
+};
+[_index, _mthd, _mthdArgs] remoteExec ["Promise_Receiver", _rcv, false];
+[_index, _stamp]
--- a/addons/promise/fnc_done.sqf
+++ b/addons/promise/fnc_done.sqf
@@ -1,0 +1,24 @@
+/* ----------------------------------------------------------------------------
+Function: CBA_fnc_promise_done
+
+Description:
+    Checks if a given promise was processed yet.
+
+Parameters:
+    _promiseHandle - Handle returned by CBA_fnc_promise_create
+
+Returns:
+    Boolean, true if it is done yet, false if it is not.
+
+Example:
+    (begin example)
+        _promiseHandle call CBA_fnc_promise_done;
+    (end)
+
+Author:
+    X39
+---------------------------------------------------------------------------- */
+#include "script_component.hpp"
+params ["_index", "_stamp"];
+private _promise = Promise_Requests select 0;
+if (_promise isEqualTo 0) { true } else { _promise select 2 != _stamp }
--- a/addons/promise/fnc_init.sqf
+++ b/addons/promise/fnc_init.sqf
@@ -1,0 +1,28 @@
+/* ----------------------------------------------------------------------------
+Function: CBA_fnc_promise_init
+
+Description:
+    Initializes the required variabels for the promise system.
+
+Parameters:
+    Nothing
+
+Returns:
+    Nothing
+
+Example:
+    (begin example)
+        [] call CBA_fnc_promise_init;
+    (end)
+
+Author:
+    X39
+---------------------------------------------------------------------------- */
+#include "script_component.hpp"
+isNil {
+	if isNil QGVAR(requests) then {
+		GVAR(requests) = [];
+	};
+};
+
+nil
--- a/addons/promise/fnc_receiver.sqf
+++ b/addons/promise/fnc_receiver.sqf
@@ -1,0 +1,32 @@
+/* ----------------------------------------------------------------------------
+Function: CBA_fnc_promise_receiver
+
+Description:
+    Method that gets called when a sender has a request.
+	
+	WARNING! Not intended to be called by client-code!
+
+Parameters:
+    _id - the promise ID, local to the sender.
+    _mthd - The method to execute, either string (var name) or code.
+    _args - Arguments to pass to the method.
+
+Returns:
+    Nothing
+
+Example:
+    Nothing
+
+Author:
+    X39
+---------------------------------------------------------------------------- */
+#include "script_component.hpp"
+
+params ["_id", "_mthd", "_args"];
+private _res = if (_mthd isEqualType "") then {
+	_args call (missionNamespace getVariable _mthd);
+} else {
+	_args call _mthd;
+};
+[_id, _res] remoteExec ["Promise_Callback", remoteExecutedOwner, false];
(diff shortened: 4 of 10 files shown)


addons/promise/fnc_done.sqf
```diff
@@ -0,0 +1,24 @@
...
+    (begin example)
+        _promiseHandle call CBA_fnc_promise_done;
+    (end)
+
+Author:
+    X39
+---------------------------------------------------------------------------- */
+#include "script_component.hpp"
+params ["_index", "_stamp"];
+private _promise = Promise_Requests select 0;
```
PabstMirror, line 23:
```suggestion
private _promise = GVAR(requests) select 0;
```

addons/promise/fnc_receiver.sqf
```diff
@@ -0,0 +1,32 @@
...
+---------------------------------------------------------------------------- */
+#include "script_component.hpp"
+
+params ["_id", "_mthd", "_args"];
+private _res = if (_mthd isEqualType "") then {
+    _args call (missionNamespace getVariable _mthd);
+} else {
+    _args call _mthd;
+};
+[_id, _res] remoteExec [QGVAR(requests), remoteExecutedOwner, false];
```
commy2, line 31: Last time I checked, the RE target id and the return value of REOwner do not match up in loaded local hosted multiplayer games, so this is not safe.

addons/promise/fnc_receiver.sqf
```diff
@@ -0,0 +1,32 @@
...
+
+Example:
+    Nothing
+
+Author:
+    X39
+---------------------------------------------------------------------------- */
+#include "script_component.hpp"
+
+params ["_id", "_mthd", "_args"];
```
commy2, line 25: _mthd

->

    _method

addons/promise/fnc_done.sqf
```diff
@@ -0,0 +1,24 @@
...
+        _promiseHandle call CBA_fnc_promise_done;
+    (end)
+
+Author:
+    X39
+---------------------------------------------------------------------------- */
+#include "script_component.hpp"
+params ["_index", "_stamp"];
+private _promise = GVAR(requests) select 0;
+if (_promise isEqualTo 0) { true } else { _promise select 2 != _stamp }
```
commy2, line 24: Missing newline @eof

Also could be simplified to:

```sqf
_promise isEqualTo 0 || {_promise select 2 != _stamp}
```
probably.

addons/promise/fnc_create.sqf
```diff
@@ -0,0 +1,51 @@
...
+
+Parameters:
+    _objects - Object or array of objects that perform Say <OBJECT, ARRAY>
+    _params  - [sound, maxTitlesDistance,speed] or "sound" <ARRAY, STRING>
+
+    _rcv - remoteExecCall target.
+    _mthd - either string (methodname) or code to execute on target.
+    _mthdArgs - custom arguments to pass to the target.
+    _args - args to pass to the callback (see next param)
+    _cb - callback for the promise (executed on current machine).
```
commy2, line 15: _receiver
_method
_methodArgs
_callback

addons/promise/fnc_create.sqf
```diff
@@ -0,0 +1,51 @@
...
+            }
+        ] call CBA_fnc_promise_create;
+    (end)
+
+Author:
+    X39
+---------------------------------------------------------------------------- */
+#include "script_component.hpp"
+params ["_rcv", "_mthd", "_mthdArgs", "_args", "_cb"];
+private _stamp = diag_tickTime;
```
commy2, line 42: Should be noted then that this framework breaks on servers that run for 11 days.

commy2:
It bugs me that the function names are

CBA_fnc_promise_xyz

and not

cba_promise_fnc_xyz

Would prefer either PREP or camel case: CBA_fnc_promiseXyz

addons/promise/fnc_create.sqf
```diff
@@ -0,0 +1,51 @@
+/* ----------------------------------------------------------------------------
+Function: CBA_fnc_promise_create
+
+Description:
+    Creates and executes a new Promise.
+
+Parameters:
+    _objects - Object or array of objects that perform Say <OBJECT, ARRAY>
+    _params  - [sound, maxTitlesDistance,speed] or "sound" <ARRAY, STRING>
```
commy2, line 9: ?

addons/promise/fnc_create.sqf
```diff
@@ -0,0 +1,51 @@
...
+---------------------------------------------------------------------------- */
+#include "script_component.hpp"
+params ["_rcv", "_mthd", "_mthdArgs", "_args", "_cb"];
+private _stamp = diag_tickTime;
+private _index = 0;
+isNil {
+    _index = GVAR(requests) find 0;
+    if (_index == -1) then
+    { _index = GVAR(requests) pushBack [_args, _cb, _stamp]; }
+    else { GVAR(requests) set [_index, [_args, _cb, _stamp]]; }
```
commy2, line 48: This formatting is messed up.

addons/promise/script_component.hpp
```diff
@@ -0,0 +1,4 @@
+#define COMPONENT promise
+#include "\x\cba\addons\main\script_mod.hpp"
+
+#include "\x\cba\addons\main\script_macros.hpp"
```
commy2, line 4: Missing newline @ eof

addons/promise/fnc_init.sqf
```diff
@@ -0,0 +1,28 @@
+/* ----------------------------------------------------------------------------
+Function: CBA_fnc_promise_init
+
+Description:
+    Initializes the required variabels for the promise system.
```
commy2, line 5: "variables"

addons/promise/fnc_init.sqf
```diff
@@ -0,0 +1,28 @@
...
+    X39
+---------------------------------------------------------------------------- */
+#include "script_component.hpp"
+isNil {
+    if isNil QGVAR(requests) then {
+        GVAR(requests) = [];
+    };
+};
+
+nil
```
commy2, line 28: This function seems like bloat to me. Just run this in preInit and tell people to add requiredAddons if they really need this in preInit as well.

addons/promise/fnc_receiver.sqf
```diff
@@ -0,0 +1,32 @@
...
+
+Description:
+    Method that gets called when a sender has a request.
+    
+    WARNING! Not intended to be called by client-code!
+
+Parameters:
+    _id - the promise ID, local to the sender.
+    _mthd - The method to execute, either string (var name) or code.
+    _args - Arguments to pass to the method.
```
commy2, line 12: <data type> missing

addons/promise/fnc_done.sqf
```diff
@@ -0,0 +1,24 @@
...
+Function: CBA_fnc_promise_done
+
+Description:
+    Checks if a given promise was processed yet.
+
+Parameters:
+    _promiseHandle - Handle returned by CBA_fnc_promise_create
+
+Returns:
+    Boolean, true if it is done yet, false if it is not.
```
commy2, line 11: \<data type\>

addons/promise/fnc_done.sqf
```diff
@@ -0,0 +1,24 @@
+/* ----------------------------------------------------------------------------
+Function: CBA_fnc_promise_done
+
+Description:
+    Checks if a given promise was processed yet.
+
+Parameters:
+    _promiseHandle - Handle returned by CBA_fnc_promise_create
```
commy2, line 8: \<data type\>

addons/promise/fnc_create.sqf
commy2: Also \<data type\>

commy2:
Why does it constantly talk about "methods"? Shouldn't it be "functions" or is this specifically a thing about promises I am missing.

addons/promise/fnc_done.sqf
dedmen: Syntax error, missing then.
also if true then true else..
ugh

addons/promise/fnc_receiver.sqf
X39: What do you suggest as alternative?

addons/promise/fnc_create.sqf
X39: diag_tickTime is only used to get some value that changes enough to allow reusage of the same slots and checking in `isDone`. As long as it changes enough, it is fine. However, i may also rework it into a simple counter that counts up. That however could introduce a race condition

addons/promise/fnc_create.sqf
X39: Copy-Paste issue

addons/promise/fnc_init.sqf
X39: I am not aware of how to realize that with the macros in CfgFunctions.hpp `PATHTO_FNC(promise_init);`

X39:
> Why does it constantly talk about "methods"? Shouldn't it be "functions" or is this specifically a thing about promises I am missing.

Can alter that. There is nothing you are missing here.

X39:
> It bugs me that the function names are
> 
> CBA_fnc_promise_xyz
> 
> and not
> 
> cba_promise_fnc_xyz
> 
> Would prefer either PREP or camel case: CBA_fnc_promiseXyz

Which of the two would you prefer?

addons/promise/fnc_create.sqf
commy2: It will return NaN +INF after 11.5 days or so.

addons/promise/fnc_init.sqf
commy2: Just put it in XEH preInit.

commy2:
> Can alter that. There is nothing you are missing here.

Please do.

> Which one of the two would you prefer?

Camel case, since PREP is not used much for public functions due to the advantage of the functions viewer.

commy2 (changes requested):
Honestly, the CBA event system makes these promises either superfluous or promises should be wrappers around CBA events.

PR #1415, opened 2021-02-06
Remove use of __EVAL macro

PabstMirror:
HEMTT does not understand this macro
this feel very dirty, but it works

### DIFF
--- a/addons/ui/RscDisplayOptionsLayout.hpp
+++ b/addons/ui/RscDisplayOptionsLayout.hpp
@@ -12,7 +12,7 @@ class RscDisplayOptionsLayout {
             onMouseButtonUp = "with uiNamespace do {['mouseButtonUp',_this,''] call RscDisplayOptionsLayout_script;};";
         };
         #define ADD_ELEMENT(var1) class Element##var1: Element022 {\
-            idc = 12000 + var1;\
+            idc = 12##var1;\
         }
         ADD_ELEMENT(023);
         ADD_ELEMENT(024);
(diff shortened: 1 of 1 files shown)


mharis001 (approved):

kymckay (approved):

jonpas (approved):
This actually makes more sense if you look at it from a preprocessing view.

addons/ui/RscDisplayOptionsLayout.hpp
```diff
@@ -12,7 +12,7 @@ class RscDisplayOptionsLayout {
             onMouseButtonUp = "with uiNamespace do {['mouseButtonUp',_this,''] call RscDisplayOptionsLayout_script;};";
         };
         #define ADD_ELEMENT(var1) class Element##var1: Element022 {\
-            idc = __EVAL(12000 + var1);\
+            idc = 12##var1;\
```
commy2, line 15:
```suggestion
            idc = QUOTE(12000 + (var1));\
```

PabstMirror: right now it's a number which is ideal
but `1+1 = 11` type "addition" is weird and hard to understand
so I'm fine making it a statement that needs to be evaluated

commy2: It's a plus b. What's not to understand? This addresses the issue while keeping the original numbers.

PabstMirror: (12##023) = 12023 = (12000 + 023)

PR #1419, opened 2021-02-09
Use configOf command

PabstMirror:


veteran29 (approved):

jonpas (approved):

commy2 (approved):

PabstMirror:
Command also tested working on "CfgNonAIVehicles" type so we can optimize the common case
```
x = "ProxyFLAG_MiniUGV" createVehicle (getpos player);
x call CBA_fnc_getObjectConfig // returns expected
```

addons/main/script_mod.hpp
```diff
@@ -10,7 +10,7 @@
 #define VERSION_AR  MAJOR,MINOR,PATCHLVL,BUILD
 
 // MINIMAL required version for the Mod. Components can specify others..
-#define REQUIRED_VERSION 1.98
+#define REQUIRED_VERSION 2.00
```
dedmen, line 13: configOf is 2.02

PR #1420, opened 2021-02-11
XEH - Fix missing XEH on AoW objects (2.02)

PabstMirror:


commy2:
> GalleryDioramaUnit

Huh?

addons/xeh/CfgVehicles.hpp
```diff
@@ -397,4 +397,9 @@ class CfgVehicles {
     class Land_MapBoard_01_Map_Syndikat_Tanoa_F: Land_Map_blank_F {
         XEH_ENABLED;
     };
+
+    // AOW (2.02)
+    class GalleryDioramaUnit_01_base_F: Items_base_F {
+        XEH_ENABLED;
+    };
```
commy2, line 404: This change does not necessarily require 2.02, but I guess it is a good thing to update this anyway? Would merge the PR now if the requirement was not set.

PabstMirror: This is based on current RC and sometimes things change, so we shouldn't merge until 2.02 is officially released

PabstMirror: still needed in release

commy2 (approved):

PR #1423, opened 2021-02-14
deploy.py: Pylint fixes

Killswitch00:
* Added docstrings.
* Script can use either the Pygithub3 or PyGithub libraries.

- Fix pylint issues found in the deploy script
- Add the ability to use an alternative Python GitHub API library

### DIFF
--- a/tools/deploy.py
+++ b/tools/deploy.py
@@ -1,54 +1,59 @@
 #!/usr/bin/env python3
 
+"""
 ####################################
 # CBA automatic deployment script  #
 # ================================ #
 # This is not meant to be run      #
 # directly!                        #
 ####################################
+"""
 
 import os
 import sys
-import shutil
 import traceback
 import subprocess as sp
-from pygithub3 import Github
+from subprocess import SubprocessError
+try:
+    from pygithub3 import Github
+except ImportError:
+    from github import Github
 
-
+# https://github.com/CBATeam/CBA_A3/issues/754
 TRANSLATIONISSUE = 754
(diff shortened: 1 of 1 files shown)


tools/deploy.py
```diff
@@ -1,54 +1,59 @@
...
         return 1
     else:
-        print("Token sucessfully obtained.")
+        print("Token successfully obtained.")
 
     print("\nUpdating translation issue ...")
     try:
         update_translations(token)
-    except:
+    except (SubprocessError, Exception): # pylint: disable=broad-except
```
commy2, line 56: Trying to understand this. Doesn't `SubprocessError` inherit from `Exception` anyway?

Killswitch00: You are correct - it does. Pylint requires you to list the exact exceptions that can occur. The two GitHub API libraries are different in which exceptions they throw. The github library throws exceptions based on GithubException, but the pygithub3 library can throw several different exceptions that are all based on Exception.

The list is meant to convey that "either a subprocess module error occurs, or some other error from either the pygithub3 or github module is thrown", so it's  a half-way between list each exception specifically and "catch them all and ignore the pylint warning"

jonpas: Pylint follows mentality of only catching specifics so you get a traceaback on unknown errors. This doesn't exactly apply here where we want to catch any error though, as we rely on API to work correctly and that can sometimes not respond.

Killswitch00: Is the explanation sufficient or would you prefer just catching the general Exception, which will pass pylint due to the disable directive anyway?

jonpas: I think I would prefer catching any exception here, but I am fine with either really.

jonpas (approved):

PR #1438, opened 2021-03-15
Docs data Update for 3.15.3

jokoho48:
- Update the Data for the Docs for 3.15.3
- Add Groups for the new Modules
- Fix some Internal Functions were marked as public functions
- Fix some issues with the make_docs script to work on Ubuntu 20.10


addons/ai/fnc_waypointGarrison.sqf
```diff
@@ -1,6 +1,6 @@
 #include "script_component.hpp"
 /* ----------------------------------------------------------------------------
-Script: fnc_waypointGarrison
+Script: CBA_fnc_waypointGarrison
```
commy2, line 3: How many times do I have to reject this? It is not a function, but compiled as script by a waypoint via config entry. 

`CBA_fnc_waypointGarrison` does not exist. fnc_waypointGarrison.sqf does and it is only meant to be used as uncompiled script.

addons/xeh/fnc_supportMonitor.sqf
```diff
@@ -1,6 +1,6 @@
 #include "script_component.hpp"
 /* ----------------------------------------------------------------------------
-Function: CBA_fnc_supportMonitor
+Internal Function: CBA_fnc_supportMonitor
```
commy2, line 3: This has uses for debugging too.

jokoho48: but it needs that tag to not show up in the Docs

addons/ai/fnc_waypointGarrison.sqf
jokoho48: do we want to hide that from the Docs then? or what to do with it?

commy2: Hide it if you want, but it's no big deal. Do not name it "`CBA_fnc_waypointGarrison`" though, because that is not a thing.

addons/xeh/fnc_supportMonitor.sqf
commy2: Why shouldn't it show up in the docs?

jokoho48: because in the Description it says "Internal use only." but i can just remove the Internal an it will show in the Docs

tools/make_docs
```diff
@@ -1,38 +1,39 @@
 #!/bin/sh
```
Killswitch00, line 2: Add

```shopt -s xpg_echo```

to have Bash expand the escape sequences in echo. I tried that with dash (the default shell in Debian/Ubuntu) and everything still works.

addons/xeh/fnc_supportMonitor.sqf
Killswitch00: So, internal or not? The description in this file does not mention "internal use only".

jokoho48: @Killswitch00 that was exactly my point.

Killswitch00: OK - to clarify, commy2 asked why it should not show up in the docs, to which you replied 'because in the Description it says "Internal use only"'

However, the description does not say "Internal use only", so it appears to me that the file was fine as it was. In other words: the CBA_fnc_supportMonitor function was visible in the documentation before this PR and should remain visible in the documentation.

jokoho48: oh damn, I was in the totally wrong function.

Killswitch00 (approved):

PR #1439, opened 2021-03-15
Update docs

jokoho48:
- Update Docs to 3.15.3


Killswitch00 (approved):

Killswitch00 (changes requested):
The function index page looks strange - in the "C" part.

jokoho48:
> 
> 
> The function index page looks strange - in the "C" part.

that has always been like that. I don't think the tool can handle it differently. even the current version has that.

Killswitch00:
It would appear so.

Killswitch00 (approved):
Well, it's not worse than before.

PR #1442, opened 2021-03-28
use compileScript in script_macros_common.hpp

jokoho48:
- Title

ref https://community.bistudio.com/wiki/compileScript

### DIFF
--- a/addons/main/script_macros_common.hpp
+++ b/addons/main/script_macros_common.hpp
@@ -809,7 +809,7 @@ Author:
 
 // This only works for binarized configs after recompiling the pbos
 // TODO: Reduce amount of calls / code..
-#define COMPILE_FILE2_CFG_SYS(var1) compile preprocessFileLineNumbers var1
+#define COMPILE_FILE2_CFG_SYS(var1) compileScript [var1]
 #define COMPILE_FILE2_SYS(var1) COMPILE_FILE2_CFG_SYS(var1)
 
 #define COMPILE_FILE_SYS(var1,var2,var3) COMPILE_FILE2_SYS('PATHTO_SYS(var1,var2,var3)')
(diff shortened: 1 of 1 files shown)


veteran29 (approved):

PabstMirror:
Need to look at how this effects https://github.com/CBATeam/CBA_A3/blob/master/addons/xeh/fnc_compileEventHandlers.sqf#L86

commy2 (changes requested):
https://github.com/CBATeam/CBA_A3/blob/master/addons/xeh/fnc_compileEventHandlers.sqf#L86

tbh, changing ^ to accept both probably is a net loss. I don't see what's so great about this command.

jokoho48:
> 
> 
> https://github.com/CBATeam/CBA_A3/blob/master/addons/xeh/fnc_compileEventHandlers.sqf#L86
> 
> tbh, changing ^ to accept both probably is a net loss. I don't see what's so great about this command.

compileScript is from what dedmen wrote about 10 times faster than compile preprocessFileLineNumbers.
what we may want to look into than is not changing the macro but only the fnc_compileEventHandlers.sqf to use compileScript.

PabstMirror:
sqfc is a bit odd because it contains **both** the optimized version and the original input.  
And both are also part of the CODE returned by `compileScript.  
Any attempt to string-ify will just return the original sqf and the optimizations will be lost. Which is exactly what CBA does

https://github.com/CBATeam/CBA_A3/blob/master/addons/xeh/fnc_preStart.sqf#L69-L70
```
 // cache incompatible classes that are needed in preInit
GVAR(incompatibleClasses) = compileFinal str ([false, true] call CBA_fnc_supportMonitor);
```

e.g. `diag_dumpScriptAssembly compileFinal str compileScript ["some sqfc test"]`
will give the original sqf's assembly, not the optimized one
so I don't think we want to use `compileScript` inside fnc_compileEventHandlers

Another complication is that if we update macro but another addon doesn't use SQFC they will lose all the speed benefits of compiling once at preStart (for plain old sqf, there is no benefit to compileScript)

imho the best bet is to create a new macro `#define CALL_COMPILESCRIPT(x) call compileScript [QPATHTOF(x)];`
so addons can do exactly what they want

PabstMirror:
master (nofilepatching so it can use the fnc_compileEventHandlers optimization)
```
[CBA] (xeh) INFO: [0,11.2,0] PreStart started.
[CBA] (xeh) INFO: [0,15.916,0] PreStart finished.

[CBA] (xeh) INFO: [1356,36.242,0] PreInit started. v3.15.3.000000
[ACE] (common) INFO: Parsed Settings Configs [32.0 ms]
[CBA] (xeh) INFO: [1356,36.432,0] PreInit finished.
[CBA] (xeh) INFO: [1362,36.775,0] PostInit started.
[CBA] (xeh) INFO: [1362,36.863,0] PostInit finished.
```
sqfc (also no filepatching)
```
[CBA] (xeh) INFO: [0,9.692,0] PreStart started.
[CBA] (xeh) INFO: [0,10.34,0] PreStart finished.

[CBA] (xeh) INFO: [9980,184.56,0] PreInit started. v3.15.3.000000
[ACE] (common) INFO: Parsed Settings Configs [27.0 ms]
[CBA] (xeh) INFO: [9980,184.745,0] PreInit finished.
[CBA] (xeh) INFO: [9982,185.057,0] PostInit started.
[CBA] (xeh) INFO: [9982,185.147,0] PostInit finished.
```

loading a sqfc is so fast we don't need to worry about it not supporting the caching optimization

commy2 (approved):

PabstMirror:
I think this should be closed and replaced with the new macro `COMPILE_SCRIPT` in https://github.com/CBATeam/CBA_A3/pull/1450

from XEH's point of view that will let us know which to cache
`compile preProcessFileLineNumbers` - we know it's plain old sqf and can cache it at prestart
`compileScript` - might be sqfc, call each mission

PR #1219, opened 2019-09-05
Update issue templates

jonpas:
**When merged this pull request will:**
- Make use of new issue template framework from GitHub.
- Split bug report, feature request and enhancement idea templates - shows up as separate buttons when creating new issues.
- Empty issue can still be created with a button below templates.

PR #1449, opened 2021-04-25
Remove Feature And Enhancement Request Templates

commy2:
- This PR removes the "Feature Request" and "Enhancement Request" templates when opening an issue on GitHub.

Why?

Github is a nice platform to discuss code changes and document bugs. However, letting anybody post their "great" ideas on the issue tracker leads to a few problems:

- The issues tab is already used for serious and not so serious bugs. Having (in most cases badly labeled) requests mixed in distracts from these.
- Requests clog up the issue tracker with "issues" that will realistically never get addressed anyway.
- These requests have a terrible signal to noise ratio, which can be seen by [checking their track record so far](https://github.com/CBATeam/CBA_A3/issues?q=is%3Aissue+Is+your+enhancement+related+to+a+problem%3F+is%3Aclosed)
- Requests will be opened without the author ever checking if the feature is even desirable or feasible AT ALL. The bar is simply too low compared with pull requests.
- These templates encourage this unwated behavior and legitimize it.
- These templates were added a few years ago without ever having been discussed with me

The issues tab should be reserved for actual bug reports. If a good idea is brought up, a dev can still open an issue describing the enhancement or feature briefly and even assign it to themselves; or tell the person with the idea (after vetting it for feasibility) to open an issue in their name, which has been a way more successful procedure in my experience.


jonpas:
> These templates were added a few years ago without ever having been discussed with me

https://github.com/CBATeam/CBA_A3/pull/1219 You approved and merged them actually. ;)

---

Note as a reminder: If this gets merged, wiki pages need to be updated as well.

commy2:
Ok, then I appologize. However, I do want to revert this change.

Killswitch00 (approved):

PR #1450, opened 2021-05-08
Add sqfc support

PabstMirror:


tools/make.py
```diff
@@ -392,6 +392,7 @@ def copy_optionals_for_building(mod,pbos):
 
     try:
         for dir_name in src_directories:
+            if (os.path.isfile(os.path.join(optionals_root,dir_name))): continue # ignore .pbos from hemmit
```
jonpas, line 395:
```suggestion
            # Ignore PBOs from HEMMT
            if (os.path.isfile(os.path.join(optionals_root,dir_name))):
                continue
```

sqfc.json
```diff
@@ -0,0 +1,17 @@
...
+    "excludeList": [
+        "\\initsettings.sqf",
+        "\\initkeybinds.sqf",
+        "\\xeh_prep.sqf",
+        "\\backwards_comp.sqf",
+        "settings\\gui_createCategory.sqf"
+    ],
+    "outputDir": "P:/",
+    "workerThreads": 12
+}
```
jonpas, line 17:
```suggestion
}

```

jonpas (approved):

commy2:
I know it literally uses `compileScript`, but can we get a better name on `COMPILESCRIPT_FILE` that somehow explains what is going on?

PabstMirror:
Open to suggestions

commy2 (approved):

PabstMirror:
I disabled compiling on 
```
        "diagnostic\\fnc_initExtendedDebugConsole.sqf",
        "xeh\\fnc_initDisplay.sqf",
        "xeh\\fnc_startLoadingScreen.sqf",
        "xeh\\fnc_endLoadingScreen.sqf"
```
as they all have the basic format of 
```
call {
    #include "\a3\functions_f\GUI\fn_initDisplay.sqf"
};
// cba code here
```

If A3 were to change these functions then it could cause issues because CBA would be stuck on a older version

commy2:
> If A3 were to change these functions then it could cause issues because CBA would be stuck on a older version

Huh? I don't get it.

PabstMirror:
When compiling it will take the current version in your p-drive and "statically links" it into the sqfc
So if in arma 2.08 BIS changes something in `\a3\functions_f\GUI\fn_initDisplay.sqf`, the sqfc would still have the old version.

By not compiling, it will do the `#include "` at preStart, just like it works now before the sqfc stuff.
It's probably never going to be an issue, but it's a tiny bit safer to just not do these 4 files.

commy2:
Not a perfect solution though. Can't we mark it somehow in the script file that it should not be proessed like that? That seems more explicit and maintainable.

PabstMirror:
JSON doesn't support comments for some reason
could add comments as data, but I don't really like that

it's just the few functions we hijack via `class CfgFunctions { class A3{ ....`

PabstMirror:
This PR is ready so I'll merge, but it does not necessarily mean that next CBA release will be using compiled SQFC
we're still using make.py//pboProject so we'll have to manually compile

severgun:
what is a problem to add compile stage to make.py?

PabstMirror:
probably 1 line to call `compile_sqf.py`
tbh I'd rather do it manually for the first few releases to look for problems

dedmen:
> Not a perfect solution though. Can't we mark it somehow in the script file that it should not be proessed like that? That seems more explicit and maintainable.

https://github.com/dedmen/ArmaScriptCompiler/issues/12

Just add `#pragma SQFC_NO_COMPILE` to top of the file.
Its not implemented yet, but I will

PR #1452, opened 2021-05-09
Common - Remove editor init fix

PabstMirror:
2.04
>Fixed: Not being able to use scripts in Eden Editor that do not return Nothing


commy2:
Nice. I think that this solved another issue as well though, but I can't remember it.

commy2 (approved):

veteran29 (approved):

PR #1453, opened 2021-05-10
Use execVM for all unit tests

Dahlgren:
- Run as unit tests with execVM instead of call

```sqf
execVM "\x\cba\addons\main\test.sqf"
```

557 Test OK
1 Test FAIL


commy2:
Looks pretty good.

Except that there is a short hand for

```sqf
0 spawn compile preprocessFileLineNumbers
```

which is the dreaded `execVM`. :)

commy2 (approved):

Killswitch00 (approved):

PR #1454, opened 2021-05-10
Fix shuffle test

Dahlgren:
- Fix shuffle test

### DIFF
--- a/addons/arrays/test_shuffle.sqf
+++ b/addons/arrays/test_shuffle.sqf
@@ -21,6 +21,7 @@ TEST_OP(count _result,==,count _original,_fn);
     TEST_OP(_x,in,_original,_fn);
 } forEach _result;
 
+// Result will be empty array because the function expects an array as the first param (test throws a param error)
 _original = [1, 2, 3];
 _result = _original call CBA_fnc_shuffle;
 TEST_OP(_result,isEqualTo,[],_fn);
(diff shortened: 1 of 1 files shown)


addons/arrays/test_shuffle.sqf
```diff
@@ -21,10 +21,6 @@ TEST_OP(count _result,==,count _original,_fn);
     TEST_OP(_x,in,_original,_fn);
 } forEach _result;
 
-_original = [1, 2, 3];
-_result = _original call CBA_fnc_shuffle;
-TEST_OP(count _result,==,1,_fn);
-
 _original = [];
 _result = [_original] call CBA_fnc_shuffle;
 TEST_OP(count _result,==,count _original,_fn);
```
commy2, line 26: To me it seems like, after the above test is removed, this one fails with the non-equation 1 == 0.

commy2:
Ah, I missed that. Could just replace the assert line with

```sqf
 TEST_OP(_result,isEqualTo,[],_fn); 
```

Dahlgren:
Sure, will do

PabstMirror (approved):

addons/arrays/test_shuffle.sqf
```diff
@@ -23,7 +23,7 @@ TEST_OP(count _result,==,count _original,_fn);
```
PabstMirror, line 23:
```suggestion

// Result will be empty array because the function expects an array as the first param (test throws a param error)
```

PR #1457, opened 2021-05-22
Replace "Community Base Addons -" by "CBA" in the `Network` menu

Elgin675:
- For better consistency with the rest of the mod, the "Community Base Addons -" prefix has been replaced by "CBA" for the Network menu.

As a reminder, here is the current appearance, before modification :
CBA_network_menu


commy2:
Why do you have CBA Weapons twice?

PabstMirror (approved):

jonpas:
Commy, they aren't twice, one is currently selected, other is from the list.

PR #1459, opened 2021-05-23
Add function to create UUIDs

BaerMitUmlaut:
- Add a simple UUID generator function

### DIFF
--- a/addons/common/CfgFunctions.hpp
+++ b/addons/common/CfgFunctions.hpp
@@ -130,6 +130,7 @@ class CfgFunctions {
             PATHTO_FNC(waitAndExecute);
             PATHTO_FNC(waitUntilAndExecute);
             PATHTO_FNC(compileFinal);
+            PATHTO_FNC(createUUID);
         };
 
         class Broken {
--- a/addons/common/fnc_createUUID.sqf
+++ b/addons/common/fnc_createUUID.sqf
@@ -1,0 +1,40 @@
+#include "script_component.hpp"
+/* ----------------------------------------------------------------------------
+Function: CBA_fnc_createUUID
+
+Description:
+    Creates a version 4 UUID (universally unique identifier).
+
+Parameters:
+    None
+
+Returns:
+    UUID [String]
+
+Example:
+    (begin example)
+        private _uuid = call CBA_fnc_createUUID;
+    (end)
+
+Author:
+    BaerMitUmlaut
+--------------------------------------------------------------------------- */
+SCRIPT(createUUID);
+
+private _hexDigits = [
+    "0", "1", "2", "3", "4", "5", "6", "7",
+    "8", "9", "a", "b", "c", "d", "e", "f"
+];
+private _versionByte = "4";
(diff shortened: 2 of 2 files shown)


commy2:
It's not unique.

BaerMitUmlaut:
I didn't chose that name.

commy2:
I chose to complain about it.

veteran29 (approved):

commy2 (changes requested):
not unique

BaerMitUmlaut:
This is an implementation that meets the RFC spec. If you want to complain about semantics, you'll have to complain to the IETF.

commy2:
Not challenging the spec. But I am saying that this is not suited for anything in Arma.

BaerMitUmlaut:
Arma can only do floats, so in use cases where you can potentially exceed 2^23 IDs, UUIDs would be useful. The chance of a collision is ridiculously low and way less of a risk than running out of numbers.

commy2:
There is no need to do this when everyone already has a unique client id.

BaerMitUmlaut:
Okay, and that helps how exactly?

commy2:
Those are unique and therefore useful. I see no use case for this.

BaerMitUmlaut:
Now you have one unique ID per client. What if I need 2^24 unique IDs per client? Say, for example, PFH IDs?

commy2:
> What if I need 2^24 unique IDs per client?

I would use a function that actually provides unique ids, so not this one.

BaerMitUmlaut:
Well, then provide a better one. They generated UUIDs are more than good enough, but feel free to prove me wrong. However I'm not willing to wait a hundred years for you to generate a duplicate.

jonpas (approved):
Looks fine to me.

This shouldn't be a discussion of its use, it's a standard implementation and CBA as a general library can and should provide those if there is any use for it. It is up to the library consumer to use it or not, there can even be multiple implementations of a "unique ID" generator in CBA.

commy2:
> CBA as a general library can and should provide those if there is any use for it

What is the use case? Providing unique ids is not one of them.

BaerMitUmlaut:
Providing an ID that is unique within the next 100 years.

jokoho48 (approved):

dedmen:
You can use https://community.bistudio.com/wiki/systemTimeUTC to make it mooooooooost likely unique, down to the millisecond.

BaerMitUmlaut:
The OS time is by default not that accurate, but sure, let's say it is.

The chances of a duplicate UUID are far lower than those of a duplicate timestamp. Imagine assigning a unique ID to every player on a server. If that code is executed at the same time for every client, the chances of an overlap are probably higher than 1/1000. Consider that this code might be executed many times, not just per day, but across multiple servers and for every session. You're going to have duplicates.

jonpas:
Merge it @BaerMitUmlaut!

BaerMitUmlaut:
I lack the rights to merge.

PR #1460, opened 2021-06-02
Fix CBA_fnc_turretDir for certain UAV turrets

PabstMirror:
For "AR-2 Darter" (little quad copter) `CBA_fnc_turretDir` always returns `[0,0]`


commy2 (approved):
Why is this a thing, BI?

PR #1473, opened 2021-07-12
Add missing includes for SQFC

Dahlgren:
- Fix ~4~3 failing compilations with ArmaScriptCompiler when using files in `include` folder


commy2:
urg

PabstMirror:
Looks like I meant to disable `fn_isDebugConsoleAllowed` but got it mixed up with `fnc_initExtendedDebugConsole`
so I can fix the sqfc.json and we don't need fn_isDebugConsoleAllowed in the `/include`

ref https://github.com/CBATeam/CBA_A3/pull/1450#issuecomment-846632661

Dahlgren:
Nice, one less file 🙂

commy2:
I don't like this. Who maintains these copy pasted files? Just pull them from the vanilla project files.

jonpas:
Commy, they are includes for build tools, so P-drive is not required. You don't have to use them.

jonpas (approved):

PR #1484, opened 2021-07-25
Revert "Common - Remove editor init fix"

jokoho48:
- Reverts changed from #1452 because it breaks usage of local variables in init/editor fields

### DIFF
--- a/addons/common/Cfg3DEN.hpp
+++ b/addons/common/Cfg3DEN.hpp
@@ -1,3 +1,9 @@
+
+#define POS_H_GROUP(N) (5 + N * 3.5) * (pixelH * pixelGrid * 0.50)
+#define POS_H_BACKGROUND(N) (N * 3.5 + 0.6 * 5) * (pixelH * pixelGrid * 0.50)
+#define POS_H_TITLE(N) (N * 3.5 + 1 * 5) * (pixelH * pixelGrid * 0.50)
+#define POS_H_VALUE(N) N * 3.5 * (pixelH * pixelGrid * 0.50)
+
 class ctrlCombo;
 
 class Cfg3DEN {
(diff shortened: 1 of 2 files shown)


commy2:
Local variables being possible in init boxes due to the CBA hack is not reason enough to re-introduce this feature. I do not recommend this PR to be merged, but instead to tell people that local variables are not possible in init boxes, with or without ("vanilla") the newest CBA version.

jokoho48:
> 
> 
> Local variables being possible in init boxes due to the CBA hack is not reason enough to re-introduce this feature. I do not recommend this PR to be merged, but instead to tell people that local variables are not possible in init boxes, with or without ("vanilla") the newest CBA version.

the point is that the users got used to beeing able to use private variables in init/editorCode boxes, and this will create alot of false errors/error reports to not use CBA but also other mods.

commy2:
Or we could make dedmen implement them properly.

jokoho48:
> 
> 
> Or we could make dedmen implement them properly.

and as long that is not done we should keep the current fix.

PabstMirror:
@dedmen - do you know if this could be fixed in 2.06
otherwise I would say we merge

dedmen:
wuht? local variables how?
local vars shouldn't be a problem, if they are then fixi fixi.
Need repro

commy2:
I can't repro it in vanilla either.

commy2:
Some more details in this German thread: https://armaworld.de/forum/thread/5089-call-befehl-in-init-feld-nach-cba-update-auf-v3-15-5-210713/?postID=40944

dedmen:
But still no repro.
Need repro or else no fix

jokoho48:
@dedmen 
0. load with latest CBA version v3.15.5.210713
1. place a trigger 
2. set activation to Radio Alpha
3. put this as condition
4. put snipped into activation
```sqf
private _grp1 = [getPosMarkerPos "aPosition", east, 5] call BIS_fnc_spawnGroup;
[_grp1] call CBA_fnc_taskDefend
```
5. press ok
6. see error Popup

dedmen:
Actually the association "error displayed == doesn't work" is not correct.
The "Have to return nothing" thing was also just a bogus error, I simply removed the error message to make it work.
If you would insert a script another way it would've worked just fine.

dedmen:
Fixed
Arma3NoOptInt_DX11_x64_FuaX18Yo9N

commy2:
Thank you dedmen!!! Now people can finally use that editor box to the fullest, I won't have to read any myths about what a "global space" is supposed to be, we can close this pull request, and another ugly hack is gone from CBA.

dedmen:
Programmers didn't want to implement new errors messages (neither did I) so the generic "variable namespace" error is just reused everywhere where its about variables.

Local namespace doesn't exist?
"Local variable in global space"
Missing the underscore in params?
"Local variable in global space"

Issue #1356, opened 2020-07-13
Allow Disposable Launchers to be replaced in Backpacks


Eogos:
**Is your enhancement related to a problem?**
Currently the Disposable Launcher system, when active, replaces instances of the ammo with the disposable launcher it goes to, however backpacks are excluded from this and only have the ammo removed but no corresponding launcher added to replace the now removed ammo. This can lead to instances of scenarios designed with a certain amount of AT in mind being harder due to missing AT launchers depending on the settings of the host.

**Solution you'd like:**
Ammo in backpacks should be handled the same as every other container and have the ammo directly replaced with disposable launchers rather than just removed.

**Alternatives you've considered:**
Not replacing ammo with launchers anywhere, for consistency.

**Additional context:**
This came up after discussion on the CUP discord where it was learned that backpacks are specifically excluded for some unknown reason and a request to make an issue report was made. To me and some others, it does not make sense to exclude backpacks from the replacement process specifically when it is still done everywhere else.

commy2:
By "instances of the ammo" and "Ammo in backpacks", you mean the loaded launcher classes?

Drofseh:
The idea is that a rocket should get turned into a rocket launcher, regardless of what container it's in.

Eogos:
What Drofseh said

commy2:
No, magazines should not turn into launchers.

Drofseh:
They do though, everywhere except backpacks.

This means that if a mod or mission maker gives a unit a launcher and an additional magazine in their backpack (2 shots total) then when played with CBA the extra magazine gets deleted but not replaced, so the unit now has only a single shot. 

The contention is that magazines in backpacks ought to be replaced with a launcher in the same manner as magazines that are in any other container. 

https://github.com/CBATeam/CBA_A3/blob/ce4cda6320a9b1526cb9c06491c6b818143e1c28/addons/disposable/fnc_replaceMagazineCargo.sqf#L59

dedmen:
Never knew that they get replaced.. because I usually only have the ammo in backpack. Why would anyone put launcher ammo in uniform or vest anyway 🤔

jonpas:
No it isn't. There are various smaller disposable launchers.

bux:
> There are various smaller disposable launchers.

![image](https://user-images.githubusercontent.com/1235520/87405351-9c4def00-c5bf-11ea-8d91-0eae41f46786.png)

commy2:
[...]

PR #1485, opened 2021-07-27
Disposable - Replace magazines in backpack when launcher fits backpacks

10Dozen:
- Allow replacement of magazines in backpacks with launchers, if loaded launcher item fits backpacks (has backpack slot id in `launcher_loaded >> WeaponSlotsInfo >> allowedSlots` array, implements https://github.com/CBATeam/CBA_A3/issues/1356)

### DIFF
--- a/addons/disposable/XEH_preInit.sqf
+++ b/addons/disposable/XEH_preInit.sqf
@@ -27,7 +27,7 @@ GVAR(NormalLaunchers) = [] call CBA_fnc_createNamespace;
 GVAR(LoadedLaunchers) = [] call CBA_fnc_createNamespace;
 GVAR(UsedLaunchers) = [] call CBA_fnc_createNamespace;
 GVAR(magazines) = [];
-GVAR(BackpackReplaceableMagazines) = [];
+GVAR(BackpackLaunchers) = createHashMap;
 GVAR(MagazineLaunchers) = [] call CBA_fnc_createNamespace;
 
 private _cfgWeapons = configFile >> "CfgWeapons";
@@ -36,8 +36,8 @@ private _cfgMagazines = configFile >> "CfgMagazines";
 {
     private _launcher = configName _x;
     private _magazine = configName (_cfgMagazines >> (getArray (_cfgWeapons >> _launcher >> "magazines") select 0));
-    private _replaceInBackpack = [_cfgMagazines >> _magazine, QGVAR(launcherFitsBackpack), 0] call BIS_fnc_returnConfigEntry > 0;
     getArray _x params ["_loadedLauncher", "_usedLauncher"];
+    private _fitsInBackpacks = TYPE_BACKPACK in getArray(configFile >> "CfgWeapons" >> _loadedLauncher >> "WeaponSlotsInfo" >> "allowedSlots");
 
     GVAR(LoadedLaunchers) setVariable [_launcher, _loadedLauncher];
     GVAR(UsedLaunchers) setVariable [_launcher, _usedLauncher];
@@ -50,8 +50,8 @@ private _cfgMagazines = configFile >> "CfgMagazines";
         GVAR(MagazineLaunchers) setVariable [_magazine, _loadedLauncher];
     };
 
-    if (_replaceInBackpack) then {
-        GVAR(BackpackReplaceableMagazines) pushBackUnique _magazine;
+    if (_fitsInBackpacks) then {
+        GVAR(BackpackLaunchers) set [_loadedLauncher, true];
     };
 
     // check if mass entries add up
--- a/addons/disposable/XEH_preInit.sqf
+++ b/addons/disposable/XEH_preInit.sqf
@@ -27,6 +27,7 @@ GVAR(NormalLaunchers) = [] call CBA_fnc_createNamespace;
 GVAR(LoadedLaunchers) = [] call CBA_fnc_createNamespace;
 GVAR(UsedLaunchers) = [] call CBA_fnc_createNamespace;
 GVAR(magazines) = [];
+GVAR(BackpackReplaceableMagazines) = [];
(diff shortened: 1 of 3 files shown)


addons/disposable/XEH_preInit.sqf
```diff
@@ -27,6 +27,7 @@ GVAR(NormalLaunchers) = [] call CBA_fnc_createNamespace;
 GVAR(LoadedLaunchers) = [] call CBA_fnc_createNamespace;
 GVAR(UsedLaunchers) = [] call CBA_fnc_createNamespace;
 GVAR(magazines) = [];
+GVAR(BackpackReplaceableMagazines) = [];
```
commy2, line 30:
```suggestion
GVAR(BackpackLaunchers) = createHashMap;
```

addons/disposable/XEH_preInit.sqf
```diff
@@ -35,6 +36,7 @@ private _cfgMagazines = configFile >> "CfgMagazines";
 {
     private _launcher = configName _x;
     private _magazine = configName (_cfgMagazines >> (getArray (_cfgWeapons >> _launcher >> "magazines") select 0));
+    private _replaceInBackpack = [_cfgMagazines >> _magazine, QGVAR(launcherFitsBackpack), 0] call BIS_fnc_returnConfigEntry > 0;
```
commy2, line 39:
```suggestion
    private _fitsInBackpacks = TYPE_BACKPACK in getArray(configFile >> "CfgWeapons" >> _launcher >> "WeaponSlotsInfo" >> "allowedSlots");
```
with
```sqf
#define TYPE_VEST 701
#define TYPE_UNIFORM 801
#define TYPE_BACKPACK 901
```
somewhere on top or macro header script_components.hpp file.

addons/disposable/XEH_preInit.sqf
```diff
@@ -48,6 +50,10 @@ private _cfgMagazines = configFile >> "CfgMagazines";
         GVAR(MagazineLaunchers) setVariable [_magazine, _loadedLauncher];
     };
 
+    if (_replaceInBackpack) then {
+        GVAR(BackpackReplaceableMagazines) pushBackUnique _magazine;
```
commy2, line 54:
```suggestion
    if (_fitsInBackpacks) then {
        GVAR(BackpackLaunchers) set [_launcher, true];
```

addons/disposable/fnc_replaceMagazineCargo.sqf
```diff
@@ -56,7 +56,7 @@ private _isBackpack = getNumber (configOf _box >> "isBackpack") != -1;
     _x params ["_magazine", "_ammo"];
 
     if (_magazine in GVAR(magazines)) then {
-        if !(_isBackpack) then {
+        if (!_isBackpack || {_magazine in GVAR(BackpackReplaceableMagazines)}) then {
             _box addWeaponCargoGlobal [GVAR(MagazineLaunchers) getVariable _magazine, 1];
```
commy2, line 60:
```suggestion
        private _launcher = GVAR(MagazineLaunchers) getVariable _magazine;
        if (!_isBackpack || _launcher in GVAR(BackpackLaunchers)) then {
            _box addWeaponCargoGlobal [_launcher, 1];
```

commy2 (changes requested):
1. No need to make a config entry for this, because it already is one with the "WeaponsSlotsInfo\allowedSlots" array.
2. No need to maintain a list when we can abuse HashMaps as sets. Lookup in list is O(n) at worts, while lookup in set is O(1).

Disclaimer, suggestions not tested, so there may be typos or logic holes.

10Dozen:
> commy2 is like: I'll never add this feature! 
Also commy2: I rewrite every single line of the feature's PR!

`>_<`

addons/disposable/XEH_preInit.sqf
commy2: May need to be moved down and _launcher be `_loadedLauncher`.

commy2:
Such is life.

addons/disposable/XEH_preInit.sqf
10Dozen: Should i ~~break~~ change other lists/namespaces into hashMaps too?

commy2:
lgtm

Did you test if it works?

10Dozen:
E.g. for CUP, as i understand, some ammo was intentionally made lighter/smaller than launcher to make ATs more vanilla-like. But when converted to launchers - disposals takes too much space. 
I feel like some extra space check is needed on conversion in that case (e.g. PzF3 + 3 rockets converts to PzF3 + 1 extra PzF3 in backpack, the rest is ignored as there is no more space).

10Dozen:
> Did you test if it works?

Yeah, works fine.

commy2 (approved):

commy2:
> Should i break change other lists/namespaces into hashMaps too?

I don't know. It seems not really necessary.

addons/disposable/XEH_preInit.sqf
```diff
@@ -36,6 +37,7 @@ private _cfgMagazines = configFile >> "CfgMagazines";
     private _launcher = configName _x;
     private _magazine = configName (_cfgMagazines >> (getArray (_cfgWeapons >> _launcher >> "magazines") select 0));
     getArray _x params ["_loadedLauncher", "_usedLauncher"];
+    private _fitsInBackpacks = TYPE_BACKPACK in getArray(configFile >> "CfgWeapons" >> _loadedLauncher >> "WeaponSlotsInfo" >> "allowedSlots");
```
commy2, line 40:
```suggestion
     private _fitsInBackpacks = TYPE_BACKPACK in getArray (configFile >> "CfgWeapons" >> _loadedLauncher >> "WeaponSlotsInfo" >> "allowedSlots");
```
We need a PEP-8 of SQF.

commy2:
Good job :)

PR #1486, opened 2021-07-28
Item Context Menu - Change option selection to Single click instead of Double click

10Dozen:
- Change the method of the option selection in Item Context Menu -- select and execute option via single click (instead of default double click)

**Note:**
Currently to use inventory action player needs to double click the item, then find needed option and double click it again. 
It's work fine in tests and without much interactions available.

But after hours of playing it becomes not so trivial. Especially when there is a need to interact with multiple items in short period of time (e.g. pack individual magazines manually - 4x clicks (to open menu and select options) * number of items in backpack of the average milsimer = ∞ number of clicks!). 
So new options is just a small step to reduce number of the clicks needed.

p.s. Maybe option to open menu via Ctrl+Click is also needed, but didn't investigated possibility of this.


Dystopian:
Also maybe single scroll click is a good idea instead of double left click for menu call (if it possible).

10Dozen:
> Also maybe single scroll click is a good idea instead of double left click for menu call (if it possible).
Not a problem for slots (e.g. uniform, rifle), but hard to make for items in containers. 

Now context menu invoked by LBDblClick event (returns selected element). If we replace it with MouseButtonUp for LMB - it will work (click will select item in the listbox, so it's possible to track curSel element), but now option menu will be opened each time we click somewhere inside listbox too. For MMB it won't even select item in listbox, so it's not easy to get item that was clicked.

Maybe if we can track mouse coursor over specific listbox item...

commy2:
This setting seems like bloat to me. It is a setting about how you can either use single or double clicks in an ui...

CBA devs should make the choice whatever makes more sense. I am open to this being changed, but a setting for this? Just no.

jonpas:
Single click makes more sense to me.

10Dozen:
I prefer single click (reason in PR topic). Removed setting.

jonpas:
Does this single-click implementation work reliably?

10Dozen:
I don't have A3 now and it's like 1 year old, so someone needs to test =)
If UI handling didn't change, then it should work.

jonpas:
Could put this into next update if somebody tests it.

Mike-MF:
I tested this, it works.

PR #1487, opened 2021-07-30
Have CBA_fnc_players return player logic entities

Freddo3000:
- Make CBA_fnc_players return player logic entities

Strange that there's a check for headless clients even though there was seemingly no way for the function to list logic entities previously. This allows the function to list player spectator/Zeus logic entities and similar.

### DIFF
--- a/addons/common/fnc_players.sqf
+++ b/addons/common/fnc_players.sqf
@@ -8,7 +8,7 @@ Description:
     Unlike "BIS_fnc_listPlayers", this function will not report the game logics of headless clients.
 
 Parameters:
-    None
+    _listLogics - List player game logics (Zeus, Spectator, etc). (optional, default: false) <BOOL>
 
 Returns:
     List of all player objects <ARRAY>
@@ -23,4 +23,11 @@ Author:
 ---------------------------------------------------------------------------- */
 SCRIPT(players);
 
-(units sideLogic + allUnits + allDeadMen) select {isPlayer _x && {!(_x isKindOf "HeadlessClient_F")}}
+params [["_listLogics", false]];
+
+private _units = allUnits + allDeadMen;
+if _listLogics then {
+    _units append units sideLogic;
+};
+
+_units select {isPlayer _x && {!(_x isKindOf "HeadlessClient_F")}}
--- a/addons/common/fnc_players.sqf
+++ b/addons/common/fnc_players.sqf
@@ -23,4 +23,4 @@ Author:
 ---------------------------------------------------------------------------- */
 SCRIPT(players);
 
-(allUnits + allDeadMen) select {isPlayer _x && {!(_x isKindOf "HeadlessClient_F")}}
+(units sideLogic + allUnits + allDeadMen) select {isPlayer _x && {!(_x isKindOf "HeadlessClient_F")}}


PabstMirror:
I'm worried about side effects from including virtual units in this list

e.g. 
acre_sys_rack uses `_player = ([] call CBA_fnc_players) select 0;` as a backup default value
I'm not sure what will happen if it now uses spectator or virtual-zeus for that (but I haven't tested it)

perhaps we should add an optional arg to include virtual entities?

Freddo3000:
> I'm worried about side effects from including virtual units in this list
> 
> e.g.
> acre_sys_rack uses `_player = ([] call CBA_fnc_players) select 0;` as a backup default value
> I'm not sure what will happen if it now uses spectator or virtual-zeus for that (but I haven't tested it)
> 
> perhaps we should add an optional arg to include virtual entities?

I've got similar thoughts, but right now the description of the function seems to imply that logic entities should be included (but currently aren't).

PabstMirror:
>Reports all (human) player objects

I don't think a spectator is human

Freddo3000:
> > Reports all (human) player objects
> 
> I don't think a spectator is human

Fair enough, was more looking at the Headless Client part.

addons/common/fnc_players.sqf
```diff
@@ -8,7 +8,7 @@ Description:
     Unlike "BIS_fnc_listPlayers", this function will not report the game logics of headless clients.
 
 Parameters:
-    None
+    _listLogics - List player game logics (Zeus, Spectator, etc). (optional, default: false) <BOOL>
```
jonpas, line 11:
```suggestion
    _includeLogics - Include player game logics (Zeus, Spectator, etc). <BOOLEAN> (default: false)
```

addons/common/fnc_players.sqf
```diff
@@ -23,4 +23,11 @@ Author:
 ---------------------------------------------------------------------------- */
 SCRIPT(players);
 
-(allUnits + allDeadMen) select {isPlayer _x && {!(_x isKindOf "HeadlessClient_F")}}
+params [["_listLogics", false]];
+
+private _units = allUnits + allDeadMen;
+if _listLogics then {
```
jonpas, line 29:
```suggestion
if (_includeLogics) then {
```

addons/common/fnc_players.sqf
```diff
@@ -23,4 +23,11 @@ Author:
 ---------------------------------------------------------------------------- */
 SCRIPT(players);
 
-(allUnits + allDeadMen) select {isPlayer _x && {!(_x isKindOf "HeadlessClient_F")}}
+params [["_listLogics", false]];
```
jonpas, line 26:
```suggestion
params [["_includeLogics", false]];
```

addons/common/fnc_players.sqf
```diff
@@ -23,4 +23,11 @@ Author:
...
+
+private _units = allUnits + allDeadMen;
+if _listLogics then {
+    _units append units sideLogic;
```
jonpas, line 30:
```suggestion
    _units append (units sideLogic);
```

jonpas (changes requested):

jonpas:
I am not sure about this, seems like a very very edge case. And are you sure `allUnits` doesn't include headless clients? When a headless client entity is marked as "playable" it should be included.

Freddo3000:
> I am not sure about this, seems like a very very edge case. And are you sure `allUnits` doesn't include headless clients? When a headless client entity is marked as "playable" it should be included.

I can't say off the top of my head, I don't have a workable Arma install at the moment to check and this PR is two years old. Feel free to commit your suggestions, as I'm getting an github error trying to commit them via the web.

jonpas:
Fair. I think best way to go about this is to remove the `!(_x isKindOf "HeadlessClient_F")` check from the function. @Mike-MF has confirmed for us that `allUnits` never returns Headless Clients entities, be it playable, player etc.

I believe this condition originally came from the [deprecation of this function in favour of `allPlayers`, but that was quickly changed as `allPlayers` returns virtual entities](https://github.com/Freddo3000/CBA_A3/commit/fe3aaf2d5bdf105e13a58acfe13d01c278601964). This condition was probably left in by mistake of first writing it using `allPlayers`, but was changed to `allUnits + allDead(Men)` - this is a guess as commit history only confirms the first part of this theory.

Anyways, as for this function returning logics, one should just use `allPlayers` command for that, I see no reason for additional argument. Wiki has some nice examples of filtering that too.

Alternatively, we just don't change the function at all, as there might have been a forgotten case of `allUnits` or `allDead(Men)` returning Headless Clients... Now that I think about it `allDead(Men)` wiki says it includes agents, so there might be a case of dead Headless Client entities?

I will close this and if someone wants to do more research and provide proof, I would be happy to change it.

PR #1489, opened 2021-08-03
Accessory - Removes action from pointers while not in slot

10Dozen:
- fnc_switchAttachment operates only mounted items, but inventory action was available everywhere (and didn't work). Pull request changes action availabness to `POINTER` and `OPTICS` slots.

NOTE:
Currently if you place pointer with laser/flashlight MRT options (or optics) to e.g. untiform container - it will still have enabled 'Next rail item state' action in the item context menu. But when selected - item is not changed (because called function only works when item is attached to a gun).

### DIFF
--- a/addons/accessory/XEH_preInit.sqf
+++ b/addons/accessory/XEH_preInit.sqf
@@ -31,7 +31,7 @@ if (!hasInterface) exitWith {};
 ] call CBA_fnc_addItemContextMenuOption;
 
 [
-    "##AccessorySights", "POINTER", [LSTRING(opticNext), LSTRING(opticNext_tooltip)], nil, nil, {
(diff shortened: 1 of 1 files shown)


commy2:
These are supposed to work on all slots, not just pointers?

10Dozen:
> These are supposed to work on all slots, not just pointers?
Doesn't make sense to me to be able to enable flashlight while LLM-01 is in the backpack. 
Maybe it will be useful for ACE map flashlight system, but not sure now.

commy2:
Now it makes more sense to me, but I don't understand the need for this change. What is actually broken here?

10Dozen:
Currently if you place pointer with laser/flashlight  MRT options to e.g. untiform container - it will still have enabled 'Next rail item state' action in the item context menu. But when selected - item is not changed (because called function only works when item is attached to a gun).

veteran29 (approved):

PabstMirror (approved):

Issue #1490, opened 2021-08-04
Item Context Menu - Longer Display Names get cut off


Drofseh:
**Mods (complete and add to the following information):**
- **Arma 3:** `2.04`
- **CBA:** `v3.15.5.210713`

**Description:**
When adding a context menu option to an inventory item, the display name gets cut off (at 23 characters on UI small).

**Steps to reproduce:**
- Place a unit in the edit
- Play as that unit
- Run this code in the console
```sqf
[
    "#All",
    "ALL",
    [
        "Display Name34567890123456789012345678901234567890",
        "ToolTip."
    ],
    nil,
    nil,
    {
        true
    },
    {
        systemChat "Context Statement";
        diag_log "Context Statement";
        true
    },
    false,
    []
] call CBA_fnc_addItemContextMenuOption;
```
- See the Display Name get cut off.

**Expected behavior:**
The full display name should be visible, or at least a more reasonable limit on the number of characters.

**Where did the issue occur?**
- Editor (Singleplayer)

**Screenshots:**
![](https://i.gyazo.com/90d0c92e5c94931794a4626e7b108662.png)

PR #1492, opened 2021-08-06
Item Context Menu - Flexible menu width 

10Dozen:
- Max width of the Context Menu now calculated depending on the size of the actions names. If there are long names - menu will be wide, otherwise - with default width (fixes https://github.com/CBATeam/CBA_A3/issues/1490)

### DIFF
--- a/addons/ui/fnc_openItemContextMenu.sqf
+++ b/addons/ui/fnc_openItemContextMenu.sqf
@@ -48,6 +48,10 @@ if (_options isEqualTo []) exitWith {};
 // ctrlSetBackgroundColor command does not seem to work for RscListBox.
 private _list = _display ctrlCreate [QGVAR(ItemContextMenu), -1];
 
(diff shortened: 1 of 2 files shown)


jonpas (approved):

commy2:
Why multiply the length by 1.1? Doesn't this mean whitespace grows with text length?

10Dozen:
> Why multiply the length by 1.1? Doesn't this mean whitespace grows with text length?

I assume it was for some margin for text inside the control, but dumb one, agree. 
[BIKI:getTextWidth](https://community.bistudio.com/wiki/getTextWidth) mentions hardcoded 0.008 left and right margins. Guess this may be used instead.

10Dozen:
Tried to use only width + margins but as we are using listbox with pictures - we need to add margin for picture as well. I searched for it in defines and config, but didn't get any result. 

From tests it should be something like 0.05, but maybe somebody know exact value or different approach?

veteran29 (approved):

PR #1497, opened 2021-09-01
Fix CBA_fnc_removeMagazineCargo for ground containers

PabstMirror:
Based on https://github.com/acemod/ACE3/pull/8399

Test code
```
xx = "GroundWeaponHolder" createVehicle (getpos player); 
xx addMagazineCargo ["smokeshell", 2];
```
then
```
[xx, "SmokeShell", 1] call CBA_fnc_removeMagazineCargo;
```

left with nothing

### DIFF
--- a/addons/common/fnc_removeMagazineCargo.sqf
+++ b/addons/common/fnc_removeMagazineCargo.sqf
@@ -64,6 +64,13 @@ private _magazinesAmmo = magazinesAmmoCargo _container;
 // Clear cargo space and readd the items as long it's not the type in question
 clearMagazineCargoGlobal _container;
 
+// Engine will agressively cleanup "empty" ground containers, even if magazines are re-added in same frame, so re-create a new container
+private _containerType = typeOf _container;
+if (((toLower _containerType) in ["groundweaponholder", "weaponholdersimulated"]) 
+&& {(weaponCargo _container) isEqualTo []} && {(itemCargo _container) isEqualTo []} && {(backpackCargo _container) isEqualTo []}) then {
+    _container = createVehicle [_containerType, getPosATL _container, [], 0, "CAN_COLLIDE"];
+};
+
 {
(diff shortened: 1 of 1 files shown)


addons/common/fnc_removeMagazineCargo.sqf
```diff
@@ -64,6 +64,13 @@ private _magazinesAmmo = magazinesAmmoCargo _container;
 // Clear cargo space and readd the items as long it's not the type in question
 clearMagazineCargoGlobal _container;
 
+// Engine will agressively cleanup "empty" ground containers, even if magazines are re-added in same frame, so re-create a new container
+private _containerType = typeOf _container;
+if (((toLower _containerType) in ["groundweaponholder", "weaponholdersimulated"]) 
+&& {(weaponCargo _container) isEqualTo []} && {(itemCargo _container) isEqualTo []} && {(backpackCargo _container) isEqualTo []}) then {
```
jonpas, line 70:
```suggestion
if (((toLower _containerType) in ["groundweaponholder", "weaponholdersimulated"]) 
    && {(weaponCargo _container) isEqualTo []}
    && {(itemCargo _container) isEqualTo []}
    && {(backpackCargo _container) isEqualTo []}) then {
```
Think this is a bit more readable.

commy2:
Should use `isKindOf`, even if it requires two uses to handle child classes - e.g. containers spawned via editor ground placed items.

PabstMirror:
`class GroundWeaponHolder_Scripted: GroundWeaponHolder {`
Don't want it running on this
and editor placed will probably only have a single thing in them

commy2:
Maybe check the `forceSupply` entry then?

> and editor placed will probably only have a single thing in them

The weapons have at least a weapon and one magazine iirc.

dedmen:
Engine also checks "forceSupply" for whether its a ground holder.

commy2 (approved):

Issue #1363, opened 2020-08-21
Muzzle player event handler not returning documented returns


LorenLuke:
**Mods (complete and add to the following information):**
- **Arma 3:** `1.98.146373` 
- **CBA:** `v3.15.1` (workshop version as of 8/21/2020)

**Description:**
Muzzle player event handler not returning documented returns

**Steps to reproduce:**
- place the following code in your player unit:
```["muzzle", { 
    params ["_unit", "_newMuzzle", "_oldMuzzle"]; 
    systemChat format ["%1", _this]; 
}, true] call CBA_fnc_addPlayerEventHandler;
```
- Run scenario and change weapon
- See that the first parameter is, in fact, not the unit

**Expected behavior:**
Expected returns to be as documented under the 'muzzle' entry;

**Where did the issue occur?**
- Editor (Singleplayer)

**Log Files:**
- None

**Additional context:**
- N/A

**Screenshots:**
![image](https://user-images.githubusercontent.com/6456096/90920133-a45a3500-e39c-11ea-8955-a68e2fece197.png)
![image](https://user-images.githubusercontent.com/6456096/90920187-b3d97e00-e39c-11ea-90c8-31393f9d8c1b.png)
![image](https://user-images.githubusercontent.com/6456096/90920215-c9e73e80-e39c-11ea-80a0-f729e065c553.png)

Ansible2:
https://github.com/CBATeam/CBA_A3/blob/578b409cf91394866491a671a5a73b08f046a1d4/addons/events/fnc_playerEvent.sqf#L95
Just missing `_unit` from here.

`[QGVAR(muzzleEvent), [_unit, _newMuzzle select 2, GVAR(oldMuzzle) select 2]] call CBA_fnc_localEvent; `

PR #1502, opened 2021-09-18
Events - Make player muzzle event args consistent for retro

PabstMirror:
Fix https://github.com/CBATeam/CBA_A3/issues/1363

playerEvent.sqf does (just the 2 muzzles, no unit)
```
    if (_newMuzzle isNotEqualTo GVAR(oldMuzzle)) then {
        [QGVAR(muzzleEvent), [_newMuzzle select 2, GVAR(oldMuzzle) select 2]] call CBA_fnc_localEvent;
```
We need to make the two args arrays the same, and I think it's much safer to modify the retro one

### DIFF
--- a/addons/events/fnc_addPlayerEventHandler.sqf
+++ b/addons/events/fnc_addPlayerEventHandler.sqf
@@ -73,7 +73,7 @@ private _id = switch (_type) do {
     };
     case "muzzle": {
         if (_applyRetroactively) then {
-            [GVAR(oldUnit), currentMuzzle GVAR(oldUnit), ""] call _function;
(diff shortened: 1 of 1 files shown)


commy2:
I don't like this. It should pass unit-newMuzzle-oldMuzzle.

PabstMirror:
I agree it probably should have, but I think changing it now can break a lot more than this
This just changes the "retro" args which are called at most 0 to 1 times.
The playerEvent loop would be called many many more often.

IE, code written for the muzzle EH is going to be expecting [newMuzzle, oldMuzzle]

PR #1507, opened 2021-10-09
Common - Add support for commander turrets to `cba_common_fnc_turretDir`

PabstMirror:
Some turrets like the blu wheeled IFV don't have gunBeg defined because they don't have any weapons.

https://community.bistudio.com/wiki/eyeDirection
> Returns the direction object is watching (eyes, or a vehicle primary observer).


commy2:
What if the turret has `primaryObserver` as well as gunBeg and gunEnd?

PabstMirror:
this is after the `if (_gunEnd isEqualTo _gunBeg) then {` check so it's just a fallback, but they should be the same in either case

PabstMirror:
Tested on 
"B_APC_Wheeled_01_cannon_F"
"B_AFV_Wheeled_01_cannon_F"
"B_MBT_01_cannon_F"
"O_APC_Tracked_02_cannon_F"

All go from just `[0,0]` to giving actual angles that match the hud dir/elev display

PR #1508, opened 2021-10-11
Add bipod support for accessory/switchAttachment to header info

dedmen:


commy2:
From that function:
```sqf
if (_configs isNotEqualTo []) then {
    _switchItem = getText (_configs select 0);
};
```
Does that even work?

PR #1516, opened 2021-10-27
Add support for encoding and decoding JSON with hash maps

BaerMitUmlaut:
- Enable parsing JSON objects as hash maps
- Enable encoding hash maps as JSON

Since JSON, unlike hash maps, accepts only strings as keys, every other key type is stringified.

### DIFF
--- a/addons/hashes/fnc_encodeJSON.sqf
+++ b/addons/hashes/fnc_encodeJSON.sqf
@@ -16,6 +16,7 @@ Description:
     - STRING
     - TASK
     - TEAM_MEMBER
+    - HASHMAP
     - Everything else will simply be stringified.
 
 Parameters:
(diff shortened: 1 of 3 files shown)


commy2 (approved):

addons/hashes/fnc_encodeJSON.sqf
```diff
@@ -77,6 +78,19 @@ switch (typeName _object) do {
         };
     };
 
+    case "HASHMAP": {
+        private _json = ((_object toArray false) apply {
+            _x params ["_key", ["_value", objNull]];
+
+            if !(_key isEqualType "") then {
+                _key = str _key;
+            };
```
commy2, line 87: Why not `str` on STRINGs as well? You kind of lose the information of "false" vs false and "1" vs 1 with this now.

PR #1517, opened 2021-11-04
Item Context Menu -- Add class-specific actions support 

10Dozen:
- Rework Inventory actions to use HashMap to store actions.
- Add a way to make action apply to specific class only (without being inherited by child classes). 

**Note:**
Framework allows to apply actions(options) to item class, item type or "#All" wildcard. Actions added to class will be inherited by all it's child classes. 
Mods often do not have strict inheritance scheme (class_base (scope=0) -> various class_variants (scope=2)), but may inherit from classes with scope=2. If I want to add some specific action for parent item - it now inherited by all it's childs, which is sometimes unwanted.

This PR adds new way to specify `_item` parameter -- `!<classname>` -- which prevents actions for being inherited.

<details>
<summary>Example</summary>

Config:

```cpp
  class CUP_acc_ANPEQ_15: ItemBase;
  class CUP_acc_ANPEQ_15_Black: CUP_acc_ANPEQ_15;
```

Add action to `CUP_acc_ANPEQ_15` only:

```sqf
[
  "!CUP_acc_ANPEQ_15",
  ["VEST_CONTAINER"], 
  "Paint to black", 
  [], "", {true}, { _this call fnc_replaceItem }, true, 
  "CUP_acc_ANPEQ_15_Black"
] call CBA_fnc_addItemContextMenuOption;
```
</details>

### DIFF
--- a/addons/ui/fnc_addItemContextMenuOption.sqf
+++ b/addons/ui/fnc_addItemContextMenuOption.sqf
@@ -9,6 +9,9 @@ Parameters:
     _item                   - Item classname <STRING>
                               Can be base class.
 
+                              Can be specific class without being inherited by child classes:
+                                 "!MyMod_Class1"
+
                               Can be item type as reported by BIS_fnc_itemType:
                                 ["Equipment","Headgear"]
                                 ->
@@ -133,7 +136,7 @@ if (!hasInterface) exitWith {};
 
 // Initialize system on first execution.
 if (isNil QGVAR(ItemContextMenuOptions)) then {
-    GVAR(ItemContextMenuOptions) = false call CBA_fnc_createNamespace;
+    GVAR(ItemContextMenuOptions) = createHashMap;
 
     ["CAManBase", "InventoryOpened", {
         params ["_unit", "_container1", "_container2"];
@@ -222,11 +225,11 @@ _condition params [
     ["_conditionShow", {true}, [{}]]
 ];
 
-private _options = GVAR(ItemContextMenuOptions) getVariable _item;
+private _options = GVAR(ItemContextMenuOptions) get _item;
 
 if (isNil "_options") then {
     _options = [];
-    GVAR(ItemContextMenuOptions) setVariable [_item, _options];
+    GVAR(ItemContextMenuOptions) set [_item, _options];
 };
 
 _options pushBack [_slots, _displayName, _tooltip, _color, _icon, _conditionEnable, _conditionShow, _statement, _consume, _params];
--- a/addons/ui/fnc_openItemContextMenu.sqf
+++ b/addons/ui/fnc_openItemContextMenu.sqf
@@ -30,15 +30,19 @@ private _config = _item call CBA_fnc_getItemConfig;
 
 private _options = [];
 while {
-    _options append (GVAR(ItemContextMenuOptions) getVariable configName _config);
+    _options append (GVAR(ItemContextMenuOptions) get configName _config);
     _config = inheritsFrom _config;
     !isNull _config
 } do {};
 
+// Read unique class options (not inherited)
+_options append (GVAR(ItemContextMenuOptions) getOrDefault [format["!%1", _item], []]);
+
+// Read type and wildcard options
 _item call BIS_fnc_itemType params ["_itemType1", "_itemType2"];
-_options append (GVAR(ItemContextMenuOptions) getVariable [format ["##%1", _itemType2], []]);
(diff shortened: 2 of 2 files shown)


commy2:
Don't like the ! syntax. ! means not.
But really people should fix their inheritance.

10Dozen:
> Don't like the ! syntax. ! means not.

What about `$` ? Afaik it's not used in sqf
```sqf
["$CUP_acc_ANPEQ_15",["VEST_CONTAINER"], ...] call CBA_fnc_addItemContextMenuOption;
```

> But really people should fix their inheritance.

Impossible, imho. Some mods have it for years (and used in compats of other mods), it will break backward compatibility.

commy2:
Just use an explicit argument for the function.

10Dozen:
> Just use an explicit argument for the function.

Added new parameter

commy2:
[addClassEventHandler](https://github.com/CBATeam/CBA_A3/blob/92bf7f0a62955f8b2a8ed35bae60e2360e91bb00/addons/xeh/fnc_addClassEventHandler.sqf#L12) uses `_allowInheritance` with default `true`. That seems to fit better as var name.

10Dozen:
> addClassEventHandler uses _allowInheritance with default true. That seems to fit better as var name.

Heh, this was my first guess, but I thought that it should be a bit simplier for public function. Will make similar to what is used now.

> Still don't like the implementation with the prepended classname. Probably should be another hashmap for disallowed inheritance classes.

What will be the profit? Prepended classname way is just similar to handling type options  (`#<itemType1`/`##<itemType2>`) and adding more stuff into current map won't affect access time. The only thing that comes into my mind is to get rid of `format []` on access, but feels like type's options needs separate map more.

commy2:
> I thought that it should be a bit simplier for public function. Will make similar to what is used now.

Both are public functions.

> adding more stuff into current map won't affect access time.

Performance seems non-issue here. 

#(#) syntax is used for implicit typing. Think of them as virtual parents, used where the shitty trees made by BI fail us.

Prepending type names with arcane symbols is questionable. A `bcd` is not a `#bcd`, since they have different type names. Making the setting an explicit parameter better describes the behaviour.

*I would like if the code refected this as well.*

10Dozen:
Sounds reasonable, thanks for explanation.

10Dozen:
Hi, any chance to finish the review? 
I still want this feature as literally every mod inherits from scope=2 items sometimes (even ACE https://github.com/acemod/ACE3/blob/1180be3cdb1d5b92425bb4a8a1216bde51817968/addons/laserpointer/CfgWeapons.hpp#L18)

addons/ui/fnc_addItemContextMenuOption.sqf
```diff
@@ -228,11 +235,16 @@ _condition params [
...
+if (!_allowInheritance) then {
+    _contextMenuOptions = GVAR(ItemContextMenuUniqueOptions);
+};
+
+private _options = _contextMenuOptions get _item;
 
 if (isNil "_options") then {
-    _options = createHashMap;
-    GVAR(ItemContextMenuOptions) set [_item, _options];
+    _options = [];
```
BrettMayson:
```suggestion
    _options = createHashMap;
```

addons/ui/fnc_addItemContextMenuOption.sqf
```diff
@@ -122,11 +127,11 @@ Examples:
             params ["_unit", "_container", "_item", "_slot", "_params"];
             systemChat str [name _unit, typeOf _container, _item, _slot, _params];
             true
-        }, false, [0,1,2]] call CBA_fnc_addItemContextMenuOption;
+        }, false, [0,1,2], false] call CBA_fnc_addItemContextMenuOption;
```
BrettMayson:
```suggestion
        }, false, [0,1,2], "DebugAction", false] call CBA_fnc_addItemContextMenuOption;
```

addons/ui/fnc_openItemContextMenu.sqf
```diff
@@ -37,6 +37,10 @@ while {
     !isNull _config
 } do {};
 
+// Read unique class options (not inherited)
+_options merge (GVAR(ItemContextMenuUniqueOptions) getOrDefault [_item, []]);
```
BrettMayson:
```suggestion
_options merge (GVAR(ItemContextMenuUniqueOptions) getOrDefault [_item, createHashMap]);
```

PR #1519, opened 2021-11-08
New function: prettyFormat

7erra:
- Add the prettyFormat function to CBA's functions library


addons/arrays/fnc_prettyPrint.sqf
```diff
@@ -0,0 +1,67 @@
...
+Author:
+   Terra 
+---------------------------------------------------------------------------- */
+SCRIPT(prettyPrint);
+
+if (count _this == 0) exitWith {
+	"[]"
+};
+private _lines = ["["];
+// Recursive function calls have _depth already defined
```
commy2, line 39: This means that the function behaves differently if _depth already exists. This is very bad design wise. I recommend making _depth an explict function argument instead. Could use a local function if calling signature of this function should be kept, but also consider changing that, since passing arrays is dodgy too.

commy2:
Tabs must be replaced by 4 spaces each. I'd also say that this includes the pretty printed string.

addons/arrays/fnc_prettyPrint.sqf
```diff
@@ -0,0 +1,67 @@
...
+		// "[
+		//		0,
+		//		1,
+		//		[
+		//			""22"",
+		//			33,
+		//			[]
+		//		],
+		//		4
+		//	]"
```
commy2, line 27: Whitespace at end of line must be deleted.

addons/arrays/fnc_prettyPrint.sqf
```diff
@@ -0,0 +1,75 @@
...
+{
+    private _line = "";
+    for "_i" from 1 to (_depth + 1) do {
+        _line = _tab + _line;
+    };
+    _line = if (_x isEqualType []) then {
+        _line + ([_x, _tab, _depth + 1] call CBA_fnc_prettyPrint);
+    } else {
+        _line + str _x;
+    };
```
commy2, line 63:
```suggestion
    if (_x isEqualType []) then {
        _line = _line + ([_x, _tab, _depth + 1] call CBA_fnc_prettyPrint);
    } else {
        _line = _line + str _x;
    };
```

commy2:
About the name of this function.
Shouldn't it be prettyFormat? The `print` equivalent in SQF is either `systemChat` or `hint`. This function does not use either, but just returns a STRING.

addons/arrays/fnc_prettyPrint.sqf
```diff
@@ -0,0 +1,75 @@
...
+    if (_forEachIndex != count _array - 1) then {
```
Dystopian, line 64: `count _array` should be moved out of loop to variable to not execute it on each iteration. Maybe even `count _array - 1` but I'm not sure if it's worth it (the same is for `_depth + 1` above).

addons/arrays/fnc_prettyPrint.sqf
```diff
@@ -0,0 +1,75 @@
...
+
+Author:
+   Terra
+---------------------------------------------------------------------------- */
+SCRIPT(prettyPrint);
+params [["_array", [], [[]]], ["_tab", toString[9], [""]], ["_depth", 0, [0]]];
+if (count _array == 0) exitWith {
+    "[]"
+};
+private _lines = ["["];
```
Dystopian, line 52: looks like first line in output is always without indent even if depth > 0

addons/arrays/fnc_prettyPrint.sqf
7erra: Any reason for that change? My understanding of programming is the less you write the easier it is to change later.

addons/arrays/fnc_prettyPrint.sqf
7erra: I agree with `count _array - 1`, this can saved as a variable. Moving `_depth + 1` outside of the loop would add the need for more expressions since the variable is used before and after the loop. The performance gain would depend on the number of subarrays.

addons/arrays/fnc_prettyPrint.sqf
Dystopian: > 
> 
> the less you write the easier it is to change

not always. `_var = if ...` construction is less readable then what commy suggested and without performance hit.

addons/arrays/fnc_prettyFormat.sqf
```diff
@@ -0,0 +1,71 @@
...
+    _tabs + "[]"
+};
+private _lines = [_tabs + "["];
+private _lastIndex = count _array - 1;
+{
+    private _line = "";
+    if (_x isEqualType []) then {
+        _line = _line + ([_x, _tab, _depth + 1] call CBA_fnc_prettyPrint);
+    } else {
+        _line = _tab + _tabs + _line + str _x;
```
Dystopian, line 62:
```suggestion
        _line = [_x, _tab, _depth + 1] call CBA_fnc_prettyFormat;
    } else {
        _line = _tab + _tabs + str _x;
```

addons/arrays/fnc_prettyFormat.sqf
```diff
@@ -0,0 +1,71 @@
...
+private _tabs = [];
+_tabs resize _depth;
+_tabs = _tabs apply {_tab} joinString "";
+if (_array isEqualTo []) exitWith {
+    _tabs + "[]"
+};
+private _lines = [_tabs + "["];
+private _lastIndex = count _array - 1;
+{
+    private _line = "";
```
Dystopian, line 58: after changes we do have small performance hit with this useless assignment. @commy2 do you think it's worth it?

addons/arrays/fnc_prettyFormat.sqf
```diff
@@ -0,0 +1,71 @@
+#include "script_component.hpp"
+/* ----------------------------------------------------------------------------
+Function: CBA_fnc_prettyPrint
```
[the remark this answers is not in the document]
7erra: I actually already did that but for some reason did not commit

addons/arrays/fnc_prettyFormat.sqf
commy2: I don't get what this is going for. The conditional below does `_line = _line + `, but `_line` is always the empty string?

Seems to me like below it should just be `_line =`. Put `private "_line";` above the loop to not leak it.

addons/arrays/fnc_prettyFormat.sqf
```diff
@@ -0,0 +1,71 @@
...
+};
+private _lines = [_tabs + "["];
+private _lastIndex = count _array - 1;
+{
+    private _line = "";
+    if (_x isEqualType []) then {
+        _line = _line + ([_x, _tab, _depth + 1] call CBA_fnc_prettyFormat);
+    } else {
+        _line = _tab + _tabs + _line + str _x;
+    };
```
Dystopian, line 63:
```suggestion
    private "_line";
    if (_x isEqualType []) then {
        _line = [_x, _tab, _depth + 1] call CBA_fnc_prettyFormat;
    } else {
        _line = _tab + _tabs + str _x;
    };
```

7erra: Wouldn't this be a use case for using the return of the if statement as done in the original code?
```sqf
private _line = if (...
```

Dystopian: this is exactly why I asked commy above. I would run some tests on large arrays with subarrays (e.g. `configClasses` can return huge arrays) for both your and commy's cases and if the difference is not so big I prefer commy's one just for readability. If tests show significant difference your case is preferred.

7erra: From my tests there is an improvement by using the `_var = if ` version. Tested with this code:
```sqf
_array = getUnitLoadout (configFile >> "CfgVehicles" >> "B_Soldier_A_F");
diag_codePerformance [CBA_fnc_prettyFormat, [_array]];
```

addons/arrays/fnc_prettyFormat.sqf
```diff
@@ -0,0 +1,70 @@
...
+        _tab + _tabs + str _x;
+    };
+    if (_forEachIndex != _lastIndex) then {
+        _line = _line + ",";
+    };
+    _lines pushBack _line;
+} forEach _array;
+
+_lines pushBack (_tabs + "]");
+_lines joinString endl
```
commy2, line 70: Without having any opportunity to test this, isn't this just:
```sqf
private _lines = _array apply {
    if (_x isEqualType []) then {
        [_x, _tab, _depth + 1] call CBA_fnc_prettyFormat // return
    } else {
        _tab + _tabs + str _x // return
    };
};

_tabs + "[" + (_lines joinString (";" + endl)) + _tabs + "]" // return
```
Speaking of testing, this is a prime candidate for unit tests.

7erra: Indeed it is. Just instead of ";" it has to be ","

addons/arrays/fnc_prettyFormat.sqf
```diff
@@ -0,0 +1,65 @@
...
+_tabs resize _depth;
+_tabs = _tabs apply {_tab} joinString "";
+
+if (_array isEqualTo []) exitWith {
+    _tabs + "[]"
+};
+
+private _lines = _array apply {
+    if (_x isEqualType []) then {
+        [_x, _tab, _depth + 1] call CBA_fnc_prettyFormat;
```
commy2, line 59:
```suggestion
        [_x, _tab, _depth + 1] call CBA_fnc_prettyFormat
```
apply block returns this, should not use semi-colon to indicate return value.

addons/arrays/fnc_prettyFormat.sqf
```diff
@@ -0,0 +1,65 @@
...
+    } else {
+        _tab + _tabs + str _x;
```
commy2, line 61:
```suggestion
        _tab + _tabs + str _x
```
dito

commy2:
lgtm
Setting up _tabs every layer seems not very elegant though.
Needs unit test.

7erra:
Another idea is to add an option to set what linebreak char to use. currently it is endl but it would be useful to also allow "<br/>" for structured text.

commy2:
Or "\n" for `hint`, although I think endl should work there too?

addons/arrays/fnc_prettyFormat.sqf
```diff
@@ -0,0 +1,65 @@
...
+    (end)
+
+Author:
+   Terra, Dystopian, commy2
+---------------------------------------------------------------------------- */
+SCRIPT(prettyFormat);
+params [["_array", [], [[]]], ["_tab", toString [9], [""]], ["_depth", 0, [0]]];
+private _tabs = [];
+_tabs resize _depth;
+_tabs = _tabs apply {_tab} joinString "";
```
commy2, line 51: linter is dumb
```suggestion
_tabs = (_tabs apply {_tab}) joinString "";
```
to fix.

addons/arrays/fnc_prettyFormat.sqf
```diff
@@ -45,7 +45,7 @@ Author:
    Terra, Dystopian, commy2
 ---------------------------------------------------------------------------- */
 SCRIPT(prettyFormat);
-params [["_array", [], [[]]], ["_tab", toString [9], [""]], ["_depth", 0, [0]]];
+params [["_array", [], [[]]], ["_tab", toString [9], [""]], ["_lineBreak", endl, [""]], ["_depth", 0, [0]]];
```
commy2, line 48: needs doc in function header :)

commy2:
Should default indent really be tab? Seems unusual for Arma/SQF.

commy2:
That said `_tabs` -> `_indents`

PR #1520, opened 2021-11-10
diagnostic - target watch variable is broadcasted to all

### LINKED ISSUE #1521: Optimize CBA_targetWatch
Today we had an event with 56 players and we noticed that the TargetWatch variable generates unnecessary traffic. 
We have calculated that over 3 hours that would be 230 MB of traffic. After an optimization it would be only 4MB traffic.

To improve: https://github.com/CBATeam/CBA_A3/blob/master/addons/diagnostic/XEH_preInit.sqf#L48

Player A would like to know target debug variable, but CBA sends them to ALL

Also it would be optimal to set the interval of the query as a global CBA setting.

shukari:
- fix #1521
- this change will send the result only to the person how ask for it
- you can now set the refreshRate in the settings

### DIFF
--- a/addons/diagnostic/stringtable.xml
+++ b/addons/diagnostic/stringtable.xml
@@ -115,7 +115,7 @@
             <English>Refresh rate target watcher field</English>
         </Key>
         <Key ID="STR_CBA_Diagnostic_WatchInfoRefreshRateTooltip">
-            <English>Sets the refresh rate in seconds for the CBA target watcher fields to the right of the debug console</English>
+            <English>Refresh rate (in seconds) for the CBA target watcher fields to the right of the debug console.</English>
         </Key>
         <Key ID="STR_CBA_Diagnostic_ConsoleIndentSpaces">
             <English>4 Spaces</English>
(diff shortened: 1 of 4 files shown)


addons/diagnostic/stringtable.xml
```diff
@@ -111,6 +111,12 @@
             <French>Type d'indentation qui peut être employée dans la console de débogage.\nTab ajoute une indentation, et Shift + Tab en supprime une.</French>
             <Spanish>Tipo de sangría que se puede agregar a la expresión en la consola de depuración presionando la tecla Tab o eliminar presionando Shift + Tab</Spanish>
         </Key>
+        <Key ID="STR_CBA_Diagnostic_WatchInfoRefreshRate">
+            <English>Refresh rate target watcher field</English>
+        </Key>
+        <Key ID="STR_CBA_Diagnostic_WatchInfoRefreshRateTooltip">
+            <English>Sets the refresh rate in seconds for the CBA target watcher fields to the right of the debug console</English>
```
commy2, line 118:
```suggestion
            <English>Refresh rate (in seconds) for the CBA target watcher fields to the right of the debug console.</English>
```

commy2 (approved):
lgtm, maybe pabst double check

PabstMirror (approved):
LGTM

PabstMirror:
imho target debugging was really designed more for development and not "in-production"
but if it's getting used this way, there's no reason not to make it better

PR #1523, opened 2021-11-14
prevent switching accessory while in Zeus

ampersand38:
- title


commy2:
Seems sloppy. Like a proper fix would be needed anyway later. But w/e

commy2 (approved):

PR #1524, opened 2021-11-14
Update Mikero Tools to public version released 2021-11-06

Dahlgren:
- Update make.py to work both with previous and latest public tools
- Update GitHub Actions CI to use public Mikero Tools as of 2021-11-06


commy2:
What's broken in this version?

jonpas (approved):

PR #1527, opened 2021-11-27
Common - improve CBA_fnc_addMagazine and CBA_fnc_addMagazineCargo

Drofseh:
- improve CBA_fnc_addMagazineCargo by allowing the number of rounds to be specified.
- improve CBA_fnc_addMagazine by trying to add the magazine to vehicle _unit before throwing it out the window onto the ground, and also pass ammo count to CBA_fnc_addMagazineCargo, so that a partial mag doesn't magically become full
- requires #1528

### DIFF
--- a/addons/common/fnc_addMagazine.sqf
+++ b/addons/common/fnc_addMagazine.sqf
@@ -58,8 +58,8 @@ if (_verify) then {
 
         _return = true;
     } else {
-        private _vehicleUnit = vehicle _unit;
-        if (_vehicleUnit isEqualTo _unit) then {
+        private _vehicle = vehicle _unit;
+        if (_vehicle isEqualTo _unit) then {
             _unit switchMove "ainvpknlmstpslaywrfldnon_1";
 
             private _weaponHolder = nearestObject [_unit, "WeaponHolder"];
@@ -71,7 +71,7 @@ if (_verify) then {
 
             _return = [_weaponHolder, _item, 1, _verify, _ammo] call CBA_fnc_addMagazineCargo;
         } else {
-            _return = [_vehicleUnit, _item, 1, _verify, _ammo] call CBA_fnc_addMagazineCargo;
+            _return = [_vehicle, _item, 1, _verify, _ammo] call CBA_fnc_addMagazineCargo;
         };
     };
 } else {
--- a/addons/common/fnc_addMagazine.sqf
+++ b/addons/common/fnc_addMagazine.sqf
@@ -49,7 +49,7 @@ if (!isClass _config || {getNumber (_config >> "scope") < 2}) exitWith {
 };
 
 if (_verify) then {
-    if (_unit canAdd _item || {[_unit, _item, 1, true, true, true] call CBA_fnc_canAddItem}) then { // canAdd works for vehicles and units, CBA_fnc_canAddItem is more precise for units, but only try if canAdd fails for them
+    if ([_unit, _item, 1, true, true, true] call CBA_fnc_canAddItem) then {
         if (_ammo < 0) then {
             _unit addMagazine [_item, 1E6]; // addMagazine STRING is not synched when used on remote units. addMagazine ARRAY is.
         } else {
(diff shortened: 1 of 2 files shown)


addons/common/fnc_addMagazine.sqf
```diff
@@ -58,16 +58,21 @@ if (_verify) then {
...
+            if (isNull _weaponHolder || {_unit distance _weaponHolder > 2}) then {
+                _weaponHolder = createVehicle ["GroundWeaponHolder", [0,0,0], [], 0, "NONE"];
+                _weaponHolder setPosASL getPosASL _unit;
+            };
 
-        [_weaponHolder, _item] call CBA_fnc_addMagazineCargo;
+            _return = [_weaponHolder, _item, 1, _verify, _ammo] call CBA_fnc_addMagazineCargo;
+        } else {
+            _return = [_vehicleUnit, _item, 1, _verify, _ammo] call CBA_fnc_addMagazineCargo;
+        };
```
commy2, line 75:
```suggestion
        private _vehicle = vehicle _unit;
        if (_vehicle isEqualTo _unit) then {
            _unit switchMove "ainvpknlmstpslaywrfldnon_1";

        private _weaponHolder = nearestObject [_unit, "WeaponHolder"];
            private _weaponHolder = nearestObject [_unit, "WeaponHolder"];

        if (isNull _weaponHolder || {_unit distance _weaponHolder > 2}) then {
            _weaponHolder = createVehicle ["GroundWeaponHolder", [0,0,0], [], 0, "NONE"];
            _weaponHolder setPosASL getPosASL _unit;
        };
            if (isNull _weaponHolder || {_unit distance _weaponHolder > 2}) then {
                _weaponHolder = createVehicle ["GroundWeaponHolder", [0,0,0], [], 0, "NONE"];
                _weaponHolder setPosASL getPosASL _unit;
            };

        [_weaponHolder, _item] call CBA_fnc_addMagazineCargo;
            _return = [_weaponHolder, _item, 1, _verify, _ammo] call CBA_fnc_addMagazineCargo;
        } else {
            _return = [_vehicle, _item, 1, _verify, _ammo] call CBA_fnc_addMagazineCargo;
        };
```

addons/common/fnc_addMagazine.sqf
```diff
@@ -49,7 +49,7 @@ if (!isClass _config || {getNumber (_config >> "scope") < 2}) exitWith {
 };
 
 if (_verify) then {
-    if (_unit canAdd _item) then {
+    if (_unit canAdd _item || {[_unit, _item, 1, true, true, true] call CBA_fnc_canAddItem}) then { // canAdd works for vehicles and units, CBA_fnc_canAddItem is more precise for units, but only try if canAdd fails for them
```
commy2, line 52: Wouldn't inline a comment this long.

> CBA_fnc_canAddItem is more precise for units, but only try if canAdd fails for them

I have more questions now. What does `precise` mean in this context?

Drofseh: `canAdd` looks at total unit load and can return false for units when they do actually have space in a container.
`CBA_fnc_canAddItem` just checks each container directly and returns true if there is room.

I'd like to only use `CBA_fnc_canAddItem`, but `CBA_fnc_addMagazine` can take a vehicle as input, so we have to use `canAdd` as well for them.

commy2: Okay, so `canAdd` implies `canAddItem` for units.

Shouldn't we rewrite `canAddItem` to work on vehicles as well and drop `canAdd` entirely for this?

commy2:
- [ ] requires https://github.com/CBATeam/CBA_A3/pull/1528 to work

#1528 requires 2.08 commands. But perhaps I am missing something and it can be simplified to use [canAdd](https://community.bistudio.com/wiki/canAdd) instead? I am wondering why `CBA_fnc_canAddItem` does not use `canAdd` on the containers already.

PabstMirror (approved):
lgtm

PR #1533, opened 2022-02-16
Add CBA_fnc_weekDay

jonpas:
- Add a function that calculates date's day of the week using a modified Rata Die method with fractional years.
- Add unit tests for the function.

### DIFF
--- a/addons/common/CfgFunctions.hpp
+++ b/addons/common/CfgFunctions.hpp
@@ -114,6 +114,10 @@ class CfgFunctions {
             PATHTO_FNC(getNearestBuilding);
         };
 
+        class DateTime {
+            PATHTO_FNC(getWeekDay);
+        };
+
         class Misc {
             PATHTO_FNC(addPerFrameHandler);
             PATHTO_FNC(removePerFrameHandler);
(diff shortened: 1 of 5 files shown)


commy2:
CBA_fnc_weekDay

https://en.wikipedia.org/wiki/ISO_week_date

Should sunday be 0 or 7?
Is -1 a reasonable fallback?
Could the function or a variant thereof return strings, e.g. "MON"? Perhaps even (translated) stringtable entry?

jonpas:
Whichever.
Yes.
Sure.

jonpas:
For reference, `-1` will only return on invalid input. If we switch to 1-7 for days, invalid could also be 0.

PabstMirror (approved):

PabstMirror:
`[[2101,1,1]] call CBA_fnc_weekDay;` = 0
but https://www.timeanddate.com/calendar/?year=2101 says it should be a saturday
I think it's because 2100 isn't a leap year (even though it's divisible by 4)

hopefully arma 4 is out by then, so this shouldn't really be an issue 😄 (max mission date in arma is 2050)

PR #1538, opened 2022-03-23
Add magazine wells for GM CDLC

Dahlgren:
- Enable GM weapons to use magazines implementing CBA magazine wells
- Enable GM magazines to be used in other weapons implementing CBA magazine wells


Dahlgren:
Might be worth moving the magazine well definitions over into the conditional GM config to avoid cluttering config with invalid magazine definitions when GM isn't loaded

PabstMirror (approved):

addons/jam/jam_gm/config.cpp
```diff
@@ -1,8 +1,8 @@
 #include "script_component.hpp"
 
 class CfgPatches {
-    class ADDON {
-        name = ECSTRING(jam,component);
+    class SUBADDON  {
```
jonpas, line 4:
```suggestion
    class SUBADDON {
```

Issue #1544, opened 2022-04-26
SearchNearby Endless Loop


mrzachhigginsofficial:
**Mods (complete and add to the following information):**
- **Arma 3:** `x.xx` [e.g. 1.00 stable, rc,  dev]
- CBA: [fe3eba0](https://github.com/CBATeam/CBA_A3/commit/fe3eba001b0926129ba137a728fc64fdbc21f7fd)
_Make sure to reproduce the issue with only CBA on a newly created mission!_

**Description:**
CBA_fnc_SearchNearby is susceptible to an endless loop when pathing issues occur. In the excerpt below, you can see that the position is not removed until the unit is ready. If the unit does not reach its destination, it will never be ready. Therefore one unit in a group can cause this thread to stay open forever.
https://github.com/CBATeam/CBA_A3/blob/release/addons/ai/fnc_searchNearby.sqf

```
 while {_positions isNotEqualTo []} do {
        // Update units in case of death
        private _units = (units _group) - [_leader];

        // Abort search if the group has no units left
        if (_units isEqualTo []) exitWith {};

        // Send all available units to the next available position
        {
            if (_positions isEqualTo []) exitWith {};
            if (unitReady _x) then {
                private _pos = _positions deleteAt 0;
                _x commandMove _pos;
                sleep 2;
            };
        } forEach _units;
    };
```

In my opinion, there should be a failsafe in this loop (maybe a timeout).
```
private _starttime = time;
private _timeout = 240; //???

... stuff

while {_positions isNotEqualTo [] && __starttime + _timeout > time} do {
        // Update units in case of death
        private _units = (units _group) - [_leader];

        // Abort search if the group has no units left
        if (_units isEqualTo []) exitWith {};

        // Send all available units to the next available position
        {
            if (_positions isEqualTo []) exitWith {};
            if (unitReady _x) then {
                private _pos = _positions deleteAt 0;
                _x commandMove _pos;
                sleep 2;
            };
        } forEach _units;
    };
```

Not sure, just need some way to break this loop if there are pathing issues.

**Steps to reproduce:**
- Assign this task to group of 5 or more units. Eventually, they will get stuck.

**Expected behavior:**
- Units don't get stuck... forever.

**Where did the issue occur?**
- Self Hosted Multiplayer (probably everywhere)

**Log Files:**
- Not applicable. No error reported.

**Additional context:**
I will make a PR after work with my suggestion.

mrzachhigginsofficial:
[...]

PR #1545, opened 2022-04-27
SearchNearby Endless Loop #1544

mrzachhigginsofficial:
- Performance and reliability adjustments for SearchNearby function
- Set a max timeout for each unit move order (positions *15 and no more than 2 minutes).
- Set a time start variable "CBASearchTime" on each unit.
- Added a quarter-second pause between unit iterations.
- If the timeout is exceeded before the unit is ready, the unit will receive a new command for the next position. Command time is updated to reflect timeout for the next move command.

### DIFF
--- a/addons/ai/fnc_searchNearby.sqf
+++ b/addons/ai/fnc_searchNearby.sqf
@@ -20,7 +20,7 @@ Author:
 
 ---------------------------------------------------------------------------- */
 
-params ["_group"];
+params ["_group",["_timeoutcoef",-1],["_maxtimeout",-1]];
 
 _group = _group call CBA_fnc_getGroup;
 if !(local _group) exitWith {}; // Don't create waypoints on each machine
@@ -29,7 +29,7 @@ private _building = nearestBuilding (leader _group);
 if ((leader _group) distanceSqr _building > 250e3) exitwith {};
 
 [_group, _building] spawn {
-    params ["_group", "_building"];
+    params ["_group", "_building","_timeoutcoef","_maxtimeout"];
     private _leader = leader _group;
 
     // Add a waypoint to regroup after the search
@@ -52,8 +52,12 @@ if ((leader _group) distanceSqr _building > 250e3) exitwith {};
     private _positions = _building buildingPos -1;
 
     // Create limiters for each unit
-    private _timeout = (count _positions * 15) min 120;
-    private _timetag = "CBASearchTime";
+    private _timeout = if (_timeoutcoef > -1 && _maxtimeout > -1) then {(
+        count _positions * _timeoutcoef) min _maxtimeout
+    } else {
+        -1 
+    };
+    private _timetag = QGVAR(StartSearchTime);
     {_x setVariable [_timetag,time]} forEach units _group;
 
     while {_positions isNotEqualTo []} do {
@@ -65,10 +69,8 @@ if ((leader _group) distanceSqr _building > 250e3) exitwith {};
 
         // Send all available units to the next available position
         {
-            private _starttime = _x getVariable _timetag;
-
             if (_positions isEqualTo []) exitWith {};
-            if (unitReady _x || {_starttime + _timeout < time}) then {
+            if (unitReady _x || {if (_timeout > -1) then {(_x getVariable _timetag) + _timeout < time} else {false}) then {
                 private _pos = _positions deleteAt 0;
                 _x commandMove _pos;
                 _x setVariable [_timetag,time];
--- a/addons/ai/fnc_searchNearby.sqf
+++ b/addons/ai/fnc_searchNearby.sqf
@@ -51,7 +51,7 @@ if ((leader _group) distanceSqr _building > 250e3) exitwith {};
     // Search while there are still available positions
     private _positions = _building buildingPos -1;
 
-    // Create Limiters For Each Unit
+    // Create limiters for each unit
     private _timeout = (count(_positions) * 15) min 120;
     private _timetag = "CBASearchTime";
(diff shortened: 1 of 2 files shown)


mrzachhigginsofficial:
Proposed PR for issue #1544

addons/ai/fnc_searchNearby.sqf
```diff
@@ -59,13 +65,18 @@ if ((leader _group) distanceSqr _building > 250e3) exitwith {};
...
+                _x setVariable [_timetag,time];
                 sleep 2;
             };
+
+            sleep .25;
         } forEach _units;
     };
     _group lockWP false;
-};
+};
```
commy2, line 82: Newline at EOF is required.

commy2:
> Added a quarter-second pause between unit iterations.

Why?

> Set a time start variable "CBASearchTime" on each unit.

Should not be a hard coded string in  a local variable, but a QGVAR, i.e. QGVAR(SearchTimeout)

Not sure about using `time`. It uses the SQF NUMBER type single precision float, i.e. this change will make the script fail after roughly 11.5 days uptime.

addons/ai/fnc_searchNearby.sqf
```diff
@@ -50,6 +50,12 @@ if ((leader _group) distanceSqr _building > 250e3) exitwith {};
 
     // Search while there are still available positions
     private _positions = _building buildingPos -1;
+
+    // Create limiters for each unit
+    private _timeout = (count(_positions) * 15) min 120;
```
commy2, line 55: private _timeout = (count _positions * 15) min 120;

addons/ai/fnc_searchNearby.sqf
```diff
@@ -50,6 +50,12 @@ if ((leader _group) distanceSqr _building > 250e3) exitwith {};
...
+    private _timetag = "CBASearchTime";
+    {_x setVariable [_timetag,time]} forEach units _group;
```
commy2, line 57: Why not below use getVariable default instead?

addons/ai/fnc_searchNearby.sqf
```diff
@@ -59,13 +65,18 @@ if ((leader _group) distanceSqr _building > 250e3) exitwith {};
 
         // Send all available units to the next available position
         {
+            private _starttime = _x getVariable _timetag;
+
             if (_positions isEqualTo []) exitWith {};
-            if (unitReady _x) then {
+            if (unitReady _x || {_starttime + _timeout < time}) then {
```
commy2, line 71: If lazy eval is used here, may as well inline the whole local variable.

commy2:
Timeout should probably be a function parameter instead. Then it is opt in (default it would be off to keep the current behaviour), and tweakable.

addons/ai/fnc_searchNearby.sqf
mrzachhigginsofficial: The units would eventually skip every building as the start time would never be updated. Perhaps the variable name is confusing.

mrzachhigginsofficial:
> Timeout should probably be a function parameter instead. Then it is opt in (default it would be off to keep the current behaviour), and tweakable. I would not make it dependent on group size at all.

This is not dependent on group size but is dependent on the number of positions in the chosen building. Unit timeouts are evaluated the same whether there are 2 or 100.

```
private _positions = _building buildingPos -1;
...
private _timeout = (count _positions * 15) min 120;
```

addons/ai/fnc_searchNearby.sqf
mrzachhigginsofficial: Perhaps QGVAR(StartSearchTime) would make more sense.

mrzachhigginsofficial:
Regarding parameterizing the function, I agree. I suggest the following parameters:

_timeoutcoef
_maxtimeout

```
private _positions = _building buildingPos -1;
_timeout =  (count _positions * _timeoutcoef) min _maxtimeout;
```

addons/ai/fnc_searchNearby.sqf
mrzachhigginsofficial: As such?
```
if (unitReady _x || {(_x getVariable _timetag) + _timeout < time}) then {...}
```

mrzachhigginsofficial:
@commy2 - I need to test that latest commit some more. Please review and let me know if it resolves any of your concerns (especially the opt-out).

addons/ai/fnc_searchNearby.sqf
mrzachhigginsofficial: Updated in https://github.com/CBATeam/CBA_A3/pull/1545/commits/aa63e9a12138b64bc8b0826b5dc3fa8447c983ba

mrzachhigginsofficial:
I'm closing this PR for now. The more I dig into this, the more I see more opportunities for these units to get stuck.

Another example in **fnc_searchNearby.sqf**
```
    private _wp = _group addWaypoint [getPosASL _leader, -1, currentWaypoint _group];
    private _cond = "({unitReady _x || !(alive _x)} count thisList) == count thisList";
    private _comp = format ["this setFormation '%1'; this setBehaviour '%2'; deleteWaypoint [group this, currentWaypoint (group this)];", formation _group, behaviour _leader];
    _wp setWaypointStatements [_cond, _comp];
```

Once again, if the units are never ready, this "regroup" waypoint will never complete. I will open a new PR once I have a more comprehensive solution. @commy2 brings up a good point that this should be parameterized as to not change existing functionality, so I will keep that in mind. It is unfortunate so much reliance is put on the AI sim working as intended (especially since it's virtually impossible to account for every scenario).

PR #1554, opened 2022-09-13
WIP: Terrain library

Seb105:
## Terrain library

The basic idea is a wrapper library for setTerrainHeight which addresses some of its [shortcomings ](https://community.bistudio.com/wiki/setTerrainHeight)(network optimisations, savegame serialisation, ease of use).

### Network optimisation:

Split terrain heights up into `n*n` sized chunks of terrain cells, and expose a set of commands which update these chunks instead of whatever was passed verbatim.

The `CBA_fnc_setTerrainHeight` function uses the exact same syntax as the vanilla command, but under the hood will only ever modify terrain height in groups of `n*n` chunks aligned to a grid.

Any mod can use this command and know that the JIP queue isn't spammed any more than it needs to be.

### Savegames;

Chunks are uninitialsed by default, but when a function touches that chunk any changes are stored so they can be retrieved during savegame load. Therefore modders don't have to manage this themselves, which could be a source of issues if two mods try and modify the same terrain and save/load it themselves.

### Ease of use commands

Set terrain height is a bit unwieldy. I have created the addTerrainHeight as an example command for how life could be made easier by working in areas instead of groups of points

## TODO:

- [ ] Create chunking system
- [ ] Create savegame serialisation
- [ ] Create suite of helper functions that use areas instead of coordinates:
- - [ ] addTerrainHeight
- - [ ] flattenTerrainArea, 
- - [ ] restoreTerrainHeight
- - [ ] any other QOL stuff?
- [ ] Bugtest

**WIP proof of concept, all the commands added thus for do work but I'm not going to spend any more time unless this is something people are actually interested in**

### DIFF
--- a/addons/terrain/$PBOPREFIX$
+++ b/addons/terrain/$PBOPREFIX$
@@ -1,0 +1,1 @@
+x\cba\addons\terrain
--- a/addons/terrain/CfgEventHandlers.hpp
+++ b/addons/terrain/CfgEventHandlers.hpp
@@ -1,0 +1,5 @@
+class Extended_PreInit_EventHandlers {
+    class ADDON {
+        init = QUOTE(call COMPILE_SCRIPT(XEH_preInit));
+    };
+};
--- a/addons/terrain/CfgFunctions.hpp
+++ b/addons/terrain/CfgFunctions.hpp
@@ -1,0 +1,8 @@
+class CfgFunctions {
+    class CBA {
+        class Terrain {
+            PATHTO_FNC(nearestTerrainPoint);
+            PATHTO_FNC(setTerrainHeight);
+        };
+    };
+};
--- a/addons/terrain/CfgFunctions.hpp
+++ b/addons/terrain/CfgFunctions.hpp
@@ -1,8 +1,10 @@
 class CfgFunctions {
     class CBA {
         class Terrain {
+            PATHTO_FNC(addTerrainHeight);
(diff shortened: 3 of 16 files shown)


Drofseh:
It would be nice to have options for the smoothing, to allow slopes of different angles, or even curves

Seb105:
> It would be nice to have options for the smoothing, to allow slopes of different angles, or even curves

I'm not sure about different smoothing curves... Probably overcomplicating things. I could implement a threshold though, where 0 is no smoothing, 0.5 is 50% of the area being at _height and the rest being transitional, 1 is all smoothed.

Drofseh:
> > It would be nice to have options for the smoothing, to allow slopes of different angles, or even curves
> 
> I'm not sure about different smoothing curves... Probably overcomplicating things. I could implement a threshold though, where 0 is no smoothing, 0.5 is 50% of the area being at _height and the rest being transitional, 1 is all smoothed.

Different smoothing curves are basically required for different terrain features though. Smoothing for a rounded hill should not be the same as smoothing for a crater, or for a peaked ridge.

095677584a76251159d851d0ba31c472

Seb105:
I could do smoothIn, smoothOut, linear and smoothInOut and let you control the power as well, that lets you change everything you'd need to I guess

Seb105:
This is what I consider to be feature and complete and is ready for someone else to take a look.

PR #1562, opened 2022-11-24
Common - Improve CBA_fnc_addItem, Fix CBA_fnc_addMagazine return

Drofseh:

Improve CBA_fnc_addItem
If the unit doesn't have enough space then try to add the item to the unit's vehicle instead of always chucking it out the window onto the ground where it gets lost.

### DIFF
--- a/addons/common/fnc_addItem.sqf
+++ b/addons/common/fnc_addItem.sqf
@@ -11,7 +11,7 @@ Description:
 Parameters:
     _unit   - the unit <OBJECT>
     _item   - name of the weapon to add <STRING>
-    _verify - if true, then put item on the ground if it can't be added <BOOLEAN>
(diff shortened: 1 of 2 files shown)


commy2 (commented):
Changes rv from false to true when dropping items on the ground as far as I can tell.

Drofseh:
> Changes rv from false to true when dropping items on the ground as far as I can tell.

Should change it from always returning false to returning whatever CBA_fnc_addItemCargo returns if it is called, so true if it went somewhere basically.
Ok with to change that back if you'd prefer, but this is the same behaviour that CBA_fnc_addMagazine currently has

commy2:
Yes, because it breaks the behaviour documented in the functions header, and then also fix https://github.com/CBATeam/CBA_A3/pull/1527

Drofseh:
Done and done

jonpas (approved):

PR #1564, opened 2023-01-10
Settings - Fix decimal formatting not pushed to variable

Daniel-H123:
- Fixed bug that slider variable does not take in to account the decimal count


PabstMirror:
```
    private _editText = if (_isPercentage) then {
        format [localize "STR_3DEN_percentageUnit", round (_value * 100), "%"]
    } else {
        if (_trailingDecimals < 0) then {
            _value = round _value;
        };
        _value = [_value, 1, _trailingDecimals max 0] call CBA_fnc_formatNumber;
    };
```
I think `_editText` will be nil?
Has this been tested?

mharis001:
And `_value` will be a string and not number.

Daniel-H123:
Yeah you're right. I assumed it too fast. But the _value needs to be updated with the correct amount of decimals right?

Daniel-H123:
Maybe something like this:
```
private _editText = if (_isPercentage) then {
        format [localize "STR_3DEN_percentageUnit", round (_value * 100), "%"]
    } else {
        if (_trailingDecimals < 0) then {
            _value = round _value;
        };
        [_value, 1, _trailingDecimals max 0] call CBA_fnc_formatNumber;

       _value = round(_value * pow(10, _trailingDecimals)) / pow(10, _trailingDecimals);
    };
```

Daniel-H123:
Or better:
`_value = _value toFixed _trailingDecimals;`

Daniel-H123:
The "bug" is definitly there, but my solution was obviously not the right one (thx for pointing out 💯).
I think I fixed it now, but changed it to a draft. I will test this as soon as I can.

Thoughtthis would be an 'easy' fix 🤡

Daniel-H123:
This is now working code. Updated that decimal count is enforced in:
- Silder speed (when clicking inside slider)
- Step count (amount changed when arrow keys are pressed)
- All updates of the GUI

I tested it in-game and have tested the code for errors, but let me know if there are any problems.

PR #1568, opened 2023-02-14
Optics - Increase width of hidePeripheralVision slightly to prevent gaps

PabstMirror:
ref https://github.com/acemod/ACE3/issues/9141

I can reproduce at those exact resolution and scale/

mine has a gap on the left which is easy to see, but there also is a gap on the right at the edge of the image.

Checking the 3 `ctrlPosition`
Left: `[-0.918847,-0.563663,0.0890598,2.12733]`
Body: `[-0.829787,-1.27305,2.65957,3.5461]`
Right: `[1.82979,-0.563663,0.0890598,2.12733]`

Add up the widths and you get `2.83769` which is identical to the safezoneW
But I guess it's just a rounding error and the left ends a pixel short of where the body begins?
I think just adding half a pixel should be enough?

### DIFF
--- a/addons/optics/fnc_initDisplayInterrupt.sqf
+++ b/addons/optics/fnc_initDisplayInterrupt.sqf
@@ -46,7 +46,7 @@ private _ctrlBlackRight = _display ctrlCreate ["RscText", IDC_BLACK_RIGHT];
 private _width = THIRD_SCREEN_WIDTH;
 
 if (GVAR(hidePeripheralVision)) then {
-    _width = 0.5 - (_bodyPosition select 2)/2 - safezoneXAbs;
+    _width = 0.5 - (_bodyPosition select 2)/2 - safezoneXAbs + pixelW/2;
 };
 
 _ctrlBlackLeft ctrlSetBackgroundColor [0,0,0,1];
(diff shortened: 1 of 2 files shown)


mharis001 (approved):

commy2:
How would making it half a pixel wider result in the gap on the left be closed? Right sure, perhaps, if half pixels are even recognized, but left is controlled entirely by X value, which shouldn't change by changing W. Right?

PabstMirror:
```
Left: [-0.918847,-0.563663,0.0890598,2.12733]
Body: [-0.829787,-1.27305,2.65957,3.5461]
Right: [1.82979,-0.563663,0.0890598,2.12733]
```
Left X + Width: 
-0.918847 + 0.0890598 = -0.829787
Which looks like the same as Body X:
-0.829787

But might not be because of floating point?
Or the engine rounds down on the number of actual pixels to render?

This PR does fix the issue and shouldn't effect people at other resolutions.

commy2:
It would make more sense to decrease left/top by 1 pixel and increase width/height by 2 if this is an issue.

PabstMirror:
I would prefer to change the value as little as possible; just enough to fix the issue

veteran29 (approved):
> It would make more sense to decrease left/top by 1 pixel and increase width/height by 2 if this is an issue.

might be safer for all resolutions but if the current solution works too... LGTM

PR #1580, opened 2023-06-28
Fix JSON parsing function of numerals formatted as scientific notation with uppercase E

initramfs:
On both the [json.org](https://www.json.org) site and the relevant [ECMA-404](https://www.ecma-international.org/wp-content/uploads/ECMA-404_2nd_edition_december_2017.pdf) specification a `number` symbol is defined as containing an `exponent` symbol which may be: empty, "e" (U+0065, LATIN SMALL LETTER E), or "E" (U+0045, LATIN CAPITAL LETTER E).

The current parsing code will error on numbers using the capital "E" syntax such as `1.2E5`.

Implement the ability to parse numbers in scientific notation using capital "E" by adding the relevant token to the `_numeric` variable. The SQF script command `parseNumber` already handles numbers using uppercase "E" as an exponent delimiter without issue so no changes are required elsewhere (i.e. we do not need to casefold the lexed numeric token prior to feeding into the scripting engine).


jonpas:
@BaerMitUmlaut

BaerMitUmlaut:
Since we now have regex, this could be improved further for more edge cases.

initramfs:
@BaerMitUmlaut I can modify the tokenizer to use a regex to validate the number parsed, but what should it do if an invalid numeric token is found? The rest of the code seems to do no validation whatsoever and just assumes the JSON is valid (e.g [here](https://github.com/CBATeam/CBA_A3/blob/master/addons/hashes/fnc_parseJSON.sqf#L122)).

If the intent is for an invalid JSON document to crash the script is regex parsing necessary (since `parseNumber` would likely complain anyway).

BaerMitUmlaut:
Good point. Let's leave it like this (with your change) unless we add proper error handling.

BaerMitUmlaut (approved):

jonpas (approved):

PabstMirror (approved):

PR #1583, opened 2023-07-09
Optics - Fixed magnification text flickering at 1.0x zoom level

johnb432:
- Stops the magnification text in the top right corner from flickering when the magnification level is 1.0x.

---

```SQF
0.25 call CBA_fnc_getFov select 1
```
does not return a stable value (it oscillates lightly around 1).

An example scope is the `hlc_optic_atacr` from NIArms.

### DIFF
--- a/addons/optics/fnc_animateScriptedOptic.sqf
+++ b/addons/optics/fnc_animateScriptedOptic.sqf
@@ -75,7 +75,7 @@ private _zoom = 0.25 call CBA_fnc_getFov select 1;
 
 // To avoid flickering of the magnification text, anything above and including 0.995 is rounded to 1
 if (_zoom >= 0.995) then {
-	_zoom = 1 max _zoom;
+    _zoom = 1 max _zoom;
 };
 
 _ctrlMagnification ctrlSetText format [
--- a/addons/optics/fnc_animateScriptedOptic.sqf
+++ b/addons/optics/fnc_animateScriptedOptic.sqf
@@ -73,6 +73,11 @@ if (cameraView == "GUNNER") then {
 // Add magnification to zeroing control.
 private _zoom = 0.25 call CBA_fnc_getFov select 1;
 
+// To avoid flickering of the magnification text, anything above and including 0.995 is rounded to 1
+if (_zoom >= 0.995) then {
+	_zoom = 1 max _zoom;
+};
+
 _ctrlMagnification ctrlSetText format [
     "(%1x)",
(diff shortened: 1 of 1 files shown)


addons/optics/fnc_animateScriptedOptic.sqf
```diff
@@ -73,6 +73,11 @@ if (cameraView == "GUNNER") then {
 // Add magnification to zeroing control.
 private _zoom = 0.25 call CBA_fnc_getFov select 1;
 
+// To avoid flickering of the magnification text, anything above and including 0.995 is rounded to 1
+if (_zoom >= 0.995) then {
+	_zoom = 1 max _zoom;
```
PabstMirror, line 78:
```suggestion
    _zoom = 1 max _zoom;
```

johnb432: God damn it, I thought I had the .editorconfig settings set up... oh well.

PabstMirror (approved):

commy2:
Now it flickers even harder at 0.995 +-x

Instead, it should remember the last value and only update if delta > threshold.

PabstMirror:
are there any scopes with magnification less than 1?

johnb432:
> Now it flickers even harder at 0.995 +-x
> 
> Instead, it should remember the last value and only update if delta > threshold.

Good idea - any suggestions for a delta?

commy2:
If you had success with x=0.995, maybe dx=0.005. Doesn't have to be a constant either, can depend on magnification if that helps. Needs trial and error.

johnb432:
> are there any scopes with magnification less than 1?

[It doesn't cache magnification levels below 1](https://github.com/CBATeam/CBA_A3/blob/263286f95453697bfb296937cb1a896c7885e682/addons/optics/fnc_animateScriptedOptic.sqf#L133-L135), so I'm not sure if there are any/if it's even supported at all.

In [669eddd](https://github.com/CBATeam/CBA_A3/pull/1583/commits/669eddd97184af363acb5a0adfd392454d135e45) I moved that code. I'm not sure if it's a good or bad idea to cache magnification levels `>= 1` only. I tried without the `>= 1` limitation briefly with BWMod and NIArms scopes and it seemed to work, but the tests weren't thorough. With the limitation I tested a bit more thoroughly albeit not enough, I couldn't find any problems.

PabstMirror (approved):

PR #1584, opened 2023-07-10
Common - Added CBA_fnc_execAfterNFrames from ZEN

### LINKED ISSUE #1578: [Feature Request]: Add an extra parameter to CBA_fnc_execNextFrame for settable frame delay
**Description:**
I feel like having a CBA function that allows you to delay the execution of code by x frames would be welcome.

Suggestion:
Adding an extra parameter to `CBA_fnc_execNextFrame` could be a way of implementing it:
```SQF
[_func, _args, _numFrames] call CBA_fnc_execNextFrame
```

**Additional context:**
The current method to delay by x frames is to either nest `CBA_fnc_execNextFrame` within itself x times, which is hard to read.
Alternatively, people have made their own versions: See [ZEN](https://github.com/zen-mod/ZEN/blob/master/addons/common/functions/fnc_execAfterNFrames.sqf) for example.

johnb432:
- Adds ~~an extra parameter to `CBA_fnc_execNextFrame`~~ a new function `CBA_fnc_execAfterNFrames` that allows you to specify by how many frames you want the execution of the passed code to be delayed by. Closes #1578.

~~For robustness' sake I left the existing code of `CBA_fnc_execNextFrame` alone, which means if the delay is 1 frame, then it uses the old code.~~

During testing I noticed there is a bug, where if you call the function in a certain manner, it (EDIT: it = `CBA_fnc_execNextFrame`) will add an extra frame of waiting. This also happens in the current steam release however, so I'm not sure if it's worth trying to fix.
Here's the code I used to test this part specifically:

```SQF
private _delay = 1;

for "_index" from 1 to 10000 do {
    [{
        // If expected result, don't print result
        if (diag_frameno == ((_this select 0) + (_this select 1))) exitWith {};

        systemChat format ["%1, %2", diag_frameno, _this select 0];
    }, [diag_frameno, _delay], _delay] call CBA_fnc_execNextFrame;
	
    uiSleep 0.00005;
};
```

### DIFF
--- a/addons/common/fnc_execAfterNFrames.sqf
+++ b/addons/common/fnc_execAfterNFrames.sqf
@@ -1,0 +1,35 @@
+#include "script_component.hpp"
+/* ----------------------------------------------------------------------------
+Function: CBA_fnc_execAfterNFrames
+
+Description:
+    Executes the given code after the specified number of frames.
+
+Parameters:
+    _function - The function to run. <CODE>
+    _args     - Parameters passed to the function executing. This will be the same array every execution. (optional, default: []) <ANY>
+    _frames   - The amount of frames the execution of the function should be delayed by. (optional, default: 0) <NUMBER>
+
+Returns:
+    Nothing
+
+Examples:
+    (begin example)
+        [{hint "Done!"}, [], 5] call cba_fnc_execAfterNFrames;
+    (end)
+
+Author:
+    mharis001, donated from ZEN
+---------------------------------------------------------------------------- */
+
+if (canSuspend) exitWith {
+    [CBA_fnc_execAfterNFrames, _this] call CBA_fnc_directCall;
+};
+
+params [["_function", {}, [{}]], ["_args", []], ["_frames", 0, [0]]];
+
+if (_frames > 0) exitWith {
+    [CBA_fnc_execAfterNFrames, [_function, _args, _frames - 1]] call CBA_fnc_execNextFrame;
+};
+
+_args call _function;
--- a/addons/common/fnc_execAfterNFrames.sqf
+++ b/addons/common/fnc_execAfterNFrames.sqf
@@ -11,7 +11,7 @@ Parameters:
     _frames   - The amount of frames the execution of the function should be delayed by. (optional, default: 0) <NUMBER>
 
 Returns:
-    Nothing
+    Nothing Useful
 
 Examples:
     (begin example)
--- a/addons/common/fnc_execNextFrame.sqf
+++ b/addons/common/fnc_execNextFrame.sqf
@@ -3,11 +3,12 @@
 Function: CBA_fnc_execNextFrame
 
 Description:
-    Executes a code once in non sched environment on the next frame.
+    Executes a code once in non sched environment delayed by n frames.
 
 Parameters:
     _function - The function to run. <CODE>
-    _args     - Parameters passed to the function executing. This will be the same array every execution. [optional] <ANY>
+    _args     - Parameters passed to the function executing. This will be the same array every execution. (optional, default: []) <ANY>
+    _frames   - The number of frames the execution will be delayed by. (optional, default: 1) <NUMBER>
 
 Returns:
     Nothing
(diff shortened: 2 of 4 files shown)


addons/common/fnc_execNextFrame.sqf
```diff
@@ -21,9 +22,17 @@ Author:
...
 
-params [["_function", {}, [{}]], ["_args", []]];
+params [["_function", {}, [{}]], ["_args", []], ["_frames", 1, [0]]];
 
-if (diag_frameno != GVAR(nextFrameNo)) then {
+// Do not allow negative or 0 frame delays
+_frames = _frames max 1;
+
+if (_frames > 1) exitWith {
+    GVAR(nextFrameNBuffer) pushBack [diag_frameNo + _frames, _args, _function];
```
PabstMirror, line 31: I don't think this will work with save games
`diag_frameNo` will be completely different on load after restart

johnb432: Could add something like [this](https://github.com/CBATeam/CBA_A3/blob/c472830703a5c335bfa990c393a6b095d66d74bc/addons/common/init_perFrameHandler.sqf#L90) to fix that.

johnb432:
> imho this func has fairly rare use and I'd prefer not adding any additional overhead to the eachFrame block maybe we could just borrow https://github.com/zen-mod/ZEN/blob/master/addons/common/functions/fnc_execAfterNFrames.sqf

I could rewrite it in that format, however how would it work with the licensing/crediting?

mharis001:
> I could rewrite it in that format, however how would it work with the licensing/crediting?

ZEN license (GPLv3) is compatible with CBA (GPLv2). In either case, feel free to borrow and adapt as needed.

johnb432:
> > I could rewrite it in that format, however how would it work with the licensing/crediting?
> 
> ZEN license (GPLv3) is compatible with CBA (GPLv2). In either case, feel free to borrow and adapt as needed.

Thank you very much.

johnb432:
Ported the code from ZEN directly over.

veteran29 (approved):
LGTM

PabstMirror (approved):

addons/common/fnc_execAfterNFrames.sqf
```diff
@@ -0,0 +1,35 @@
...
+Description:
+    Executes the given code after the specified number of frames.
+
+Parameters:
+    _function - The function to run. <CODE>
+    _args     - Parameters passed to the function executing. This will be the same array every execution. (optional, default: []) <ANY>
+    _frames   - The amount of frames the execution of the function should be delayed by. (optional, default: 0) <NUMBER>
+
+Returns:
+    Nothing
```
commy2, line 14: It returns the RV of _function if _frames <= 0. Sorry, pedantry.

addons/common/fnc_execAfterNFrames.sqf
```diff
@@ -0,0 +1,35 @@
...
+    [CBA_fnc_execAfterNFrames, _this] call CBA_fnc_directCall;
+};
+
+params [["_function", {}, [{}]], ["_args", []], ["_frames", 0, [0]]];
+
+if (_frames > 0) exitWith {
+    [CBA_fnc_execAfterNFrames, [_function, _args, _frames - 1]] call CBA_fnc_execNextFrame;
+};
+
+_args call _function;
```
veteran29, line 35:
```suggestion
_args call _function;

nil
```

addons/common/fnc_execAfterNFrames.sqf
PabstMirror:
```suggestion
    Nothing Useful
```

addons/common/fnc_execAfterNFrames.sqf
commy2: Still returns a number if _frames > 0 (from execNextFrame)

veteran29: Exec next frame doc block is wrong too then.

johnb432: I can make them both return nil - is that what's wanted?

veteran29: I wouldn't bother due to performance, thinking about it ignore this suggestion too.

PR #1587, opened 2023-07-25
XEH - Add missing event handlers with SPE 1.0

Dahlgren:
- Fix XEH issues with Spearhead 1944


addons/xeh/CfgVehicles.hpp
```diff
@@ -546,4 +546,31 @@ class CfgVehicles {
     class Truck_02_aa_base_lxWS: Truck_02_base_F {
         XEH_ENABLED;
     };
+
+    // Spearhead 1944
+
+    class Tank;
+    class SPE_Armored_Target_Dummy: Tank {
+        class EventHandlers {
+            // This class has an empty CBA_Extended_EventHandlers, repopulate with actual event handlers
```
commy2, line 555: Wait, what? Why?

Dahlgren: Good question, this is the SPE class structure
```cpp
class SPE_Armored_Target_Dummy: Tank
{
	class Eventhandlers
	{
		class CBA_Extended_EventHandlers{};
	};
};
```
Since the `XEH_ENABLED` macro adds the `EventHandlers` classes with inheritance from the `CBA_Extended_EventHandlers_base` class it causes a load order issue and UBC problem. This structure with macros usage avoids both problems.

commy2: LOL. To fix, should put
```cpp
class SPE_Armored_Target_Dummy: Tank {
	class Eventhandlers {
		delete CBA_Extended_EventHandlers;
	};
};
```
here, and then
```cpp
    class Tank;
    class SPE_Armored_Target_Dummy: Tank {
        XEH_ENABLED;
    };
```
into [ee component](https://github.com/CBATeam/CBA_A3/blob/master/addons/ee/config.cpp).

Dahlgren: That would need a separate addon with cfgPatches `requiredAddons[] = { "Some_SPE_Addon" };` and either the new `skipWhenMissingDependencies` with 2.14 patch or `#if __has_include("some_spe_path")`

commy2: Why? `delete` complains now if no class is there? Maybe add it in patch 1, then remove it in patch 2, then XEH_enabled in patch 3. XEH already has a ton of (now sub-) module duplicates for legacy reasons. If you need to change the requiredAddons for it, go ahead.

Your current solution is whack.

PabstMirror: I think https://github.com/CBATeam/CBA_A3/pull/1589 should help with this problem

PabstMirror:
```
	class SPE_Armored_Target_Dummy: Tank
	{
		scope=1;
		class Eventhandlers
		{
			class CBA_Extended_EventHandlers
			{
			};
		};
```
it's scope=1 and nothing inherits from this

I'm guessing this is for something scripted,
they explicitly disabled eh and xeh on it
I think we might just want to trust them and not touch it

addons/xeh/compat_spe/config.cpp
```diff
@@ -0,0 +1,45 @@
...
+class CfgVehicles {
+    class Tank;
+    class SPE_Armored_Target_Dummy: Tank {
+        class EventHandlers {
+            // This class has an empty CBA_Extended_EventHandlers, repopulate with actual event handlers
+            class CBA_Extended_EventHandlers {
+                EXTENDED_EVENTHANDLERS
+            };
+        };
+        SLX_XEH_DISABLED = 0;
```
PabstMirror, line 29:
```suggestion
        SLX_XEH_DISABLED = 1;
```

Dahlgren:
This is likely to fixed in an future SPE update, should we skip this?

jonpas:
If it will be fixed, then yes. Can always merge after SPE updates if it ends up not being fixed.

jonpas:
Has this been fixed in SPE 1.02 released on September 21st?

jonpas:
@Dahlgren?

Dahlgren:
Latest patch fixed some but these are still broken,

* `SPE_MINE_TMI42_Field_30x30`
* `SPE_MINE_TMI42_Field_70x30`
* `SPE_MINE_US_M1A1_ATMINE_Field_30x30`

Config tree might be different than when this patch was maybe, it needs some love

kerckasha:
I fixed the minefields evnethandler inheritance on SPE side, should be fixed next time an update comes out

jonpas:
If those minefields were the last, I think we can close this @Drofseh ?

Dahlgren:
XEH issues fixed with SPE 1.1 patches

PR #1588, opened 2023-07-26
XEH - Add more events

LinkIsGrim:
- Add CargoLoaded, CargoUnloaded, HandleDamage, OpticsModeChanged, OpticsSwitch, RopeAttach, RopeBreak, SlotItemChanged (2.14), VisionModeChanged events support to XEH.
- Change event lists to alphabetical/wiki order.


addons/xeh/CfgEventHandlers.hpp
```diff
@@ -29,26 +29,33 @@ class Extended_InitPost_EventHandlers {};
...
+class Extended_GestureChanged_EventHandlers {};
+class Extended_GestureDone_EventHandlers {};
 class Extended_GetIn_EventHandlers {};
 class Extended_GetInMan_EventHandlers {};
 class Extended_GetOut_EventHandlers {};
 class Extended_GetOutMan_EventHandlers {};
-//class Extended_HandleDamage_EventHandlers {}; // Not implemented yet
-//class Extended_HandleHeal_EventHandlers {};   // Not implemented yet
+class Extended_HandleDamage_EventHandlers {};
+class Extended_HandleHeal_EventHandlers {};
```
commy2, line 58: Last time I checked, they ignored the return value. It was better to add handleDamage via init class event.

Does it work now? If not, don't enable these. It will only cause anger and confusion.

LinkIsGrim: Should `HandleHeal` be kept in `script_component.hpp` and `script_xeh.hpp`, or removed there as well?

commy2: Remove it.

If you want support for HandleDamage and HandleHeal in addClassEventHandler:

- make sure the RV is still ignored in current version
- if yes, make addClassEventHandler add a retroactive flag set init event handler that uses the addEventHandler command

Those would be removable by removeAllEventHandlers and removeEventHandler with hard-coded ehids, but anyone who uses the former is a terrorist and the latter goes against best practices so is on their own.

PabstMirror: I'm worried about adding HandleDamage now
it's going to take a ton of testing to make sure it doesn't cause problems with old mods that may have used this
and to make sure that the return value doesn't override existing code

LinkIsGrim:
I've commented out/removed all instances of `HandleHeal`/`HandleDamage`, though note that `HandleHeal` was previously available in some places.

PabstMirror (approved):

PR #1589, opened 2023-07-26
XEH - Move compats to individual addons

PabstMirror:
ref https://github.com/CBATeam/CBA_A3/pull/1587
get rid of werid `scope = "parseNumber` stuff
no added entries entries

it does add to mission.sqm (`addonRootClass` just doesn't work?)
```
addons[]={"objects_f_vietnam_c", "cba_ee_compat_sog" };
```
normally just adds `cba_xeh`, which isn't much different imho


BrettMayson:
Should this not be possible more cleanly with `skipWhenMissingDependencies`?

PabstMirror:
skipWhenMissingDependencies will be a drop-in thing

e.g. for [compat_contact/config.cpp](https://github.com/CBATeam/CBA_A3/pull/1589/files#diff-6bd821ad4c2961884bb448fcd9b16d0b4e8f1a9723e10f39582ebaa7895b4e09) just remove lines 5-12 and the last `endif`

BrettMayson:
I guess it depends on if there will be a CBA release before then

PabstMirror:
tested on 2.14 with `skipWhen` and everything works as expected
```
Skipped loading of addon 'cba_ee_compat_contact' as required addon 'A3_Data_F_Contact' is not present
Skipped loading of addon 'cba_ee_compat_csla' as required addon 'CSLA' is not present
Skipped loading of addon 'cba_ee_compat_sog' as required addon 'loadorder_f_vietnam' is not present
Skipped loading of addon 'cba_ee_compat_ws' as required addon 'data_f_lxWS' is not present
```
pbo check fails because tools are old, but updated inside this pr
hemmt check fails because it's still on 0.7.6

commy2:
Move ee folder back into xeh folder? 
https://github.com/CBATeam/CBA_A3/tree/master/addons/xeh

Next to xeh_a3, Extended_EventHandlers and CBA_Extended_EventHandlers.

Only reason it was moved out was tools problems: https://github.com/CBATeam/CBA_A3/pull/1199, but without scope being a config number expression, it's no longer needed to be separate.

ee subfolders could go directly into xeh.

ee has no special significance. It's there for bwc. All its former functionality has long since dumped into xeh. I only used that name for the nobin PBO, because it I didn't want to introduce another one that will eventually go out of use (once better alternative to scope exists, or the tool makers gave in to reason).

commy2:
Afaik, xeh/ee would clobber ee.

jonpas (approved):

PR #1591, opened 2023-07-30
QUOTE in version_config

BrettMayson:
- I've seen this trip up several people, is there any reason not to do this? It was just getting quoted by previous tools that HEMTT no longer does.


commy2:
It's a number, bro.

PR #1592, opened 2023-08-01
Strings - Cleanup

johnb432:
- Title.
- Minor optimizations.
EDIT:
- `CBA_fnc_capitalize` now fully supports Unicode strings.

All string tests passed successfully.

### DIFF
--- a/addons/strings/fnc_substr.sqf
+++ b/addons/strings/fnc_substr.sqf
@@ -30,7 +30,9 @@ SCRIPT(substr);
 params ["_string", "_startIndex", ["_length", 0]];
 
 // Check if _length is set else extract string to end
-if (_length <= 0) exitWith {_string select [_startIndex]};
+if (_length <= 0) exitWith {
+    _string select [_startIndex] // return
+};
 
 // Cut out string
 _string select [_startIndex, _length] // return


addons/strings/fnc_capitalize.sqf
```diff
@@ -25,16 +25,19 @@ Author:
 ---------------------------------------------------------------------------- */
 SCRIPT(capitalize);
 
-params ["_string"];
+forceUnicode 0;
```
commy2, line 28: side effect?

johnb432: The function should work properly now.

The old code would convert the first byte only, not guaranteeing if it was a char or just part of one (ANSI vs Unicode).

I understand if this change isn't wanted for backward compatibility reasons.

commy2: What I mean is, calling this function has the side effect of changing the unicode mode of the script it was called from.

johnb432: I should have looked at the wiki example a bit more in detail. I'll address that problem asap.

johnb432: I found a workaround. Now, calling `CBA_fnc_capitalize` does not affect the parent scope's unicode support at all.

PR #1607, opened 2023-09-21
Settings - Code readability improvements

LinkIsGrim:
- Change inline include to proper function.
- Add macros for getting controls from setting controlsGroup. Not all are covered, and not all instances of the code have been modified. Scope is limited to category creation.
- Change `switch` in category creation to string manipulation.

### DIFF
--- a/addons/settings/fnc_gui_addonChanged.sqf
+++ b/addons/settings/fnc_gui_addonChanged.sqf
@@ -21,7 +21,7 @@ uiNamespace setVariable [QGVAR(addonIndex), _index];
 private _selectedSource = uiNamespace getVariable QGVAR(source);
 
 if !(_display getVariable [_selectedAddon, false]) then {
-    #include "gui_createCategory.sqf"
+    [_display, _selectedAddon] call FUNC(gui_createCategory);
     _display setVariable [_selectedAddon, true];
 };
 
--- a/addons/settings/fnc_gui_createCategory.sqf
+++ b/addons/settings/fnc_gui_createCategory.sqf
@@ -1,4 +1,6 @@
-// inline function, don't include script_component.hpp
+#include "script_component.hpp"
+
+params ["_display", "_selectedAddon"];
 
 private _fnc_controlSetTablePosY = {
     params ["_control", "_tablePosY", "_height"];
@@ -102,7 +104,7 @@ private _lastSubCategory = "$START";
         // Add sub-category header
         if (_createHeader) then {
             private _ctrlHeaderGroup = _display ctrlCreate [QGVAR(subCat), -1, _ctrlOptionsGroup];
-            private _ctrlHeaderName = _ctrlHeaderGroup controlsGroupCtrl IDC_SETTING_NAME;
+            private _ctrlHeaderName = GET_CTRL_NAME(_ctrlHeaderGroup);
             _ctrlHeaderName ctrlSetText format ["%1:", _subCategory];
 
             private _tablePosY = (_ctrlOptionsGroup getVariable [QGVAR(tablePosY), TABLE_LINE_SPACING/2]);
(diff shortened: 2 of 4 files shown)


addons/settings/fnc_gui_createCategory.sqf
```diff
@@ -111,27 +113,11 @@ private _lastSubCategory = "$START";
...
-            };
-            case "COLOR": {
-                _display ctrlCreate [[QGVAR(Row_Color), QGVAR(Row_ColorAlpha)] select (count _defaultValue > 3), IDC_SETTING_CONTROLS_GROUP, _ctrlOptionsGroup]
-            };
-            case "TIME": {
-                _display ctrlCreate [QGVAR(Row_Time), IDC_SETTING_CONTROLS_GROUP, _ctrlOptionsGroup]
-            };
-            default {controlNull};
+        private _settingTypeTemp = _settingType;
+        if (_settingTypeTemp == "COLOR" && {count _defaultValue > 3}) then {
```
rautamiekka, line 117:
```suggestion
        if (_settingTypeTemp isEqualTo "COLOR" && {count _defaultValue > 3}) then {
```

rautamiekka (changes requested):

addons/settings/fnc_gui_createCategory.sqf
jonpas: Again this, why?

rautamiekka: Everything stacks. If neither the intended function or the performance doesn't change for the worse, then nothing is lost; small gains, no matter how small, are always welcome.

LinkIsGrim: Need to make sure `_settingTypeTemp` is always uppercase then. But my concern with this wasn't performance. Just reducing lines/making them easier on the eyes.

jonpas: There is one loss, readability. I don't see any gains, even small onesy though.

rautamiekka: Hardly any readability loss.

If the casing is questionable, then there should be a comment, and wherever the problem source for the questionable casing is needs to be fixed.

addons/settings/fnc_gui_createCategory.sqf
```diff
@@ -186,7 +172,7 @@ private _lastSubCategory = "$START";
         };
 
         // ----- set setting name
-        private _ctrlSettingName = _ctrlSettingGroup controlsGroupCtrl IDC_SETTING_NAME;
+        private _ctrlSettingName = GET_CTRL_NAME(_ctrlSettingGroup);
```
commy2, line 175: I do not like these. Seems like pointless indirection. The statement returning a "ctrl name" is already document by the identifier of the assignment.

addons/settings/fnc_gui_createCategory.sqf
commy2: It also loses robustness, because IET silently returns false for mismatched types, when it really should blow up instead.

addons/settings/fnc_gui_addonChanged.sqf
```diff
@@ -21,7 +21,7 @@ uiNamespace setVariable [QGVAR(addonIndex), _index];
 private _selectedSource = uiNamespace getVariable QGVAR(source);
 
 if !(_display getVariable [_selectedAddon, false]) then {
-    #include "gui_createCategory.sqf"
+    [_display, _selectedAddon] call FUNC(gui_createCategory);
```
commy2, line 24: Good change. This was always just a troll.

addons/settings/fnc_gui_createCategory.sqf
rautamiekka: I suppose that's a fair point.

addons/settings/fnc_gui_createCategory.sqf
LinkIsGrim: It was for the sake of reducing text, I found it to be an eyesore.

addons/settings/fnc_gui_addonChanged.sqf
LinkIsGrim: /shrug

PR #1609, opened 2023-09-23
Tools - Fix configs for HEMTT v1.9.0

PabstMirror:


jonpas (commented):
I don't like changing includes at all. @BrettMayson Any chance we can ignore duplicate defines in `include` folder?

jonpas (approved):

PR #1612, opened 2023-09-28
Joint Rails - Fix `_typefilter` undefined variable error

johnb432:
- Title.
- I unfortunately do not have access to Arma at this time, so I would be grateful if somebody could test this for me.


PabstMirror (approved):
tested working with
```
[] spawn { x = [currentWeapon player] call CBA_fnc_compatibleItems };
```

Mike-MF (approved):
Tested it with my original usecase for the issue, it works.

Issue #1613, opened 2023-09-28
Frame related functions dont work in Eden Editor


Eathox:
**Mods (complete and add to the following information):**
- **Arma 3:** `2.14.150957` stable
- **CBA:** `3.15.8` stable

**Description:**
Frame related functions dont work in Eden editor 

**Steps to reproduce:**
- Go to the Eden Editor (don't matter if SP or MP)
- Watch the variable `cba_common_nextFrameNo`
- Witness how it never advances

**Expected behavior:**
`cba_common_nextFrameNo` should update in the editor to allow for usage of frame related functions.

**Where did the issue occur?**
- Editor (Singleplayer) / Editor (Multiplayer)

**Log Files:**
- https://gist.github.com/Eathox/e829744f62e1acf3e6dc019ad1c12adc?ts=4

**Additional context:**
The issue occurs because XEH_postInit never happens in 3DEN ,see: https://github.com/CBATeam/CBA_A3/blob/master/addons/xeh/fnc_initDisplay3DEN.sqf. This also begs the question if there's any other functionality potentially not working correctly in 3DEN?

PR #1616, opened 2023-10-01
enable frame related functions in 3DEN

Eathox:
- Fixes: #1613 

I'm not sure about the location, I imagine it might be nicer to have somewhere in the common component but I wouldn't know where.

### DIFF
--- a/addons/xeh/fnc_initDisplay3DEN.sqf
+++ b/addons/xeh/fnc_initDisplay3DEN.sqf
@@ -6,6 +6,9 @@ private _fnc_watchDog = {
     if (!ISPROCESSED(missionNamespace)) then {
         INFO_1("missionNamespace processed [%1]", ISPROCESSED(missionNamespace));
         [] call CBA_fnc_preInit;
+
+        // Enable frame related functions (normally enabled in postInit)
+        addMissionEventHandler ["EachFrame", {call EFUNC(common,onFrame)}];
     };
 };
 


addons/xeh/fnc_initDisplay3DEN.sqf
```diff
@@ -6,6 +6,9 @@ private _fnc_watchDog = {
     if (!ISPROCESSED(missionNamespace)) then {
         INFO_1("missionNamespace processed [%1]", ISPROCESSED(missionNamespace));
         [] call CBA_fnc_preInit;
+
+        // Enable frame related functions (normally enabled in postInit)
```
rautamiekka, line 10:
```suggestion
        // Enable frame-related functions (normally enabled in postInit)
```

rautamiekka (changes requested):
Grammar fix.

addons/xeh/fnc_initDisplay3DEN.sqf
Eathox: Thanks, I have dyslexia so this is one to keep an eye out for in my PRs :P

jonpas:
@CBATeam/developers Anyone foresee any issue with this?

PabstMirror:
From discord:
> So merging this PR would break the editor when BWA3 is loaded and the "player" character is equipped with a scripted optic like ZO4x30.

I'll try to actually test this

PR #1617, opened 2023-10-12
Settings - Add "volatile" option to not persist server settings after restart

veteran29:
- Adds support for `cba_settings_volatile` root config property
- When set to `1` it will load/store server settings from `uiNamespace` instead of `profileNamespace` effectively making them persist only to the game restart


veteran29:
image

addons/settings/stringtable.xml
```diff
@@ -315,6 +315,14 @@
             <Czech>Resetovat na původní hodnotu.</Czech>
             <Korean>기본값으로 초기화합니다.</Korean>
         </Key>
+        <Key ID="STR_cba_settings_volatile">
+            <English>Changes will not persist after server restart</English>
+            <Polish>Zmiany nie zostaną zachowane po restarcie serwera</Polish>
+        </Key>
+        <Key ID="STR_cba_settings_volatile_tooltip">
+            <English>The server is configured to discard all changes to its setting after it will be restarted.\nChanges will persist between mission restarts.</English>
```
BrettMayson, line 323:
```suggestion
            <English>The server is configured to not persist changes to settings between server restarts.\nChanges will persist between mission restarts.</English>
```

.hemtt/project.toml
```diff
@@ -35,3 +36,5 @@ preset = "Hemtt"
 
 [hemtt.release]
 folder = "CBA_A3"
+
+[hemtt.launch.default]
```
BrettMayson, line 40:
```suggestion
```
fixed in 1.9.0

BrettMayson (changes requested):

BrettMayson (approved):

addons/settings/fnc_init.sqf
```diff
@@ -166,7 +166,7 @@ GVAR(default) setVariable [_setting, [_defaultValue, _setting, _settingType, _se
 private _settingInfo = GVAR(userconfig) getVariable _setting;
 
 if (isNil "_settingInfo") then {
-    private _settingsHash = profileNamespace getVariable [QGVAR(hash), HASH_NULL];
+    private _settingsHash = GVAR(serverHashNamespace) getVariable [QGVAR(hash), HASH_NULL];
```
PabstMirror, line 169: Not sure, but won't this cause clients to use the wrong namespace?

addons/settings/fnc_initDisplayMain.sqf
```diff
@@ -28,6 +28,12 @@ if (_file != "") then {
 
 uiNamespace setVariable [QGVAR(userconfig), compileFinal str _userconfig];
 
+private _volatile = getNumber (configFile >> QGVAR(volatile)) == 1;
```
PabstMirror, line 31:
```suggestion
private _volatile = (getNumber (configFile >> QGVAR(volatile)) == 1) && {!hasInterface};
```
also not sure, but I feel like this is the intent

addons/settings/fnc_init.sqf
veteran29: We might enable the feature only on servers as you've suggested below: https://github.com/CBATeam/CBA_A3/pull/1617#discussion_r1360016023

addons/settings/fnc_initDisplayMain.sqf
veteran29: After giving it a bit of thought I guess we might want to have it run only on dedicated servers. HCs don't fiddle with settings so this should be good.

PabstMirror (approved):

PabstMirror (changes requested):
actually this breaks load/save
```
Documents\Arma 3 - Other Profiles\Pabst%20Mirror\UserSaved\missions\testSave.Stratis\save.Arma3Save/GameState/Item2724/data.: Value type not matching: Namespace != Nothing
Error loading/saving variable cba_settings_serverhashnamespace
Error 'Bad file structure' at 'Documents\Arma 3 - Other Profiles\Pabst%20Mirror\UserSaved\missions\testSave.Stratis\save.Arma3Save/GameState/Item2724/data.'.
ErrorMessage: Cannot load 'Documents\Arma 3 - Other Profiles\Pabst%20Mirror\UserSaved\missions\testSave.Stratis\save.Arma3Save'.
Application terminated intentionally
ErrorMessage: Cannot load 'Documents\Arma 3 - Other Profiles\Pabst%20Mirror\UserSaved\missions\testSave.Stratis\save.Arma3Save'.
```

addons/settings/initSettings.sqf
```diff
@@ -10,10 +12,13 @@ if (isNil QGVAR(default)) then {
 
     if (isNil QGVAR(server)) then {
         GVAR(server) = NAMESPACE_NULL;
+        GVAR(serverHashNamespace) = [profileNamespace, uiNamespace] select _volatile;
```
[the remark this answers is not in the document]
veteran29: What's unsafe here? This is missionNamespace var. If somebody can mess with it they can mess with anything, can't they? I'm not against doing it as you've suggested.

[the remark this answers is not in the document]
PabstMirror: with no mods, just doing 
`x = profileNamespace` and save/load will crash the game

addons/settings/fnc_init.sqf
```diff
@@ -166,7 +166,7 @@ GVAR(default) setVariable [_setting, [_defaultValue, _setting, _settingType, _se
 private _settingInfo = GVAR(userconfig) getVariable _setting;
 
 if (isNil "_settingInfo") then {
-    private _settingsHash = profileNamespace getVariable [QGVAR(hash), HASH_NULL];
+    private _settingsHash = GET_SERVER_NAMESPACE getVariable [QGVAR(hash), HASH_NULL];
```
PabstMirror, line 169: clients run this as well, don't want them to use the GVAR(volatile)

addons/settings/initSettings.sqf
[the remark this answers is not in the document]
veteran29: Fixed in [686fb17](https://github.com/CBATeam/CBA_A3/pull/1617/commits/686fb174f2ec3572c5757d88e26e8bee02cc724f)

addons/settings/fnc_init.sqf
veteran29: Fixed in [18f785d](https://github.com/CBATeam/CBA_A3/pull/1617/commits/18f785d1470f28e1013daa50c7e0035cc2b53002). Only dedicated will use `uiNamespace`

veteran29, line 169: As this runs on client I think the macro could use some better name than `GET_SERVER_NAMESPACE` but not sure what name.

veteran29:
Crashes fixed by not storing namespaces in variables.

PabstMirror (approved):

jonpas (approved):

PR #1618, opened 2023-10-15
Arrays - Fix number bounds (ASC compatibility)

PabstMirror:
Number too big to be valid for ASC


BrettMayson (approved):

jonpas (approved):

veteran29:
Shouldn't ASC handle those as "INF" same as game does?

Issue #1620, opened 2023-10-19
Some CBA Settings fail to import


tricona3:
**Mods (complete and add to the following information):**
- **Arma 3:** `2.14`
- **CBA:** `CBA_A3 v3.16.0.231017`
- Various other mods (ACE3, TFAR Beta, CUP Weapons/Units/Vehicles, multiple other mods including private ones.
_Make sure to reproduce the issue with only CBA on a newly created mission!_

**Description:**
Some CBA settings now fail to import, giving the error in the RPT that they don't exist, despite them definitely existing. This wasn't a problem until this CBA update, and now it is doing it across various mods, including ones that haven't updated in years (so I'm ruling out them being the issue)

The issue is happening when either importing from a saved preset via the load button, or when pasting in the settings via the import button.

If I put the settings into a cba_settings.sqf file in the mission directory it does work however. But even if I then try to save that as a preset it then won't work when importing it to a fresh mission.

**Steps to reproduce:**
- Make a new mission in Eden
- Attempt to import CBA Settings from a saved preset
- Does not import some settings, gives error in RPT that they don't exist

**Expected behavior:**
I would expect all settings to import as they did previously

**Where did the issue occur?**
- Editor (Singleplayer)

**Log Files:**
- [RPT file](https://pastebin.com/aWdT0MWY)
- [The preset I am attempting to import](https://pastebin.com/1QkVrqDa)

**Additional context:**
It's really strange how it seems to be random settings that are doing this, some of these mods haven't updated in a long time so there's no reason they should have stopped working. As mentioned previously it was fine before the CBA update.

**Screenshots:**
![example1](https://github.com/CBATeam/CBA_A3/assets/108828234/52b305fa-099e-41f1-ada9-8d8460dd7306)
The screenshot just shows one of the example settings, just to prove that it definitely does exist. This is taken after attempting to import the settings, you can see it hasn't forced the setting despite that being in the preset.

johnb432:
This reason why it's throwing the errors is because it considers the category comments as part of the first setting's name.
E.g. this is considered to be a setting name:
```
Setting // ACE Advanced Ballistics
force ace_advanced_ballistics_ammoTemperatureEnabled
```

The `cba_settings.sqf` file is preprocessed, which means all of the comments are deleted before it's processed. However, when importing, this does not happen and so the comments are part of the processing - something that is not intended.

[...]

Issue #1619, opened 2023-10-19
Latest CBA version does not honor white space in the CBA settings Configuration File for Task Force Arrowhead Radio (BETA)


EagleTrooper91:
**Mods (complete and add to the following information):**
- **Arma 3:** `2.14` 
- **CBA:** `CBA_A3 v3.16.0.231017`
_Make sure to reproduce the issue with only CBA on a newly created mission!_

**Description:**
After the recent update to CBA, CBA settings for Task Force Radio Arrowhead (Beta) version 1.0.334 no longer is honored if the CBA settings for TFAR contains spaces. I believe it is related to #1581 which I believe is stripping white space to support multi line.

**Steps to reproduce:**
 - Create a Teamspeak Channel for Task Force Arrowhead Radio with Spaces in the name such as `Main Operation Server`
 - Under the CBA settings configure the TFAR settings... `TFAR_Teamspeak_Channel_Name = "Main Operations Server"`
 - Join a server.
 - You will not get pulled in to the channel but instead stay in the channel you were initially
 - Review the Server Settings and you notice that the Spaces are not honored. You must Change the Teamspeak Channel to have no spaces.
 - Edit the Teamspeak Channel to Match the CBA Settings but without spaces. 
 - You will be instantly moved into the matching channel (Ignoring Spaces)

**Expected behavior:**
 - Be pulled in to the matching Task Force Teamspeak Channel that is listed in the CBA settings Configuration under user config.

**Where did the issue occur?**
- Dedicated & Self-Hosted Multiplayer

**Log Files:**
https://gist.github.com/EagleTrooper91/597f20d60c8144271c7e225263ab42c1

**Additional context:**
This was working just prior to the October 17th 2023 update.

![image](https://github.com/CBATeam/CBA_A3/assets/73040843/a5b11b0b-30f8-47e5-b7a8-2f84ce6f31ee)

![image](https://github.com/CBATeam/CBA_A3/assets/73040843/93d5295d-6e7a-484e-b457-072ecde7f538)

![image](https://github.com/CBATeam/CBA_A3/assets/73040843/790c73a9-73c7-45a3-9fdb-181cc61dde2f)

johnb432:
https://discord.com/channels/976165959041679380/976228110078992456/1164261712371916830

I'm searching for a fix.

PR #1622, opened 2023-10-21
Settings - Fixes for whitespace and comments and unit tests

johnb432:
- Title. More specifically it should address #1619 and #1620.
- **Comments fix**: From line 29 through 34.
  - If the order of the `regexReplace` commands is switched (so removing multiline comments first, then single line), it will not work. This makes be believe both my regex' are not robust enough.
  - There are cases I have thought of that could be problematic, but I'm not sure how to test, because the .editorconfig file restricts them. E.g. having a single line comment be at the end of a file, but not having a newline (is that even possible?) or using different newline encoding styles (LF, CR and CRLF).
- **Whitespace fix**: Removal of lines 25 through 29. The comment was mostly wrong, except in one instance:

>  If in single quotes, it removed the whitespaces (this is the old code):
'  T E S T  ' becomes "TEST"

- More info can be found here: https://discord.com/channels/976165959041679380/976228110078992456/1164706891952300152 (CBA discord, a thread I made concerning the findings relating to whitespace problem).

I'd appreciate any help.


addons/settings/test_parse.sqf
```diff
@@ -0,0 +1,94 @@
...
+	["test5", "[ '  t e s t  ' , ""  T E S T  "" ]", 0],
+	["test6", "[ ""  t e s t  "" ,
+
+""""  T E S T  """" ]", 0],
+	["test7", "[ true, false ]", 0],
+	["test8", "[ ""  item_1  "" , ""  item_2  "" ]", 0],
+	["test9", "[ '  item_1  ' , '  item_2  ' ]", 0],
+	["test10", "[ '  item_1  ' , ""  item_2  "" ]", 0],
+	["test11", "[ '""  item_1  ""' , '  item_2  ' ]", 0],
+	["test12", "[ '  item_1  ' , ""'  item_2  '"" ]", 0]
```
jonpas, line 73: Tabs!

johnb432: > Tabs!

Tabs inside or outside of the strings? I saw the github action complained about tabs, but I want the tests to run with tabs (to make sure they work - unless that's unnecessary?).

jonpas: Oh right, I see now. @PabstMirror do we have a way to ignore validating parts of code?

I don't think we _need_ to test for tabs though, as whitespace is whitespace to Arma, but I may be wrong.

PabstMirror: passing CI now

johnb432:
Something I forgot to mention explicitly for the whitespace fix regarding single quoted strings:
They will no longer have their whitespace removed. Plus, they will be converted into double quoted strings.
 So `' T E S T '` will remain/become `" T E S T "`.
 
 I'm going to re-test it though, to be 100 % certain, as the tests above were affected with Advanced Developer Tool.s

jonpas (approved):

PR #1624, opened 2023-10-28
Configs - Cleanup for hemtt macro padding check

PabstMirror:
not touching flexi-menu

### DIFF
--- a/include/a3/ui_f/hpp/defineDIKCodes.inc
+++ b/include/a3/ui_f/hpp/defineDIKCodes.inc
@@ -1,6 +1,3 @@
-#pragma hemtt suppress pw3_padded_arg file
-#pragma hemtt suppress pw3_padded_arg config
-
 #ifndef DIK_ESCAPE
 
 /****************************************************************************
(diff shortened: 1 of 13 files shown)


include/a3/ui_f/hpp/defineDIKCodes.inc
```diff
@@ -1,3 +1,6 @@
+#pragma hemtt suppress pw3_padded_arg file
+#pragma hemtt suppress pw3_padded_arg config
+
```
PabstMirror, line 3: @BrettMayson - still getting with either/both line, not sure if error or just how the includes work
```
[PW3] Warning: padding a macro argument
     ╭─[/a3/ui_f/hpp/defineDIKCodes.inc:119:36]
     │
 119 │ #define DIK_NUMPADENTER     0x9C    /* Enter on numeric keypad */
     │                                    ┬
     │                                    ╰── padding a macro argument is likely unintended
```

BrettMayson: It's might be where DIK_NUMPADENTER is being used, and the error is just pointing to the wrong location. The comment will be removed, but the padded whitespace won't

PabstMirror: yeah, it's on use not on define
which is actually good because it means we don't have to touch the bi files
a nice-to-have would be to indicate usage location

PabstMirror:
Building with no warnings for hemtt 1.10
Will pass CI on merge due to changes to make.py

jonpas (approved):

PR #1630, opened 2023-11-24
HEMTT sqf compat

BrettMayson:
- Uses `.inc.sqf`
- Uses `XEH_PREP.hpp`


addons/accessory/XEH_preInit.sqf
```diff
@@ -2,7 +2,7 @@
 
 if (!hasInterface) exitWith {};
 
-#include "XEH_PREP.sqf"
+#include "XEH_PREP.hpp"
```
PabstMirror, line 5: not sure if hpp is right, its not a config, it's still just sqf?

```suggestion
#include "XEH_PREP.inc.sqf"
```

BrettMayson: I'm fine with either, just used `.hpp` since that's what ACE uses

PabstMirror: oh right, then it's fine

PabstMirror (approved):

jonpas (approved):

PR #1651, opened 2024-03-26
UI - Improve support of pictures in mission list filter

veteran29:
- title

Reaction Forces Creator DLC adds DLC pictures to mission lists, this will save and restore them during the filtering. Right picture is not used by anything but won't hurt to save/restore it in case some mod does.

Vanilla:
image

Reaction Forces:
image


veteran29:
It still removes the icons when you toggle between stock/custom missions but shows them properly when loading the display and change the worlds. I guess won't hurt to merge and possibly improve later?

jonpas (approved):
It fixes it halfway, so better than nothing. Undraft it and merge.

PabstMirror (approved):

PR #1655, opened 2024-04-04
Joint Rails - Fix `CBA_fnc_compatibleItems` undefined variable

### LINKED ISSUE #1653: CBA_fnc_compatibleItems throws undefined variable errors when called with a single parameter
**Mods (complete and add to the following information):**
- **Arma 3:** `2.16 stable`
- **CBA:** `3.17 release, whatever that is`

**Description:**
CBA_fnc_compatibleItems throws an undefined variable error when called with a single parameter in scheduled execution, eg:

```
0 spawn {
  ["arifle_MX_F"] call CBA_fnc_compatibleItems;
};
```
Results in:
```
19:37:53 Error in expression <key = format ["%1#%2", _weapon, ["all", _typefilter] select _typeFilterExists];>
19:37:53   Error position: <_typefilter] select _typeFilterExists];>
19:37:53   Error Undefined variable in expression: _typefilter
19:37:53 File /x/cba/addons/jr/fnc_compatibleItems.sqf..., line 267
```
The select is invalid in scheduled mode because `_typefilter` is dereferenced regardless of the value of `_typeFilterExists`.

This breaks several routines in Antistasi because the function also replaces BIS_fnc_compatibleItems.

jonpas:
- Fix #1653

I don't see a downside in doing `""` as default value, but maybe I am missing something.

### DIFF
--- a/addons/jr/fnc_compatibleItems.sqf
+++ b/addons/jr/fnc_compatibleItems.sqf
@@ -23,7 +23,7 @@ Author:
 ---------------------------------------------------------------------------- */
 SCRIPT(compatibleItems);
 
-params [["_weapon", "", [""]], ["_typefilter", nil, ["", 0]]];
+params [["_weapon", "", [""]], ["_typefilter", "", ["", 0]]];
 
 if (_weapon == "") exitWith {[]};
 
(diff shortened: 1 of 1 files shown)


PabstMirror:
`count (["LMG_Mk200_F", ""] call CBA_fnc_compatibleItems)` goes from 0 to 75
but I don't have a problem with this because "" was never a valid input

PR #1660, opened 2024-04-30
General - Fix missing `ADDON` variable

johnb432:
- Title.
- In its initial form this is limited only to `accessory`, but I want to fix all issues in all components with `ADDON`.
  - I've noticed stuff like https://github.com/CBATeam/CBA_A3/blob/4d29fdf198751625743a7a4c404eb871d227bdda/addons/help/XEH_preInit.sqf#L6
  where it's only set if the machine has an interface. In `ui`, it's defined regardless of interface presence or not.
  - I've notices other stuff like https://github.com/CBATeam/CBA_A3/blob/4d29fdf198751625743a7a4c404eb871d227bdda/addons/optics/XEH_preInit.sqf#L3
  where it's only set to `false` on every machine, but `true` only on select machines.
  
  This behaviour seems to be inconsistent and I was to address that in this PR. However, in order to better understand what solution might be the best: What is the purpose of `ADDON` exactly?

### DIFF
--- a/addons/accessory/XEH_preInit.sqf
+++ b/addons/accessory/XEH_preInit.sqf
@@ -2,6 +2,8 @@
 
 if (!hasInterface) exitWith {};
 
+ADDON = false;
+
 #include "XEH_PREP.hpp"
 
 GVAR(usageHash) = createHashMap;
@@ -41,3 +43,5 @@ GVAR(usageHash) = createHashMap;
         false
     }
 ] call CBA_fnc_addItemContextMenuOption;
+
+ADDON = true;
--- a/addons/accessory/XEH_preInit.sqf
+++ b/addons/accessory/XEH_preInit.sqf
@@ -1,8 +1,10 @@
 #include "script_component.hpp"
 
-if (!hasInterface) exitWith {};
-
 ADDON = false;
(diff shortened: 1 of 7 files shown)


PabstMirror:
IIRC
```
ADDON = false;
initCode
ADDON = true;
```
was designed so that you can diagnose if there is an error in init code that stops execution
(undefined/false/true)
I think it is still possible for some kinds of errors to fully break preInit/postInit for other modules

but IMHO the rpt will do a better job of showing the error location than checking these vars

PabstMirror (approved):

johnb432:
@PabstMirror I want to fix all inconsistencies with `ADDON` then. Define them where necessary and make sure that, regardless of machine type, the `ADDON` variable for each component is set to `true`, if `preInit` completes successfully. I'll do the same for ACE at some point.
However, given you've approved the PR as is, do you want me to make separate PRs for individual components, another PR with all changes combined or utilise this one?

PabstMirror:
feel free to do it all in this PR or whatever is easiest

PabstMirror (approved):

jonpas (approved):

PR #1662, opened 2024-05-04
Disposable - Updated `cba_disposable_fnc_replaceMagazineCargo`

johnb432:
- Title.
- Cleaned up `FUNC(replaceMagazineCargo)`.
- Now also checks if launchers can fit in vests and uniform and replaces them if possible.
  I think this case is rare, but it's technically possible so I think CBA should be able to handle it.
- Now handles all containers within containers (previously only backpacks within containers).
- Fixes a bug related to grenades not being throwable (https://feedback.bistudio.com/T74244), as this now only removes the necessary magazines and not every magazine.

### DIFF
--- a/addons/disposable/fnc_replaceMagazineCargo.sqf
+++ b/addons/disposable/fnc_replaceMagazineCargo.sqf
@@ -51,13 +51,8 @@ if (_magazines isEqualTo []) exitWith {};
 _containerType = if (getNumber (configOf _container >> "isBackpack") == 1) then {
     TYPE_BACKPACK
 } else {
-    private _itemType = configFile >> "CfgWeapons" >> _containerType >> "ItemInfo" >> "type";
-
-    if (isClass _itemType) then {
-        getNumber _itemType
-    } else {
-        0
-    };
+    // If uniform or vest, this config will be defined, otherwise it will default to 0
+    getNumber (configFile >> "CfgWeapons" >> _containerType >> "ItemInfo" >> "type")
 };
 
 // Replace magazines with disposable launchers
@@ -66,8 +61,8 @@ _containerType = if (getNumber (configOf _container >> "isBackpack") == 1) then 
 
     private _loadedLauncher = GVAR(MagazineLaunchers) getVariable _x;
 
-    // Ignore slot restrictions if not a uniform, vest or backpack
-    // If slot restrictions apply, don't add weapon but remove magazine
+    // Slot restrictions only apply if uniform, vest or backpack
+    // If slot restrictions apply, remove magazine but don't add weapon
     if (_containerType == 0 || {_containerType in (GVAR(allowedSlotsLaunchers) getOrDefault [_loadedLauncher, []])}) then {
         _container addWeaponCargoGlobal [_loadedLauncher, 1];
     };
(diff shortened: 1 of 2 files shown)


addons/disposable/fnc_replaceMagazineCargo.sqf
```diff
@@ -6,61 +6,41 @@ Description:
...
-};
+if (!local _container) exitWith {};
+if (missionNamespace getVariable [QGVAR(disableMagazineReplacement), false]) exitWith {};
 
-private _backpackContainer = backpackContainer _box;
-if (!isNull _backpackContainer) then {
-    _backpackContainer call FUNC(replaceMagazineCargo);
-};
+private _containers = everyBackpack _container;
+_containers append ([uniformContainer _container, vestContainer _container, backpackContainer _container] select {!isNull _x});
```
PabstMirror, line 31:
```suggestion
if (_container isKindOf "CaManBase") then {
    _containers append ([uniformContainer _container, vestContainer _container, backpackContainer _container] select {!isNull _x});
};
```
possibly, I'll test this later

PabstMirror:
looks good overall, I don't understand the `"isBackpack") != -1;` either
also wondering if `addWeaponCargoGlobal` will ignore slot restrictions, will test in a bit

johnb432:
> also wondering if `addWeaponCargoGlobal` will ignore slot restrictions, will test in a bit

It does - I'm guessing that code was to stop that then.
Do we want to keep the current behaviour? I don't like that stuff vanishes, just because it isn't allowed to be in a backpack. I'd prefer if it were dropped on the ground, but that become quite messy.

PabstMirror:
loadout for b_lat:
2 nlaws in the backpack (nlaws can't normally fit in backpacks)


But I agree that this is better than having things disappear
i.e. in vanilla you would get 3 shots, now you still have 3 shots

I guess we could always look at a setting if people really have a problem with it? (I'd rather not have to)

PabstMirror (approved):

Drofseh:
I think you should always get the number of shots the mission maker intended. So fill up the backpack if required.

johnb432:
As currently nobody wants to change it, but *if* in future we want to modify this PR, then I would not revert the code to the old stuff. It didn't take into account if launchers could fit in uniforms or vests, which is probably rare but not impossible.

johnb432:
Discussion in discord (https://discord.com/channels/976165959041679380/976228110078992456/1237139353650597908) concluded that the PR in its original form was not wanted. I made a new PR #1664 where only the fix is present.

This PR now only brings the code up-to-date and also checks if launchers can fit in uniforms and vests and replaces them if possible.

PabstMirror:
pushed a small change allowing us to keep the same args for `cba_disposable_fnc_replaceMagazineCargo`
my worry is someone might be doing something like
- add loadout to crate
- call cba_disposable_fnc_replaceMagazineCargo

and the change in args might cause problems (even though it says `Internal Function`)

PabstMirror (approved):

johnb432:
> and the change in args might cause problems (even though it says `Internal Function`)

I really feel that's a "them" problem.
If you want the changes to remain, I'll need to cleanup some leftover stuff.

johnb432:
> I really feel that's a "them" problem. If you want the changes to remain, I'll need to cleanup some leftover stuff.

The reason I added both the container and its type is because it doesn't work for uniforms and vests. If you do `typeOf uniformContainer player`, it returns something like `"Supply40"`, which can't be used to determine if a launcher is supposed to fit in a uniform/vest or not.

Usually launchers don't fit in uniforms and vests, but I'd prefer to be thorough and take every possibility into account.

PR #1663, opened 2024-05-04
Disposable - Use native hashmaps

johnb432:
- Title.
- ~~ACE would need to be updated, see https://github.com/acemod/ACE3/blob/9d168756df555cdc71db40343bc9d4c9bfd196b2/addons/arsenal/functions/fnc_onSelChangedRight.sqf#L72.~~ Already taken care of in https://github.com/acemod/ACE3/pull/9998.

### DIFF
--- a/addons/disposable/XEH_preInit.sqf
+++ b/addons/disposable/XEH_preInit.sqf
@@ -39,7 +39,7 @@ private _cfgMagazines = configFile >> "CfgMagazines";
     // Get case-sensitive config names
     private _configLauncher = _cfgWeapons >> configName _x;
     private _launcher = configName _configLauncher;
-    private _magazine = configName (_cfgMagazines >> (getArray (_configLauncher >> "magazines") select 0));
+    private _magazine = (compatibleMagazines (configName _x)) select 0;
 
     if (_magazine == "") then {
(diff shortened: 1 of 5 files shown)


addons/disposable/XEH_preInit.sqf
```diff
@@ -25,19 +25,21 @@ if (configProperties [configFile >> "CBA_DisposableLaunchers"] isEqualTo []) exi
...
 private _cfgWeapons = configFile >> "CfgWeapons";
 private _cfgMagazines = configFile >> "CfgMagazines";
 
 {
-    private _launcher = configName _x;
-    private _magazine = configName (_cfgMagazines >> (getArray (_cfgWeapons >> _launcher >> "magazines") select 0));
+    // Get case-sensitive config names
+    private _configLauncher = _cfgWeapons >> configName _x;
+    private _launcher = configName _configLauncher;
+    private _magazine = configName (_cfgMagazines >> (getArray (_configLauncher >> "magazines") select 0));
```
PabstMirror, line 42:
```suggestion
    private _magazine = (compatibleMagazines (configName _x)) select 0;
```
might be possible to have a `magazine[]= {}` and just use mag wells, not likely
but it automatically gets config-case

PabstMirror (approved):

PR #1670, opened 2024-06-18
Events - Add `CBA_fnc_addBISPlayerEventHandler` for adding EHs to just the player (from ACE3)

LinkIsGrim:
- Title, see https://github.com/acemod/ACE3/pull/10056, though this is not intended for public usage/API (it's a secret tool that will help us later).

~~~I'm not sure if the code block below is equivalent, since there's no `ACE_player` here.~~~ It isn't.
```
if (_ignoreVirtual && {(unitIsUAV focusOn) || {getNumber (configOf focusOn >> "isPlayableLogic") == 1}}) exitWith {};
```

### DIFF
--- a/addons/events/fnc_addBISPlayerEventHandler.sqf
+++ b/addons/events/fnc_addBISPlayerEventHandler.sqf
@@ -1,6 +1,6 @@
 #include "script_component.hpp"
 /* ----------------------------------------------------------------------------
-Function: CBA_fnc_addPlayerEngineEvent
+Function: CBA_fnc_addBISPlayerEventHandler
 
 Description:
     Adds an engine event handler just to the controlled entity.
@@ -15,13 +15,13 @@ Returns:
 
 Examples:
     (begin example)
-        ["example", "FiredNear", {systemChat str _this}] call CBA_fnc_addPlayerEngineEvent
+        ["example", "FiredNear", {systemChat str _this}] call CBA_fnc_addBISPlayerEventHandler
     (end)
 
 Author:
     PabstMirror
 ---------------------------------------------------------------------------- */
-SCRIPT(addPlayerEngineEvent);
+SCRIPT(addBISPlayerEventHandler);
 
 params [
     ["_type", "", [""]],
--- a/addons/events/fnc_addBISPlayerEventHandler.sqf
+++ b/addons/events/fnc_addBISPlayerEventHandler.sqf
@@ -15,7 +15,7 @@ Returns:
 
 Examples:
     (begin example)
-        ["example", "FiredNear", {systemChat str _this}] call CBA_fnc_addBISPlayerEventHandler
+        ["FiredNear", {systemChat str _this}] call CBA_fnc_addBISPlayerEventHandler
     (end)
 
 Author:
--- a/addons/events/fnc_addBISPlayerEventHandler.sqf
+++ b/addons/events/fnc_addBISPlayerEventHandler.sqf
@@ -6,16 +6,17 @@ Description:
     Adds an engine event handler just to the controlled entity.
 
 Parameters:
+    _key           - Unique identifier for the event. <STRING>
     _eventType     - Type of event to add. Can be any event supported by addEventHandler. <STRING>
     _eventCode     - Code to run when event is raised. <CODE>
-    _ignoreVirtual - Ignore virtual units (spectators, virtual zeus, UAV RC) [optional] (default: false) <BOOLEAN>
+    _ignoreVirtual - Ignore virtual units (spectators, virtual zeus, UAV RC) [optional] (default: true) <BOOLEAN>
 
 Returns:
-    None
+    Event was added <BOOLEAN>
 
 Examples:
     (begin example)
-        ["FiredNear", {systemChat str _this}] call CBA_fnc_addBISPlayerEventHandler
+        ["TAG_MyFiredNearEvent", "FiredNear", {systemChat str _this}] call CBA_fnc_addBISPlayerEventHandler
     (end)
 
 Author:
@@ -23,10 +24,15 @@ Author:
 ---------------------------------------------------------------------------- */
 SCRIPT(addBISPlayerEventHandler);
 
+if (canSuspend) exitWith {
+    [CBA_fnc_addBISPlayerEventHandler, _this] call CBA_fnc_directCall;
+};
+
 params [
+    ["_key", "", [""]],
     ["_type", "", [""]],
     ["_code", {}, [{}]],
-    ["_ignoreVirtual", false, [false]]
+    ["_ignoreVirtual", true, [true]]
 ];
 
 if (isNil QGVAR(playerEventsHash)) then { // first-run init
@@ -58,8 +64,8 @@ if (isNil QGVAR(playerEventsHash)) then { // first-run init
 };
 
 
-private _key = format [QGVAR(playerEvents_%1), toLowerANSI _type]; // should only have one per event, this isn't public
-if (_key in GVAR(playerEventsHash)) exitWith { ERROR_1("bad key %1",_this); };
+_key = format [QGVAR(playerEvents_%1), toLowerANSI _key];
+if (_key in GVAR(playerEventsHash)) exitWith { ERROR_1("bad key %1",_this); false};
 
 GVAR(playerEventsHash) set [_key, [_type, _code, _ignoreVirtual]];
 
@@ -69,3 +75,5 @@ if (_ignoreVirtual && {(unitIsUAV _player) || {getNumber (configOf _player >> "i
 // Add event now
 private _newEH = _player addEventHandler [_type, _code];
 _player setVariable [_key, _newEH];
+
+true // return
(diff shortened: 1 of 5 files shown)


addons/events/fnc_addPlayerEngineEvent.sqf
```diff
@@ -0,0 +1,70 @@
+#include "script_component.hpp"
+/* ----------------------------------------------------------------------------
+Function: CBA_fnc_addPlayerEngineEvent
```
veteran29, line 3: The function adds an Event **Handler** so the name should reflect that.

We have `addBISEventHandler` so something like `addBISPlayerEventHandler` would be fitting. I don't like how the capitalized letters merge though.

veteran29 (dismissed):

PabstMirror:
I think we can replace ace_player with `call CBA_fnc_currentUnit`
iirc focusOn might be funny when controlling a uav, not sure

addons/events/fnc_addPlayerEngineEvent.sqf
```diff
@@ -0,0 +1,70 @@
...
+    }, false] call CBA_fnc_addPlayerEventHandler;
+};
+
+
+private _key = format [QGVAR(playerEvents_%1), toLowerANSI _type]; // should only have one per event, this isn't public
+if (_key in GVAR(playerEventsHash)) exitWith { ERROR_1("bad key %1",_this); };
+
+GVAR(playerEventsHash) set [_key, [_type, _code, _ignoreVirtual]];
+
+if (_ignoreVirtual && {(unitIsUAV focusOn) || {getNumber (configOf focusOn >> "isPlayableLogic") == 1}}) exitWith {};
```
PabstMirror, line 66:
```suggestion
private _player = call CBA_fnc_currentUnit;
if (_ignoreVirtual && {(unitIsUAV _player ) || {getNumber (configOf _player >> "isPlayableLogic") == 1}}) exitWith {};
```

addons/events/fnc_addPlayerEngineEvent.sqf
```diff
@@ -0,0 +1,70 @@
...
+
+// Add event now
+private _newEH = focusOn addEventHandler [_type, _code];
```
PabstMirror, line 69:
```suggestion
private _newEH = _player addEventHandler [_type, _code];
```

addons/events/fnc_addPlayerEngineEvent.sqf
```diff
@@ -0,0 +1,70 @@
...
+focusOn setVariable [_key, _newEH];
```
PabstMirror, line 70:
```suggestion
_player setVariable [_key, _newEH];
```

addons/events/fnc_addBISPlayerEventHandler.sqf
```diff
@@ -0,0 +1,71 @@
...
+};
+
+
+private _key = format [QGVAR(playerEvents_%1), toLowerANSI _type]; // should only have one per event, this isn't public
+if (_key in GVAR(playerEventsHash)) exitWith { ERROR_1("bad key %1",_this); };
+
+GVAR(playerEventsHash) set [_key, [_type, _code, _ignoreVirtual]];
+
+private _player = call CBA_fnc_currentUnit;
+if (_ignoreVirtual && {(unitIsUAV _player ) || {getNumber (configOf _player >> "isPlayableLogic") == 1}}) exitWith {};
```
commy2, line 67:
```suggestion
if (_ignoreVirtual && {(unitIsUAV _player) || {getNumber (configOf _player >> "isPlayableLogic") == 1}}) exitWith {};
```

addons/events/fnc_addBISPlayerEventHandler.sqf
```diff
@@ -0,0 +1,71 @@
...
+            if (_ignoreVirtual && _isVirtual) then { continue };
+
+            private _newEH = _newPlayer addEventHandler [_type, _code];
+            _newPlayer setVariable [_x, _newEH];
+        } forEach GVAR(playerEventsHash);
+    }, false] call CBA_fnc_addPlayerEventHandler;
+};
+
+
+private _key = format [QGVAR(playerEvents_%1), toLowerANSI _type]; // should only have one per event, this isn't public
```
commy2, line 61: Prefixing the keys is completely unnecessary, as there can be no name collisions here.

addons/events/fnc_addBISPlayerEventHandler.sqf
```diff
@@ -0,0 +1,71 @@
...
+    _eventType     - Type of event to add. Can be any engine (not mission) event. <STRING>
+    _eventCode     - Code to run when event is raised. <CODE>
+    _ignoreVirtual - Ignore virtual units (spectators, virtual zeus, UAV RC) [optional] (default: false) <BOOLEAN>
+
+Returns:
+    None
+
+Examples:
+    (begin example)
+        ["example", "FiredNear", {systemChat str _this}] call CBA_fnc_addBISPlayerEventHandler
```
commy2, line 18: Seems wrong.

addons/events/fnc_addBISPlayerEventHandler.sqf
```diff
@@ -0,0 +1,71 @@
...
+/* ----------------------------------------------------------------------------
+Function: CBA_fnc_addBISPlayerEventHandler
+
+Description:
+    Adds an engine event handler just to the controlled entity.
+
+Parameters:
+    _eventType     - Type of event to add. Can be any engine (not mission) event. <STRING>
+    _eventCode     - Code to run when event is raised. <CODE>
+    _ignoreVirtual - Ignore virtual units (spectators, virtual zeus, UAV RC) [optional] (default: false) <BOOLEAN>
```
commy2, line 11: Is including 'virtual entities' what the average user wants? If no, the default value should be `true`.

addons/events/fnc_addBISPlayerEventHandler.sqf
```diff
@@ -0,0 +1,71 @@
+#include "script_component.hpp"
+/* ----------------------------------------------------------------------------
+Function: CBA_fnc_addBISPlayerEventHandler
+
+Description:
+    Adds an engine event handler just to the controlled entity.
+
+Parameters:
+    _eventType     - Type of event to add. Can be any engine (not mission) event. <STRING>
```
commy2, line 9: There are more non-Mission event handlers in the game: [Scripted Event Handlers](https://community.bistudio.com/wiki/Arma_3:_Scripted_Event_Handlers) or [User Interface Event Handlers](https://community.bistudio.com/wiki/User_Interface_Event_Handlers) to name some.

This works with anything from `addEventHandler` only.

```suggestion
    _eventType     - Type of event to add. Can be any event supported by addEventHandler. <STRING>
```

addons/events/fnc_addBISPlayerEventHandler.sqf
```diff
@@ -0,0 +1,71 @@
...
+
+            private _newEH = _newPlayer addEventHandler [_type, _code];
+            _newPlayer setVariable [_x, _newEH];
+        } forEach GVAR(playerEventsHash);
+    }, false] call CBA_fnc_addPlayerEventHandler;
+};
+
+
+private _key = format [QGVAR(playerEvents_%1), toLowerANSI _type]; // should only have one per event, this isn't public
+if (_key in GVAR(playerEventsHash)) exitWith { ERROR_1("bad key %1",_this); };
```
commy2, line 62: Does this mean that only one event of each type can be added? If so, how does this fit the spirit of CBA?

addons/main/script_mod.hpp
```diff
@@ -10,7 +10,7 @@
 #define VERSION_AR  MAJOR,MINOR,PATCHLVL,BUILD
 
 // MINIMAL required version for the Mod. Components can specify others..
-#define REQUIRED_VERSION 2.14
+#define REQUIRED_VERSION 2.16
```
commy2, line 13:
```suggestion
#define REQUIRED_VERSION 2.14
```
Without `focusOn`, this can remain 2.14.

commy2:
Just a few thoughts.

- It stylistically correctly omits a version to remove the event handlers.

- It is not thread-safe (`spawn`-pseudo threads), like most the rest of CBA tbf.

- This seems related to common component [addPlayerAction](https://github.com/CBATeam/CBA_A3/blob/master/addons/common/fnc_addPlayerAction.sqf), more so than events component [addPlayerEventHandler](https://github.com/CBATeam/CBA_A3/blob/master/addons/events/fnc_addPlayerEventHandler.sqf). The names are unfortunate.

I don't personally like `CBA_fnc_addPlayerAction` and always thought of it as legacy code. I would always recommend to simply add user actions (and by extension event handlers) to every unit and then to quit early via guard clause. Is the mental overhead required for using such an obscure framework justified considering one could simply write:
```sqf
if (_unit != player) exitWith {};
```
?

PabstMirror:
There is some performance savings by only adding events you need.
e.g. FiredNear could almost be considered n^2 based on unit count

We can make it safe by wrapping in a `CBA_fnc_directCall;` like https://github.com/CBATeam/CBA_A3/blob/master/addons/events/fnc_addEventHandler.sqf

PabstMirror (approved):

addons/events/fnc_addPlayerEngineEvent.sqf
PabstMirror: anyone have a different name suggestion?
otherwise I think it's fine as is

PR #1671, opened 2024-06-23
Player Events - Add debug logging & make VisionModeChanged use engine event

LinkIsGrim:
*When merged this pull request will:**
- Do what it says on the tin

Proof of concept PR, want to see what changes will be necessary for other events (delay for one frame?) and catch any edge cases (one I can think of is player not using NVG and switching to UAV with NVG on, and vice-versa. I'm not sure of the best way to handle that).

Requires #1669 & #1670

### DIFF
--- a/addons/events/fnc_addPlayerEventHandler.sqf
+++ b/addons/events/fnc_addPlayerEventHandler.sqf
@@ -200,6 +200,11 @@ if (_id != -1) then {
             };
         }, 0.5] call CBA_fnc_addPerFrameHandler);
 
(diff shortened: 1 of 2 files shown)


PabstMirror:
quick tested controlling UAVs and the event seems to work fine
eh is passed:
```
[player,1,-1,0,-1,UAV,[0]]
```
vision mode seems to just be whatever you are controlling, e.g. no nvgs
```
hmd player = ""
currentVisionMode player = 1
```

PR #1672, opened 2024-07-02
Strings - Add function that creates a random string of letters and numbers with selectable length.

AndreasBrostrom:
- Added function that creates a random string of letters and numbers with selectable length.

### DIFF
--- a/addons/strings/fnc_randomString.sqf
+++ b/addons/strings/fnc_randomString.sqf
@@ -33,4 +33,4 @@ for "_i" from 1 to _length do {
     _string = [(selectRandom _array), _string] joinString "";
 };
 
-_string;
+_string
--- a/addons/strings/fnc_randomString.sqf
+++ b/addons/strings/fnc_randomString.sqf
@@ -1,0 +1,36 @@
+#include "script_component.hpp"
+/* ----------------------------------------------------------------------------
+Function: CBA_fnc_randomString
+
+Description:
+    Creates a randomized string with letters and numbers.
+
+Parameters:
+    _length - Length of the string [String] (default: 10)
+
+Returns:
+    Randomized string [String]
+
+Example:
+    (begin example)
+        _result = [20] call CBA_fnc_randomString;
+        // _result => "03q7npfpthkyiush4yz5"
+
+        _makerName = [20] call CBA_fnc_randomString
+        _marker = createMarker [_makerName, player];
+    (end)
+
+Author:
+    AndreasBrostrom
+
+---------------------------------------------------------------------------- */
+params [["_length", 10, [0]]];
+
+private _array = ["0", "1", "2", "3", "4", "5", "6", "7", "8", "9", "a", "b", "c", "d", "e", "f", "g", "h", "i", "j", "k", "l", "m", "n", "o", "p", "q", "r", "s", "t", "u", "v", "w", "x", "y", "z"];
+private _string = "";
+
+for "_i" from 1 to _length do {
+    _string = [(selectRandom _array), _string] joinString "";
+};
+
+_string;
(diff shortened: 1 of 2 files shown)


addons/strings/fnc_randomString.sqf
```diff
@@ -0,0 +1,36 @@
...
+params [["_length", 10, [0]]];
+
+private _array = ["0", "1", "2", "3", "4", "5", "6", "7", "8", "9", "a", "b", "c", "d", "e", "f", "g", "h", "i", "j", "k", "l", "m", "n", "o", "p", "q", "r", "s", "t", "u", "v", "w", "x", "y", "z"];
+private _string = "";
+
+for "_i" from 1 to _length do {
+    _string = [(selectRandom _array), _string] joinString "";
+};
+
+_string;
```
PabstMirror, line 36:
```suggestion
_string

```

addons/strings/fnc_randomString.sqf
```diff
@@ -0,0 +1,36 @@
...
+    AndreasBrostrom
+
+---------------------------------------------------------------------------- */
+params [["_length", 10, [0]]];
+
+private _array = ["0", "1", "2", "3", "4", "5", "6", "7", "8", "9", "a", "b", "c", "d", "e", "f", "g", "h", "i", "j", "k", "l", "m", "n", "o", "p", "q", "r", "s", "t", "u", "v", "w", "x", "y", "z"];
+private _string = "";
+
+for "_i" from 1 to _length do {
+    _string = [(selectRandom _array), _string] joinString "";
```
PabstMirror, line 33: this is a little faster and I think easier to read
```suggestion
    _string = _string + selectRandom _array; 
```

commy2:
0. I don't think the difference will be significant for short strings in SQF, but you generally want to avoid making new strings each iteration. I think you were reaching for this initially:

```sqf
    params [["_length", 10, [0]]];

    private _chars = "abcdefghijklmnopqrstuvwxyz0123456789" splitString "";

    private _result = [];

    for "_" from 1 to _length do {
        _result pushBack selectRandom _chars; 
    };

    _result joinString "" // return
```

1. How is this significantly different from [createUUID](https://github.com/CBATeam/CBA_A3/blob/master/addons/common/fnc_createUUID.sqf)?

2. I really don't like the example. Choosing a random name for markers is the wrong thing to do. Markers need a unique name, not a random name, and those are different concepts. What issue are you actually trying to solve with this? If it's markers, the canonical way would probably involve some (atomic -- in case you want to support threads) incr on a global number combined with the client id.

AndreasBrostrom:
0. You are right. Your new version here is way better.

1. I did not know of its existence and did not see it. This is basically do the same but uses the entire alphabeth and numbers with no dividers have a selectable range. Can be used to generate a hash of some kind.

2. My initial plan and reason behind the marker exsample was based on a experiment with creation of markers on a ace_tagging event on the map with something like this:
   ```sqf
   _name = format ["_USER_DEFINED xx_taggedmarker_%1_xx", [10] call CBA_fnc_randomString];
   createMarker [_name , player];
   ```
   But found this in the [createMarker documentation](https://community.bistudio.com/wiki/createMarker) doing this: `format ["_USER_DEFINED #%1/%2/%3" , clientOwner,_index, _ChannelID];` and its better. But i also scraped the idea.
   
   So when i did not see any hashing or random string function i figured i toss it in here due to it not being completely useless and can be used for something.

jokoho48:
If you want a hash, why not use https://community.bistudio.com/wiki/hashValue? It is an actual working hashing function built into Arma. This doesn't have any real usage except for generating some random Password-style Gameplay Feature, but that could be done via the building hash or just hard coded. I think due to the other functions and possibilities, this does not is needed in CBA

PR #1675, opened 2024-07-24
AI - Update fnc_addWaypoint Header

prisonerMO:
addWaypoint return array
[group, waypoint]

### DIFF
--- a/addons/ai/fnc_addWaypoint.sqf
+++ b/addons/ai/fnc_addWaypoint.sqf
@@ -26,7 +26,7 @@ Example:
     (end)
 
 Returns:
-    Waypoint
+    [Group,Waypoint]
 
 Author:
     Rommel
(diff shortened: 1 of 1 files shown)


PabstMirror:
waypoint is kind of a "type" already - https://community.bistudio.com/wiki/Waypoint
but I have no problem changing it to be more explicit

addons/ai/fnc_addWaypoint.sqf
```diff
@@ -26,7 +26,7 @@ Example:
     (end)
 
 Returns:
-    Waypoint
+    [Group,Waypoint]
```
PabstMirror, line 29:
```suggestion
    Waypoint [Group, Waypoint Index] <ARRAY>
```

PabstMirror (approved):

PR #1678, opened 2024-07-26
Optics - Add scripted optics index config entry

johnb432:
- This allows one to set which optics index should be used for the scripted optic.
  I'm continuing the work on making ACE scopes use CBA optics (https://github.com/acemod/ACE3/pull/7273), but came across the limitation that the scripted optic needed to be the first optics mode of the optic. This PR changes that so that you can specify which optics mode is the scripted one.
  E.g. you can now have the primary optics mode be a collimator and the secondary optics mode be the optics using CBA optics.
- This PR assumes that you can't have more than 1 scripted optic mode per optic.
- Documentation needs to be added.
- Tested with samples from ACE (https://github.com/acemod/ACE3/pull/7273), CUP, RKSL and BWMod scopes: it doesn't seem to affect existing scopes negatively.

### DIFF
--- a/addons/optics/fnc_setOpticMagnification.sqf
+++ b/addons/optics/fnc_setOpticMagnification.sqf
@@ -24,8 +24,6 @@ Author:
 if (!isNil {parsingNamespace getVariable QGVAR(magnification)}) exitWith {};
 
 params ["_unit", "_magnification"];
-private _weapon = currentWeapon _unit;
-private _optic = _unit call FUNC(currentOptic);
 
 // store zeroing
 private _zeroing = GVAR(ZeroingDistances) find currentZeroing _unit;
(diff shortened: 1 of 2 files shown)


PabstMirror (changes requested):
seems to break restoring magnification on variable zoomed optics
e.g. `player addPrimaryWeaponItem "CUP_optic_LeupoldMk4"`
zoom a bit, unscope, rescope and it reset to lowest

PabstMirror:
it seems like restoring the `addWeaponItem` fixes.
```
_unit addWeaponItem [_weapon, _optic];
_unit setOpticsMode GVAR(scriptedOpticsIndex);
```

johnb432:
> it seems like restoring the `addWeaponItem` fixes.
> 
> ```
> _unit addWeaponItem [_weapon, _optic];
> _unit setOpticsMode GVAR(scriptedOpticsIndex);
> ```

I'll look into adding both `addWeaponItem` lines again.

johnb432:
`_unit addWeaponItem [_weapon, _optic];` reverts the optics mode to the default one and `_unit setOpticsMode GVAR(scriptedOpticsIndex);` reverts the magnification of the scope.
I need to rethink how to go about this...

johnb432:
> `_unit addWeaponItem [_weapon, _optic];` reverts the optics mode to the default one and `_unit setOpticsMode GVAR(scriptedOpticsIndex);` reverts the magnification of the scope. I need to rethink how to go about this...

Because there is no way of setting FOV, it means that it's impossible to try to do what I'd like to do.

The second `addWeaponItem` is needed, so that you can use the +/- for zooming in (I really wish I knew why that's the case, doesn't make sense in the slighest - I guess it's some sort of reset).
`setOpticsMode` is too dumb to use the magnification level defined in the config.

So it's either we lose magnification restoration or we can add this - therefore I'm closing this.

PR #1684, opened 2024-08-16
Help - Sort addons, credits and keybindings alphabetically

johnb432:
- Title.
- Fixes mods such as `O&T Expansion Eden` breaking the Addons diary entry, because of the `&` not being sanitised.
- I haven't tested if this works with keybinds that are added via config - I don't know of any mods that use that.

### DIFF
--- a/addons/help/XEH_preStart.sqf
+++ b/addons/help/XEH_preStart.sqf
@@ -38,31 +38,37 @@ private _credits = [];
     _credits pushBack format ["<font color='#bdcc9c'>%1%2 by %3</font>", _name, _version, _author];
 } forEach _addons;
 
-_credits = (_credits arrayIntersect _credits) joinString "<br/>";
+_credits = _credits arrayIntersect _credits;
+
+_credits sort true;
+
+_credits = _credits joinString "<br/>";
 
 uiNamespace setVariable [QGVAR(credits), compileFinal str _credits];
 
 // mods
 private _mods = "true" configClasses (configFile >> "CfgPatches") apply {configSourceMod _x};
-_mods = (_mods arrayIntersect _mods select {!isNumber (configfile >> "CfgMods" >> _x >> "appId")}) - [""];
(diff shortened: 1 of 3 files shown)


addons/help/XEH_preStart.sqf
```diff
@@ -38,32 +38,38 @@ private _credits = [];
...
         if (isText (_entry >> "description")) then {
-            private _description = getText (_entry >> "description") call CBA_fnc_sanitizeHTML;
+            private _description = getText (_entry >> "description");
 
             _x = _x + format ["<br/>%1<br/>", _description];
         };
     };
 
-    _x
+    _x call CBA_fnc_sanitizeHTML
```
PabstMirror, line 68: I think we want to revert this as it's sanitizing the font tag?
image

PabstMirror:
config keybinds look great (there are some demo on https://github.com/CBATeam/CBA_A3/blob/master/addons/events/config.cpp#L51)


image

they are a little strange in that you still need to add them via code
(and the keybind still shows if you don't, but that's how it worked before)

addons/help/XEH_preStart.sqf
johnb432: Never knew that was a thing, because for me steam make a folder named `@CBA_A3`, which never make it do the fancy name and description. Imo that's broken.

PabstMirror: looks like https://community.bistudio.com/wiki/modParams

is designed to take what `configSourceMod` gives us, which could be 
anything between "cba" or "@CBA_A3" or "450814997"

interesting comment on that cmd
> Later versions of PboProject and its friends with arma as the intended engine, will never allow a cfgmods to appear in a pbo. Instead, it will scribble it out to a mod.cpp in the mods\ folder that this pbo was intended for.

so I guess for most people descrption won't show, but it's not a big eal

johnb432: > so I guess for most people descrption won't show, but it's not a big eal

Could use `modParams`' `overview` or `tooltip` if they exists and are not `""`, but I don't really care either way.

PabstMirror (approved):

PR #1692, opened 2024-09-23
Common - Add color conversion functions

Drofseh:
- Add fuctions to convert to/from different colour formats.
- Unsure if colorHEXAtoDecimal is actually required. All the transparent HEX codes I found put the alpha channel at the front.
- It pains me immensely to spell the word colour wrong in the function names, but that's how BI spells it.
- Coauthored by @LambdaTiger


lambdatiger:
Apologies about that, it worked/compiled on my side so I don't exactly know how that happened.

PabstMirror (approved):

PR #1696, opened 2024-10-03
Replacing loop (fnc_startFallbackLoop) with event mission

Jonastex:
- This modification removes the loop that checks if all the new entities are properly initialized and replaces it with the mission event "EntityCreated" which allows for checking if the new entity is properly initialized and if not to do so. This modification allows for a performance gain


veteran29:
> This modification allows for a performance gain

Any benchmarks? `EntityCreated` runs for a lot more objects than `entity` command returns (bullets, impact decals, butterflies etc.) so would need filtering.

Jonastex:
> Any benchmarks?
Yes, I used the Arma Script profiler. Here are the results I got with a friend's modpack. These results are specific to each modpack because, for `CBA_fnc_StartFallBackLoop` to run, incompatible XEH mods must be used.
benchmark_cba


> EntityCreated runs for a lot more objects than entity command returns (bullets, impact decals, butterflies etc.) so would need filtering
If it's not necessary to initialize CBA on these entities, maybe filter the event handler with BIS_fnc_objectType?

addons/xeh/fnc_postInit.sqf
```diff
@@ -19,6 +19,17 @@ Author:
 isNil {
     XEH_LOG("PostInit started. " + PFORMAT_9("MISSIONINIT",missionName,missionVersion,worldName,isMultiplayer,isServer,isDedicated,CBA_isHeadlessClient,hasInterface,didJIP));
 
+    addMissionEventHandler ["EntityCreated", {
```
PabstMirror, line 22: we would only need to add this if:
we have incompatible vehicles
and
xeh were added to those vehicles

Jonastex: I added a filter to execute the code only on objects that are not compatible. 

I also added a foreach to initialize all the incompatible XEH objects that were either created earlier or are initially present on the map, and therefore are not detected by `EntityCreate`

addons/xeh/fnc_postInit.sqf
```diff
@@ -19,6 +19,28 @@ Author:
...
+    // init object incompatile XEH
+    {
+        if !(ISPROCESSED(_x)) then {
+            _x call CBA_fnc_initEvents;
+
+            if !(ISINITIALIZED(_x)) then {
+                _x call CBA_fnc_init;
+            };
+        };
+    } forEach (entities [(call (uiNamespace getVariable [QGVAR(incompatibleClasses), {[]}])), [], true, true]);
```
veteran29, line 42:
```suggestion
    } forEach (entities [call (uiNamespace getVariable [QGVAR(incompatibleClasses), {[]}]), [], true, true]);
```

veteran29, line 42: There's no point in adding the EH if there's no incompatible classes. Also if we would add this in preInit there would be no need to run the forEach loop.
```suggestion
    private _incompatibleClasses = call (uiNamespace getVariable [QGVAR(incompatibleClasses), {[]}]);
    if (count _incompatibleClasses > 0) then {

        addMissionEventHandler ["EntityCreated", {
            params ["_entity"];
            if ((!ISPROCESSED(_entity)) && (ISINCOMP(typeOf _entity))) then {
                _entity call CBA_fnc_initEvents;

                if !(ISINITIALIZED(_entity)) then {
                    _entity call CBA_fnc_init;
                };
            };
        }];
        
        // init object incompatile XEH
        {
            if !(ISPROCESSED(_x)) then {
                _x call CBA_fnc_initEvents;

                if !(ISINITIALIZED(_x)) then {
                    _x call CBA_fnc_init;
                };
            };
        } forEach (entities [_incompatibleClasses, [], true, true]);
    };
```

veteran29 (changes requested):

veteran29:
Your script profile includes only the fallback loop but nothing to compare to (EH).

addons/xeh/fnc_postInit.sqf
Jonastex: Yes its good idea

Jonastex:
Hello, here is the performance feedback I have with the EH modification
bench

PR #1697, opened 2024-10-07
Cleanup - Add private and remove some unused vars

PabstMirror:


.hemtt/lints.toml
```diff
@@ -7,3 +7,17 @@ options.ignore = [
...
+
+[sqf.unused]
+enabled = false
+options.check_params = false
+
+[sqf.shadowed]
+enabled = false
+
+[sqf.not_private]
+enabled = false
```
BrettMayson, line 23: I don't want these committed in case they change or never see main HEMTT, that was an issue in the past with random things getting copy pasted around

PR #1706, opened 2024-10-19
Common - Cleanup DEFAULT_PARAM, _this

BrettMayson:
Ref https://github.com/BrettMayson/HEMTT/pull/804

- Don't use DEFAULT_PARAM
- Use different variables for _this
- Rename the test for DEFAULT_PARAM to prevent HEMTT from yelling about it


PabstMirror (approved):

PabstMirror:
from discord

> commy2 — 10/19/2024 7:16 AM
In general, using _this has the benefit that it is shadowed inside the called function, whereas introducing a local variable makes it bleed into the new scope.
There is a reason why _this is used. I don't like this part of the PR (no pun intended).
e.g, there is a reason it's named _CBA_code and not just _code.

I'm not worried about most of these changes, except for xeh and direct call.
A counter example I can come up with is
```
[] spawn {
    private _args = ["A"]; // user code
    sleep 0.1;
    [{  
        _args pushBack 77;
    }, _this] call CBA_fnc_directCall;
    x3 = _args; // just ["A"], expecting ["A",77]
};
```

changing to `_CBA_args` seems to fix

PR #1715, opened 2024-11-09
Add `CBA_fnc_getMacro`

PabstMirror:
Lets you easily get the value of macros for use inside code that is otherwise compiled and static


BrettMayson:
Good solution IMO

BrettMayson (approved):
Good solution IMO

PR #1721, opened 2024-11-30
Add Missing DIK codes to _supportedKeys in XEH_preStart.sqf

Raven3-14:
Added missing DIK codes to _supportedKeys

- Add DIK codes "DIK_F16" through to "DIK_F24" to _supportedKeys, allowing them to be displayed as the correct key in CBA keybindings, rather than "Unknown Key ***"

This will make it match the code in perf/prof branch, and should fix the "Unknown Key ***" for F16~F24 when perf/prof is loaded.
I don't think the vanilla fix has been patched to stable. The relevant ticket in the BI feedback tracker has a note "Fixed in 2.20".


PabstMirror (approved):

PabstMirror:
just noting that we do need to wait for 2.20 or we get
```
20:08:48 [CBA] (settings) INFO: Finished reading settings from settings file.
20:08:48 Error in expression <EachPair;
} else {
[_hash, {
_namespace setVariable [_key, _value];
}] call CBA_>
20:08:48   Error position: <setVariable [_key, _value];
}] call CBA_>
20:08:48   Error Reserved variable in expression
```
on non-profiling

and because this file is compiled to sqfc the macros won't be defined until we update the includes

PabstMirror:
from what I can tell BIS never updated the dik codes in the include
so I'm still getting that script error and 
```
diag_log text format ["DIK_F22 is %1", DIK_F22];
10:16:55 DIK_F22 is any
```

PabstMirror:
As far as I can tell we are ok to manually define them based on https://feedback.bistudio.com/T125049
I do not have a keyboard with these keys so I will need someone to test this

PR #1734, opened 2025-02-25
Update scope check from 2 to 1 on magazine add/remove functions

ehylla93:
Some functions do a scope < 2 check instead of a scope < 1 check. This will hinder certain magazine items, which aren't accessible in Virtual Arsenal but in the game, to not properly removed from inventories.


PabstMirror (approved):
seems reasonable to me

PR #1753, opened 2025-03-27
JAM - Fix typo in PGM Hecate II magwell

Drofseh:


PabstMirror:
should we keep the old one for backwards compat?

Drofseh:
I don't know of anyone using it besides me.
The only two mods I'm aware of with the Hecate II are OFrP and R3F.
OFrP's current public release doesn't use it, and I assume R3F doesn't either since it hasn't been updated in several years.

PR #1756, opened 2025-04-07
Settings - Handle mismatched setting category capitalization

PabstMirror:
ref https://github.com/acemod/ACE3/issues/10838


rautamiekka:
This'll only makes devs lazier, to be very frank.

veteran29 (approved):
LGTM

PR #1757, opened 2025-05-01
Main - New Macros: QADDON, QQADDON

OverlordZorn:
Introduces QADDON and QQADDON

```hpp
#ifndef QADDON
    #define QADDON QUOTE(ADDON)
#endif
#ifndef QQADDON
    #define QQADDON QUOTE(QUOTE(ADDON))
#endif

```

happy to add more when suggested

- [x] moved Q,QQ into seperate PR: #1759

### DIFF
--- a/addons/main/script_macros_common.hpp
+++ b/addons/main/script_macros_common.hpp
@@ -1834,31 +1834,24 @@ Author:
 #define FILE_EXISTS(FILE) (fileExists (FILE))
 
 /* -------------------------------------------
-Macro: Q()
+Macro: QADDON, QQADDON
 
 Description:
-    Introduces a shortend version of QUOTE() and other frequently used quoted MARCOs
+    Quoted Variant of ADDON
 
 Parameters:
 
 Examples:
     (begin example)
-    Q(configName _x isEqualTo QQ(abe_banana)) configClasses (configFile >> QADDON)
+    QUOTE(configName _x isEqualTo QUOTE(QUOTE(abe_banana))) configClasses (configFile >> QADDON)
     (end)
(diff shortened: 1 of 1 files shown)


PabstMirror:
I'm a little worried about conflicting with existing macros in older code, mainly with `Q`

OverlordZorn:
> I'm a little worried about conflicting with existing macros in older code, mainly with `Q`

I searched CBA and ace and didnt find any macro thats defined that way.

I can make each a `#ifndef` if needed

jokoho48:
I don't think any of those macros are useful. If a mod itself wants them, they could always define them. This is just obscuring things it too much from my POV.

OverlordZorn:
> I don't think any of those macros are useful. If a mod itself wants them, they could always define them. This is just obscuring things it too much from my POV.

Just my opinion but cba already using Q in macro's like QGVAR/QQGVAR etc makes it for me the most logical step to also use Q/QQ isolated as well as combined with others like QADDON

Drofseh:
I don't see the point in shortening the general QUOTE to Q, I think it'll be less readable and the 4 character space savings don't matter.

I could see doing QADDON though, in the same way as QGVAR, as 
`#define QADDON QUOTE(ADDON)`

OverlordZorn:
moved Q,QQ into seperate PR
#1759

BrettMayson (approved):

BrettMayson:
probably should've moved these ADDON macros to another PR, so the comments regarding Q, QQ stayed relevant

PabstMirror (approved):

PR #1758, opened 2025-05-01
Common - New Function: cba_fnc_getCfgDataRandom

OverlordZorn:
## New Function: cba_fnc_getCfgDataRandom
This function extracts data from a config property.
If it is an Array, it will select a random entry from the array, otherwise it will simply return the provided data.
Will check if _cfg exists, if not, returns nil.

### DIFF
--- a/addons/common/fnc_getCfgDataRandom.sqf
+++ b/addons/common/fnc_getCfgDataRandom.sqf
@@ -4,7 +4,7 @@ Function: CBA_fnc_getCfgDataRandom
 
 Description:
     This function extracts data from a config property.
-    If it is an Array, it will select a random entry from the array, otherwise it will simply return the provided data.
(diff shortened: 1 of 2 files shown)


addons/common/fnc_getCfgDataRandom.sqf
```diff
@@ -0,0 +1,36 @@
+#include "script_component.hpp"
+/* ----------------------------------------------------------------------------
+Function: CBA_fnc_getCfgDataRandom
+
+Description:
+    This function extracts data from a config property.
+    If it is an Array, it will select a random entry from the array, otherwise it will simply return the provided data.
```
PabstMirror, line 7:
```suggestion
    If it is an Array, it will select a random entry from the array (nil if empty), otherwise it will simply return the provided data.
```

PabstMirror (approved):

PR #1759, opened 2025-05-12
Main - New Macros: Q, QQ

OverlordZorn:
Introduces a shortend version of QUOTE()

```hpp
#define Q(var1) QUOTE(var1)
#define QQ(var1) QUOTE(QUOTE(var1))
```

happy to add more when suggested


DartRuffian:
Why, just use QUOTE
I don't see why you'd need aliases for these

rautamiekka:
> Why, just use QUOTE I don't see why you'd need aliases for these

Like he said, they're shortened. I'd personally be happier with these shortened ones, since [I don't think] there's nothing else even remotely similarly named.

jonpas:
I am not a fan of this, single-letter macros are too obscuring and too easy to define in your own projects if you really want them. I think they do not fit CBA.

OverlordZorn:
> I am not a fan of this, single-letter macros are too obscuring and too easy to define in your own projects if you really want them. I think they do not fit CBA.

I'd like to argue that `Q` is already commonly known as its being used as a prefix: `QGVAR` - `GVAR`, `FUNC` - `QFUNC`, ...
If people know what `QGVAR` means, i doubt its hard to undstand what `Q` stands for as a standalone

jonpas:
Understanding is not the problem, single letter macro is. It is shortened in others for the purpose of not being so long, but `Q` and `QQ` is just crossing the line.

jokoho48:
> `QQUOTE` makes sense, ACE and many other mods (or at least my mods) define it. Adding it in CBA would make sense

We actually had the discussion and it seems like your template has the QQUOTE but not ace nor any of the other bigger mods

DartRuffian:
Ah right I mispoke, QQUOTE itself isn't defined but there is stuff like QQGVAR in ace

PR #1763, opened 2025-06-06
ItemContextMenu - New Function: removeItemContextMenuOption

OverlordZorn:
## New Function
  - fnc_removeItemContextMenuOption.sqf
    - will remove ContextMenuOption based on class and identifier.

## Modified Functions
  - fnc_addItemContextMenuOption.sqf
    - New optional Parameter: "_ident"
    - GVAR(ItemContextMenuOptions) is now a hashmap instead of a cba namespace.
    - Individual "options" are now stored as hashmap with the _ident as the key while previous data array is now the value.
    - Returns "_ident" on completion
  - fnc_openItemContextMenu.sqf
    - Adapted to handle hashmaps instead of the namespace and array.

### DIFF
--- a/addons/ui/fnc_addItemContextMenuOption.sqf
+++ b/addons/ui/fnc_addItemContextMenuOption.sqf
@@ -107,9 +107,14 @@ Parameters:
 
     _params                 - Arguments passed as '_this select 4' to condition and
                               statement (optional, default: []) <ANY>
+    
+    _ident                  - ID of the option. Can be used to remove an option.
+                              If none provided, generic ID will be generated.
+                              An already existing ID will overwrite the previous settings.
+                              (optional, default: "") <STRING>
 
 Returns:
-    Nothing/Undefined.
+    ID <STRING>.
 
 Examples:
     (begin example)
@@ -121,7 +126,7 @@ Examples:
     (end)
 
 Author:
-    commy2
+    commy2, Zorn
 ---------------------------------------------------------------------------- */
 
 // Force unscheduled environment to prevent race conditions.
@@ -133,7 +138,7 @@ if (!hasInterface) exitWith {};
 
 // Initialize system on first execution.
 if (isNil QGVAR(ItemContextMenuOptions)) then {
-    GVAR(ItemContextMenuOptions) = false call CBA_fnc_createNamespace;
+    GVAR(ItemContextMenuOptions) = createHashMap;
 
     ["CAManBase", "InventoryOpened", {
         params ["_unit", "_container1", "_container2"];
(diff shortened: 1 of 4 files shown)


addons/ui/fnc_removeItemContextMenuOption.sqf
```diff
@@ -0,0 +1,59 @@
...
+
+// Force unscheduled environment to prevent race conditions.
+if (canSuspend) exitWith {
+    [CBA_fnc_removeItemContextMenuOption, _this] call CBA_fnc_directCall;
+};
+
+if (!hasInterface) exitWith {};
+
+// Initialize system on first execution.
+if (isNil QGVAR(ItemContextMenuOptions)) exitWith { false };
```
DartRuffian: Should exit on `_id == ""` as well, removing an empty key should be invalid. Comment is also wrong, since nothing is initialized here

OverlordZorn: comment was leftover of a copy-paste, removed.

addons/ui/fnc_removeItemContextMenuOption.sqf
```diff
@@ -0,0 +1,59 @@
...
+
+// Force unscheduled environment to prevent race conditions.
+if (canSuspend) exitWith {
+    [CBA_fnc_removeItemContextMenuOption, _this] call CBA_fnc_directCall;
+};
+
+if (!hasInterface) exitWith {};
+
+if (_id == "") exitWith { false };
+if (isNil QGVAR(ItemContextMenuOptions)) exitWith { false };
```
DartRuffian: Two if statements is slower than one
```suggestion
if (isNil QGVAR(ItemContextMenuOptions) || _id == "") exitWith { false };
```

addons/ui/fnc_addItemContextMenuOption.sqf
```diff
@@ -222,11 +228,24 @@ _condition params [
...
+    _options = createHashMap;
+    GVAR(ItemContextMenuOptions) set [_item, _options];
 };
 
-_options pushBack [_slots, _displayName, _tooltip, _color, _icon, _conditionEnable, _conditionShow, _statement, _consume, _params];
+if (_id isEqualTo "") then {
+    private _keys = keys _options;
+    private _index = count _keys;
+
+    while { str _index in _keys } do { _index = _index + 1; };
```
johnb432: This is really inefficient (and likely prone to breaking). Imo better option would be to use a global counter instead, so you don't need to check the entire array of keys every time.

OverlordZorn: 3a6485e56211c7bb88cedce561fd10b57d9f0859

PR #1764, opened 2025-06-07
Common - Add `_skipAnim` parameter to `CBA_fnc_addItem` and related functions

DartRuffian:
- Add `_animate` parameter. If true (default) the unit will play an animation

### DIFF
--- a/addons/common/fnc_addItem.sqf
+++ b/addons/common/fnc_addItem.sqf
@@ -9,9 +9,10 @@ Description:
     case of trouble, or when able to add _item to _unit true in case of success.
 
 Parameters:
-    _unit   - the unit <OBJECT>
-    _item   - name of the weapon to add <STRING>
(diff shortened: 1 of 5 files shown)


addons/common/fnc_addItem.sqf
```diff
@@ -26,7 +27,7 @@ Author:
 ---------------------------------------------------------------------------- */
 SCRIPT(addItem);
 
-params [["_unit", objNull, [objNull]], ["_item", "", [""]], ["_verify", false, [false]]];
+params [["_unit", objNull, [objNull]], ["_item", "", [""]], ["_verify", false, [false]], ["_animate", true, [true]];
```
PabstMirror:
```suggestion
params [["_unit", objNull, [objNull]], ["_item", "", [""]], ["_verify", false, [false]], ["_animate", true, [true]]];
```

DartRuffian:
I in fact did not remember, thanks Pabst

PR #1766, opened 2025-06-15
Diagnostics - Add button to monitor watch expressions

BrettMayson:
image
image


addons/diagnostic/XEH_preInit.sqf
```diff
@@ -8,6 +8,10 @@ ADDON = false;
 #include "XEH_PREP.hpp"
 #include "initSettings.inc.sqf"
 
+GVAR(watchControls) = [controlNull, controlNull, controlNull, controlNull];
+GVAR(watchStatements) = ["","", "", ""];
+addMissionEventHandler ["Draw2D", {call FUNC(watchUpdate)}];
```
PabstMirror: move to the top of CBA_diagnostic_fnc_initTargetDebugConsole?
that way there is no code running for people who never access the debug console

BrettMayson: That function is called every time the pause menu is opened

PabstMirror:
```
if (isNil QGVAR(watchControls)) then {
  GVAR(watchControls) = ...
  addMissionEventHandler ["Draw2D"...
```

BrettMayson: Yeah that's what I did, I don't know why I commented that

I did put it in its own item in the extended event handler though, since it is unrelated to the target debug stuff

veteran29: From wiki for Draw2d, "Will stop executing when UI loses focus". Does it stop if game gets unfocused with `-noPause`? If so I would go for EachFrame.

addons/diagnostic/fnc_watchStop.sqf
```diff
@@ -0,0 +1,29 @@
...
+        [1] call CBA_diagnostic_fnc_watchStop;
+    (end)
+
+Author:
+    Brett Mayson
+---------------------------------------------------------------------------- */
+
+params ["_index"];
+
+private _ctrl = GVAR(watchControls) select _index;
```
veteran29: When you run `watchStart` for the first time it runs `watchStop`, `_ctrl ` will be nil in that case. Does not matter much in unscheduled as it will simply skip the `if` but I feel like it's cleaner with `param`.
```suggestion
private _ctrl = GVAR(watchControls) param [_index, controlNull];
```

addons/diagnostic/fnc_watchUpdate.sqf
```diff
@@ -0,0 +1,33 @@
...
+    if (isNull _x) then { continue };
+    private _statement = GVAR(watchStatements) select _forEachIndex;
+    if (_statement isEqualTo "") then {
+        [_forEachIndex] call CBA_diagnostic_fnc_watchStop;
+        continue;
+    };
+    (allControls _x) params ["", "_ctrlExpression", "_ctrlResult"];
+    _ctrlExpression ctrlSetText (_statement select 0);
+    _ctrlResult ctrlSetText format ["%1", (0 call (_statement select 1))];
+} forEach (GVAR(watchControls));
```
veteran29: Using select might be tiny bit faster due to SimpleVM optimizations
```suggestion
{
    private _statement = GVAR(watchStatements) select _forEachIndex;
    if (_statement isEqualTo "") then {
        [_forEachIndex] call CBA_diagnostic_fnc_watchStop;
        continue;
    };
    (allControls _x) params ["", "_ctrlExpression", "_ctrlResult"];
    _ctrlExpression ctrlSetText (_statement select 0);
    _ctrlResult ctrlSetText format ["%1", (0 call (_statement select 1))];
} forEach (GVAR(watchControls) select {!isNull _x});
```

addons/diagnostic/XEH_preInit.sqf
BrettMayson: Keeps running with noPause on borderless window, stops with exclusive fullscreen

PabstMirror:
they seem to get desynced if you start and stop
e.g. start 1, 2,
stop 2, start 4
ss

BrettMayson:
Wasn't able to reproduce that in single or multiplayer, using that order or just randomly quickly toggling them, never had an issue and it worked as expected

addons/diagnostic/fnc_watchUpdate.sqf
```diff
@@ -0,0 +1,32 @@
...
+{
+    private _statement = GVAR(watchStatements) select _forEachIndex;
+    if (_statement isEqualTo "") then {
+        [_forEachIndex] call CBA_diagnostic_fnc_watchStop;
+        continue;
+    };
+    (allControls _x) params ["", "_ctrlExpression", "_ctrlResult"];
+    _ctrlExpression ctrlSetText (_statement select 0);
+    _ctrlResult ctrlSetText format ["%1", (0 call (_statement select 1))];
+} forEach (GVAR(watchControls) select {!isNull _x});
```
PabstMirror:
```suggestion
} forEach GVAR(watchControls);
```

addons/diagnostic/fnc_watchUpdate.sqf
```diff
@@ -0,0 +1,32 @@
...
+    (begin example)
+        [1] call CBA_diagnostic_fnc_watchStart;
+    (end)
+
+Author:
+    Brett Mayson
+---------------------------------------------------------------------------- */
+
+{
+    private _statement = GVAR(watchStatements) select _forEachIndex;
```
PabstMirror, line 25:
```suggestion
    if (isNull _x) then { continue; };
    private _statement = GVAR(watchStatements) select _forEachIndex;
```

PabstMirror (approved):

PR #1775, opened 2025-06-27
Common - New FNC: `cba_fnc_getCfgDataHashmap`

OverlordZorn:
Add new Config Function to CBA.

`cba_fnc_getCfgDataHashmap` will return a class' properties as a hashmap.

## Each change in a separate line
- Add FNC
- Add FNC to CfgFunctions

### DIFF
--- a/addons/common/fnc_getCfgDataHashmap.sqf
+++ b/addons/common/fnc_getCfgDataHashmap.sqf
@@ -50,7 +50,9 @@ private _returnHashMap = createHashMap;
     if (_convert) then {
         _value = switch (_value) do {
             case "true": { true };
+            case "(true)": { true };
             case "false": { false };
+            case "(false)": { false };
             default { _value };
         };
     };


addons/common/fnc_getCfgDataHashmap.sqf
```diff
@@ -0,0 +1,65 @@
...
+private _returnHashMap = createHashMap;
+
+
+{
+    private _config = _x;
+    private _value = _x call BIS_fnc_getCfgData;
+
+    if (_convert) then {
+        _value = switch (_value) do {
+            case "true": { true };
```
PabstMirror, line 52: not sure if it should handle `condition = "(true)";`
or strings inside arrays
probably not a issue

PabstMirror (approved):

addons/common/fnc_getCfgDataHashmap.sqf
OverlordZorn: never saw sth like "(true)" in the wild but its very easy to add.

I can add something that itterates over arrays as well if required.

PabstMirror: e.g. https://github.com/CBATeam/CBA_A3/blob/master/addons/main/config.cpp#L63
used to be common because mikro would convert "true" to 1
tbh I wouldn't worry about either, seems like something the caller should handle if it's a problem

OverlordZorn: added the () variants

PR #1777, opened 2025-06-28
Events - Improve `CBA_fnc_globalEventJIP`: handle object as jipID

OverlordZorn:
Added handling for Objects as JIP:
=> Remove JIP once object is deleted

### DIFF
--- a/addons/events/fnc_globalEventJIP.sqf
+++ b/addons/events/fnc_globalEventJIP.sqf
@@ -7,11 +7,17 @@ Description:
 
     Event is put on a stack that is executed on every future JIP machine.
     Stack can be overwritten by using the same JIP-Stack-ID.
+    When object provided as jipID, the jip will get removed when object gets deleted.
 
 Parameters:
     _eventName - Type of event to publish. <STRING>
     _params    - Parameters to pass to the event handlers. <ANY>
-    _jipID     - Unique event ID. Can be used to remove or overwrite the event later. [optional] (default: create unique id) <STRING>
+    _jipID     - Unique event ID or object. [optional] (default: create unique id) <STRING, OBJECT>
+                 STRING:
+                    Custom ID - can be used to remove or overwrite the event later.
+                 OBJECT:
+                    Will remove JIP EH when object is deleted.
+        
 
 Returns:
     _jipID <STRING>
@@ -21,7 +27,12 @@ Author:
 ---------------------------------------------------------------------------- */
 SCRIPT(globalEventJIP);
 
-params [["_eventName", "", [""]], ["_params", []], ["_jipID", "", [""]]];
+params [["_eventName", "", [""]], ["_params", []], ["_jipID", "", ["", objNull]]];
+
+// store obj, reset jipID.
+private "_obj";
+if (_jipID isEqualType objNull) then { _obj = _jipID; _jipID = ""; };
+
 
 // generate string
 if (_jipID isEqualTo "") then {
@@ -39,5 +50,8 @@ GVAR(eventNamespaceJIP) setVariable [_jipID, [EVENT_PVAR_STR, [_eventName, _para
 
 // execute on every machine
 [QGVAR(eventJIP), [_eventName, _params]] call CBA_fnc_globalEvent;
+
+// Remove JIP once obj is deleted
+if (!isNil "_obj") then { [_jipID, _obj] call FUNC(removeGlobalEventJIP); };
 
 _jipID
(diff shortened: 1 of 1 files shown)


addons/events/fnc_globalEventJIP.sqf
```diff
@@ -40,4 +51,7 @@ GVAR(eventNamespaceJIP) setVariable [_jipID, [EVENT_PVAR_STR, [_eventName, _para
 // execute on every machine
 [QGVAR(eventJIP), [_eventName, _params]] call CBA_fnc_globalEvent;
 
+// Remove JIP once obj is deleted
+if (!isNil "_obj") then { [_jipID, _obj] call FUNC(removeGlobalEventJIP); };
```
johnb432: Have you checked if this actually works? I'm fairly certain that the function name expands to `CBA_fnc_events_removeGlobalEventJIP` and not `CBA_fnc_removeGlobalEventJIP`.

johnb432 (commented):
Not sure if having the `_jipID` serve two different purposes is a good idea: Imagine you only put `_obj` as parameter, when you meant to write `hashValue _obj` - makes debugging much harder.
Personally I think a new separate parameter is more suited for this.

addons/events/fnc_globalEventJIP.sqf
OverlordZorn: No i haven't. Basically went autopilot here 🙊. is there a macro for this? alternatively i'd just put the full fnc name here.

johnb432: I don't think there's a macro. In other functions they put the full function names.

OverlordZorn: found this
`#define FUNCMAIN(var1) TRIPLES(PREFIX,fnc,var1)`

addons/events/fnc_globalEventJIP.sqf
```diff
@@ -40,4 +55,7 @@ GVAR(eventNamespaceJIP) setVariable [_jipID, [EVENT_PVAR_STR, [_eventName, _para
 // execute on every machine
 [QGVAR(eventJIP), [_eventName, _params]] call CBA_fnc_globalEvent;
 
+// Remove JIP once obj is deleted
+if (!isNil "_obj") then { [_jipID, _obj] call FUNCMAIN(removeGlobalEventJIP); };
```
johnb432: Despite it being technically correct, I wouldn't use the `FUNCMAIN` macro as it's not used anywhere else in CBA.

addons/events/fnc_globalEventJIP.sqf
```diff
@@ -21,7 +27,16 @@ Author:
 ---------------------------------------------------------------------------- */
 SCRIPT(globalEventJIP);
 
-params [["_eventName", "", [""]], ["_params", []], ["_jipID", "", [""]]];
+params [["_eventName", "", [""]], ["_params", []], ["_jipOrObj", "", ["", objNull]]];
+
+private ["_obj", "_jipID"];
```
johnb432: Personally, I hate this way of declaring variables in sqf.

johnb432 (commented):
Just making sure you are aware: My reviews don't mean much in CBA, I'm only a contributor and not a maintainer.

OverlordZorn:
> Just making sure you are aware: My reviews don't mean much in CBA, I'm only a contributor and not a maintainer.

fair. just if anyone is interacting with the PR's i am very happy to adjust simply to ensure chances of successful merge.

addons/events/fnc_globalEventJIP.sqf
```diff
@@ -21,7 +27,16 @@ Author:
...
+if (_jipOrObj isEqualType "") then {
+    _jipID = _jipOrObj;
+} else {
+    _jipID = "";
+    _obj = _jipOrObj;
+};
```
PabstMirror:
```suggestion
(if (_jipOrObj isEqualType "") then {
    [_jipOrObj, nil]
} else {
    ["", _jipOrObj]
}) params ["_jipID", "_obj"];
```

addons/events/fnc_globalEventJIP.sqf
PabstMirror:
```suggestion
if (!isNil "_obj") then { [_jipID, _obj] call CBA_fnc_removeGlobalEventJIP };
```

addons/events/fnc_globalEventJIP.sqf
OverlordZorn: smort

PR #1779, opened 2025-06-29
Settings - Fix setting showing up in unicode subcategories

PabstMirror:
Before fix (Setting C1 showed up under a subcategory, instead of at the top by itself)
image

due to `sort` putting unicode above empty string
```
x = [["n", 1], ["й", 2], ["", 3]];
x sort true
[["й",2],["",3],["n",1]]
```


veteran29 (approved):
LGTM

PR #1783, opened 2025-07-26
Common - Add more loadout randomization options

DartRuffian:
- Add more loadout randomization options


addons/common/fnc_randomizeLoadout.sqf
```diff
@@ -0,0 +1,137 @@
...
+// Disabled conditions
+if (!local _unit) exitWith {true};
+
+private _randomizationDisabled = getArray (missionConfigFile >> "disableRandomization") findIf {
+    _unit isKindOf _x || {(vehicleVarName _unit) isEqualTo _x}
+} != -1;
+
+if (_randomizationDisabled || {!(_unit getVariable ["BIS_enableRandomization", true])}) exitWith {true};
+
+if (isNull _unit || !(_unit getVariable ["BIS_enableRandomization", true])) exitWith {};
```
rautamiekka: Why check the unit's BIS rando' twice ?

DartRuffian: Originally had this code outside of CBA, forgot to remove the checks from that

addons/common/fnc_fixAnimation3DEN.sqf
DartRuffian, line 1: I'm not exactly sure what causes the animation to be wrong (e.g. in the rifle holding anim when the unit only has a pistol). I don't remember ever seeing it before working on this pr though, so I added a workaround for it.

It seems to need to run on the next frame, but CBA's frame functions don't work in Eden so I had to use a sleep with a small delay.

addons/common/XEH_init3DEN.sqf
```diff
@@ -0,0 +1,20 @@
+#include "script_component.hpp"
+
+add3DENEventHandler ["OnEditableEntityAdded", {
+    params ["_entity"];
+    if (typeName _entity != "OBJECT") exitWith {};
```
rautamiekka: Is `typeName` really needed instead of `isEqualType objNull` ?
```suggestion
    if !(_entity isEqualType objNull) exitWith {};
```

addons/common/fnc_fixAnimation3DEN.sqf
```diff
@@ -0,0 +1,36 @@
...
+TRACE_1("fnc_fixAnimation",_unit);
+
+if (!is3DEN) exitWith {};
+
+// Sometimes units have the wrong animation when placed, not sure why
+_unit spawn {
+    sleep 0.1;
+    private _animation = switch (false) do {
+        case (primaryWeapon _this == ""): { "amovpercmstpsraswrfldnon" };
+        case (handgunWeapon _this == ""): { "amovpercmstpsraswpstdnon" };
```
rautamiekka:
```suggestion
        case (primaryWeapon _this isEqualTo ""): { "amovpercmstpsraswrfldnon" };
        case (handgunWeapon _this isEqualTo ""): { "amovpercmstpsraswpstdnon" };
```

addons/common/fnc_randomizeLoadout.sqf
```diff
@@ -0,0 +1,133 @@
...
+    WARNING_1("Unit [%1] is null",_unit);
+    false
+};
+
+// Disabled conditions
+if (!local _unit) exitWith {true};
+
+private _randomizationDisabled = getArray (missionConfigFile >> "disableRandomization") findIf {
+    _unit isKindOf _x || {(vehicleVarName _unit) isEqualTo _x}
+} != -1;
```
rautamiekka, line 56:
```suggestion
} isNotEqualTo -1;
```

addons/common/fnc_randomizeLoadout.sqf
```diff
@@ -0,0 +1,133 @@
...
+};
+
+// Set loadout and then add weapons to avoid issues with conflicting weapon items
+[_unit, [_loadout, _extendedInfo]] call CBA_fnc_setLoadout;
+
+if (_binocularList isNotEqualTo []) then {
+    private _item = selectRandomWeighted _binocularList;
+    private _magazine = (_item call CBA_fnc_compatibleMagazines) param [0, ""]; // For laser designators
+    _unit addWeapon _item;
+    if (_magazine != "") then {
```
rautamiekka, line 105:
```suggestion
    if (_magazine isNotEqualTo "") then {
```

addons/common/fnc_setIdentity3DEN.sqf
```diff
@@ -0,0 +1,51 @@
+#include "script_component.hpp"
+/* ----------------------------------------------------------------------------
+Function: CBA_fnc_setIdentity3DEN
+Description:
+    Sets Eden attributes relating to the given identity.
+
+Parameters:
+    _unit  - Unit <OBJECT>
```
rautamiekka:
```suggestion
    _unit      - Unit <OBJECT>
```

addons/common/fnc_setIdentity3DEN.sqf
```diff
@@ -0,0 +1,51 @@
...
+    (end)
+
+Author:
+    DartRuffian
+---------------------------------------------------------------------------- */
+
+params ["_unit", ["_identity", ""]];
+TRACE_2("fnc_setIdentity3DEN",_unity,_identity);
+
+if (_identity == "") then {
```
rautamiekka, line 26:
```suggestion
if (_identity isEqualTo "") then {
```

addons/common/fnc_setIdentity3DEN.sqf
```diff
@@ -0,0 +1,51 @@
...
+    _identity = getText (configOf _unit >> "CBA_identity");
+};
+
+if (_identity == "") exitWith {};
```
rautamiekka, line 30:
```suggestion
if (_identity isEqualTo "") exitWith {};
```

addons/common/fnc_setIdentity3DEN.sqf
```diff
@@ -0,0 +1,51 @@
...
+private _pitch = getNumber (_identityConfig >> "pitch");
+private _speaker = getText (_identityConfig >> "speaker");
+
+_unit set3DENAttribute ["unitName", _name];
+_unit set3DENAttribute ["NameSound", _nameSound];
+_unit set3DENAttribute ["face", _face];
+_unit set3DENAttribute ["pitch", _pitch];
+_unit set3DENAttribute ["speaker", _speaker];
+
+if (_glasses != "") then {
```
rautamiekka, line 48:
```suggestion
if (_glasses isNotEqualTo "") then {
```

rautamiekka (changes requested):

addons/common/fnc_fixAnimation3DEN.sqf
DartRuffian: That's doing the same thing, there's literally no difference there. You can't even argue type safety because it's a return from an engine command.

addons/common/XEH_postInit.sqf
```diff
@@ -47,3 +47,7 @@ for "_i" from 0 to (count _config - 1) do {
     _oldVeh setVehicleVarName "";
     _newVeh setVehicleVarName _vehName;
 }] call CBA_fnc_addEventHandler;
+
+// Check missionTime so Eden placed units don't have their loadouts randomized twice
+["CAManBase", "InitPost", { if (CBA_missionTime > 0.1) then { call CBA_fnc_randomizeLoadout } }] call CBA_fnc_addClassEventHandler;
```
DartRuffian, line 52: Not sure if there's a more proper way to do this

addons/common/fnc_getRandomizedEquipment.sqf
```diff
@@ -0,0 +1,54 @@
...
+    private _nvgList = getArray (_unitConfig >> "CBA_nvgList");
+
+    // If all arrays are empty, just cache `[false]` to not save a bunch of empty arrays
+    if (
+        [
+            _primaryList, _launcherList, _handgunList,
+            _uniformList, _vestList, _backpackList,
+            _headgearList, _facewearList, _binocularList,
+            _nvgList
+        ] findIf { _x isEqualTo [] } >= -1
```
PabstMirror: findIf will always be >= -1

addons/common/XEH_postInit.sqf
PabstMirror: xeh_postInit won't run on 3den (for now at least)

DartRuffian: The check is so units spawned in Eden don't have their kits randomized again, but units spawned later (e.g. script / zeus) do

Without the missionTime check Eden placed units would have their loadouts randomized on mission start

addons/common/fnc_getRandomizedEquipment.sqf
DartRuffian: Ah right, typo, just > should be fine I think

At work currently

addons/common/fnc_randomizeLoadout.sqf
```diff
@@ -0,0 +1,147 @@
...
+    // CBA's frame functions don't work in Eden
+    [_unit, _fnc_fixAnimation] spawn {
+        sleep 0.1;
+        params ["_unit", "_function"];
+        save3DENInventory [get3DENEntityID _unit];
+        _unit call _function;
+    };
+} else {
+    [_fnc_fixAnimation, _unit] call CBA_fnc_execNextFrame;
+};
```
DartRuffian: Causes an issue when creating editorPreviews for units, since they are no longer in the "relaxed" rifle pose. I have this PR in my unit mod for testing and not sure the best way to fix this. Currently I just comment out the lines when making previews

addons/common/fnc_randomizeLoadout.sqf
```diff
@@ -0,0 +1,148 @@
...
+        _this switchMove (["amovpercmstpsraswrfldnon", "amovpercmstpsraswpstdnon", "amovpercmstpsnonwnondnon"] select _weaponIndex);
+        _this selectWeapon (_weapons select _weaponIndex);
+    };
+};
+
+if (is3DEN) then {
+    // CBA's frame functions don't work in Eden
+    // And we need a decently sized delay to not mess with creating editorPreviews
+    [_unit, _fnc_fixAnimation] spawn {
+        sleep 1.2;
```
rautamiekka: I wonder if `uiSleep` would be a better idea, since it'd guarantee the sleep lasts as specified ?

DartRuffian: With some more testing the issue I was running into with it doesn't seem to be happening any more. I did some testing, got 1.2 working well, then spawned them in eden to see how weird it looked and they had the correct animations immediately.

Still not sure what was causing it before though

DartRuffian:
Been over a year since this was opened, could some take a look again?

DartRuffian:
@PabstMirror

addons/common/XEH_postInit.sqf
DartRuffian: ToDo: Should just be called from CuratorObjectPlaced eh

addons/common/XEH_preInit.sqf
```diff
@@ -74,8 +74,26 @@ activateAddons GVAR(addons);
     }];
 }] call CBA_fnc_addClassEventHandler;
 
-// Facewear randomization
-["CAManBase", "InitPost", CBA_fnc_randomizeFacewear] call CBA_fnc_addClassEventHandler;
+// Loadout randomization
+GVAR(randomLoadoutUnits) = createHashMap;
+if (is3DEN) then {
+    add3DENEventHandler ["OnEditableEntityAdded", {
```
PabstMirror, line 80: this will only work for new missions

take existing mission and `CBA_facewearList[]` gets ignored

because the unit wasn't placed in 3den
and the `InitPost` runs at mission time 0

addons/common/fnc_randomizeLoadout.sqf
```diff
@@ -0,0 +1,133 @@
...
+    _primaryList,
+    _launcherList,
+    _handgunList
+];
+
+if (is3DEN) then {
+    // CBA's frame functions don't work in Eden
+    _unit spawn {
+        sleep 0.01;
+        save3DENInventory [get3DENEntityID _this];
```
PabstMirror, line 129: I don't think we have any way to tell if this has been run,
so it's probably best to just not touch 3den loadout and just run in xeh

DartRuffian: It you randomize at mission start you lose any changes made in efen

addons/common/XEH_preInit.sqf
DartRuffian: I feel like that's probably fine. Could add a special case where units that only have facewear list get randomized at mission start too but I feel like that'd cause more unexpected behavior for people editing kits

PR #1793, opened 2025-11-09
pid - Add PID controller and associated utility functions

TheCandianVendingMachine:
- Add PID controller interface
- Taylor series for approximating derivative up to degree 2
- Integral windup prevention via integrating over short history
- Function to update PID gains of existing controller
- Function to reset PID history
- Function to update PID setpoint
- Error function for linear cases
- Error function for dealing with degree wrap-around

I was working on another PID controller for a project outside of ACE that needed 6 controllers, and after I wrote one of the PID updates wrong I realised that this would be a good fit for CBA. This PID controller can replace existing implementations in ACE missileguidance, but also would work well with an AI mod I am assisting with. I think that a PID controller is useful to a variety of mods: it can implement target tracking, smooth animations, speed control; pretty much anything that requires tracking a setpoint

I copied structure from `CBA_statemachine`. If there is any structural errors, that's ignorance on my part

### DIFF
--- a/addons/pid/fnc_create.sqf
+++ b/addons/pid/fnc_create.sqf
@@ -28,7 +28,7 @@ Returns:
 
 Examples:
     (begin example)
-        private _pid = [1, 0.05, 0.2, 5, CBA_pid_fnc_error_linear, 0, 10, 15] call CBA_pid_fnc_create;
+        private _pid = [1, 0.05, 0.2, 5, 0, 10, 15, CBA_pid_fnc_error_linear] call CBA_pid_fnc_create;
     (end)
 
 Author:
--- a/addons/pid/fnc_create.sqf
+++ b/addons/pid/fnc_create.sqf
@@ -7,13 +7,21 @@ Description:
 
 Parameters:
     _pGain          - gain for the immediate error <NUMBER>
+                      (Default: 0)
     _iGain          - gain for the persistent error over time <NUMBER>
+                      (Default: 0)
     _dGain          - gain for the error rate over time <NUMBER>
+                      (Default: 0)
     _setpoint       - initial setpoint for the controller <NUMBER>
+                      (Default: 0)
     _errorFunction  - the function which calculates the error which the controller operators <CODE>
+                      (Default: <CBA_pid_fnc_error_linear>)
     _min            - the minimum value the controller can return <NUMBER>
+                      (Default: -1e99)
     _max            - the maximum value the controller can return <NUMBER>
+                      (Default: 1e99)
     _historyLength  - how many past errors are stored and used to calculate the derivative/integral <NUMBER>
+                      (Default: 10)
 
 Returns:
     _pid            - a PID controller <LOCATION>
@@ -32,10 +40,10 @@ params [
     ["_iGain", 0, [0]],
     ["_dGain", 0, [0]],
     ["_setpoint", 0, [0]],
-    ["_errorFunction", FUNC(error_linear), [{0}]],
+    ["_errorFunction", FUNC(error_linear), [{}]],
     ["_min", -1e99, [0]],
     ["_max",  1e99, [0]],
-    ["_historyLength", 3, [0]],
+    ["_historyLength", 10, [0]],
 ];
 
 private _pid = call CBA_fnc_createNamespace;
--- a/addons/pid/fnc_create.sqf
+++ b/addons/pid/fnc_create.sqf
@@ -41,9 +41,9 @@ params [
     ["_dGain", 0, [0]],
     ["_setpoint", 0, [0]],
     ["_errorFunction", FUNC(error_linear), [{}]],
-    ["_min", -1e38, [0]],
-    ["_max",  1e38, [0]],
-    ["_historyLength", 10, [0]]
+    ["_min", -LARGE_NUMBER, [0]],
+    ["_max",  LARGE_NUMBER, [0]],
+    ["_historyLength", DEFAULT_HISTORY_LENGTH, [0]]
 ];
 
 private _pid = call CBA_fnc_createNamespace;
--- a/addons/pid/fnc_create.sqf
+++ b/addons/pid/fnc_create.sqf
@@ -42,8 +42,8 @@ params [
     ["_setpoint", 0, [0]],
     ["_min", -LARGE_NUMBER, [0]],
     ["_max",  LARGE_NUMBER, [0]],
-    ["_errorFunction", FUNC(error_linear), [{}]],
-    ["_historyLength", DEFAULT_HISTORY_LENGTH, [0]]
+    ["_historyLength", DEFAULT_HISTORY_LENGTH, [0]],
+    ["_errorFunction", FUNC(error_linear), [{}]]
 ];
 
 private _pid = call CBA_fnc_createNamespace;
--- a/addons/pid/fnc_create.sqf
+++ b/addons/pid/fnc_create.sqf
@@ -34,7 +34,7 @@ params [
     ["_setpoint", 0, [0]],
     ["_errorFunction", FUNC(error_linear), [{0}]],
     ["_min", -1e99, [0]],
-    ["_max", -1e99, [0]],
+    ["_max",  1e99, [0]],
     ["_historyLength", 3, [0]],
 ];
 
--- a/addons/pid/fnc_create.sqf
+++ b/addons/pid/fnc_create.sqf
@@ -1,0 +1,49 @@
+#include "script_component.hpp"
+/* ----------------------------------------------------------------------------
+Function: CBA_pid_fnc_create
+
+Description:
+    Creates a PID controller.
+
+Parameters:
+    _pGain          - gain for the immediate error <NUMBER>
+    _iGain          - gain for the persistent error over time <NUMBER>
+    _dGain          - gain for the error rate over time <NUMBER>
+    _setpoint       - initial setpoint for the controller <NUMBER>
+    _errorFunction  - the function which calculates the error which the controller operators <CODE>
+    _min            - the minimum value the controller can return <NUMBER>
+    _max            - the maximum value the controller can return <NUMBER>
+    _historyLength  - how many past errors are stored and used to calculate the derivative/integral <NUMBER>
+
+Returns:
+    _pid            - a PID controller <LOCATION>
+
+Examples:
+    (begin example)
+        private _pid = [1, 0.05, 0.2, 5, CBA_pid_fnc_error_linear, 0, 10, 15] call CBA_pid_fnc_create;
+    (end)
+
+Author:
+    tcvm
+---------------------------------------------------------------------------- */
+SCRIPT(create);
+params [
+    ["_pGain", 0, [0]],
+    ["_iGain", 0, [0]],
+    ["_dGain", 0, [0]],
+    ["_setpoint", 0, [0]],
+    ["_errorFunction", FUNC(error_linear), [{0}]],
+    ["_min", -1e99, [0]],
+    ["_max", -1e99, [0]],
+    ["_historyLength", 3, [0]],
+];
+
+private _pid = call CBA_fnc_createNamespace;
+_pid setVariable [QGVAR(gains), [_pGain, _iGain, _dGain]];
+_pid setVariable [QGVAR(bounds), [_min, _max]];
+_pid setVariable [QGVAR(history), []];
+_pid setVariable [QGVAR(historyLength), _historyLength];
+_pid setVariable [QGVAR(setpoint), _setpoint];
+_pid setVariable [QGVAR(errorFunction), _errorFunction];
+
+_pid
--- a/addons/pid/fnc_create.sqf
+++ b/addons/pid/fnc_create.sqf
@@ -43,7 +43,7 @@ params [
     ["_errorFunction", FUNC(error_linear), [{}]],
     ["_min", -1e99, [0]],
     ["_max",  1e99, [0]],
-    ["_historyLength", 10, [0]],
+    ["_historyLength", 10, [0]]
 ];
 
 private _pid = call CBA_fnc_createNamespace;
--- a/addons/pid/fnc_create.sqf
+++ b/addons/pid/fnc_create.sqf
@@ -14,12 +14,12 @@ Parameters:
                       (Default: 0)
     _setpoint       - initial setpoint for the controller <NUMBER>
                       (Default: 0)
+    _min            - the minimum value the controller can return <NUMBER>
+                      (Default: -1e30)
+    _max            - the maximum value the controller can return <NUMBER>
+                      (Default: 1e30)
     _errorFunction  - the function which calculates the error which the controller operators <CODE>
                       (Default: <CBA_pid_fnc_error_linear>)
-    _min            - the minimum value the controller can return <NUMBER>
-                      (Default: -1e99)
-    _max            - the maximum value the controller can return <NUMBER>
-                      (Default: 1e99)
     _historyLength  - how many past errors are stored and used to calculate the derivative/integral <NUMBER>
                       (Default: 10)
 
@@ -40,9 +40,9 @@ params [
     ["_iGain", 0, [0]],
     ["_dGain", 0, [0]],
     ["_setpoint", 0, [0]],
-    ["_errorFunction", FUNC(error_linear), [{}]],
     ["_min", -LARGE_NUMBER, [0]],
     ["_max",  LARGE_NUMBER, [0]],
+    ["_errorFunction", FUNC(error_linear), [{}]],
     ["_historyLength", DEFAULT_HISTORY_LENGTH, [0]]
 ];
 
--- a/addons/pid/fnc_create.sqf
+++ b/addons/pid/fnc_create.sqf
@@ -41,8 +41,8 @@ params [
     ["_dGain", 0, [0]],
     ["_setpoint", 0, [0]],
     ["_errorFunction", FUNC(error_linear), [{}]],
-    ["_min", -1e99, [0]],
-    ["_max",  1e99, [0]],
+    ["_min", -1e38, [0]],
+    ["_max",  1e38, [0]],
     ["_historyLength", 10, [0]]
 ];
 
--- a/addons/pid/fnc_create.sqf
+++ b/addons/pid/fnc_create.sqf
@@ -18,10 +18,10 @@ Parameters:
                       (Default: -1e30)
     _max            - the maximum value the controller can return <NUMBER>
                       (Default: 1e30)
-    _errorFunction  - the function which calculates the error which the controller operators <CODE>
-                      (Default: <CBA_pid_fnc_error_linear>)
     _historyLength  - how many past errors are stored and used to calculate the derivative/integral <NUMBER>
                       (Default: 10)
+    _errorFunction  - the function which calculates the error which the controller operators <CODE>
+                      (Default: <CBA_pid_fnc_error_linear>)
 
 Returns:
     _pid            - a PID controller <LOCATION>
--- a/addons/pid/fnc_setGains.sqf
+++ b/addons/pid/fnc_setGains.sqf
@@ -1,0 +1,50 @@
+#include "script_component.hpp"
+/* ----------------------------------------------------------------------------
+Function: CBA_pid_fnc_setGains
+
+Description:
+    Set new gains for a PID controller.
+
+Parameters:
+    _pid        - a pid controller <LOCATION>
+    _pGain      - the new proportional gain <NUMBER>
+                  OR
+                  `nil`, to keep the current gain <NONE>
+    _iGain      - the new integral gain <NUMBER>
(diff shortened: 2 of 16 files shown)


PabstMirror:
hemtt has 2 `WARN Skipping Optimization because NaN [U:-] (- inf) => -inf`
not sure where it is, and it might even be valid, but it seems like something weird is happening

addons/pid/fnc_create.sqf
```diff
@@ -0,0 +1,57 @@
...
+    tcvm
+---------------------------------------------------------------------------- */
+SCRIPT(create);
+params [
+    ["_pGain", 0, [0]],
+    ["_iGain", 0, [0]],
+    ["_dGain", 0, [0]],
+    ["_setpoint", 0, [0]],
+    ["_errorFunction", FUNC(error_linear), [{}]],
+    ["_min", -1e99, [0]],
```
PabstMirror:
```suggestion
    ["_min", -1e38; [0]],
```

TheCandianVendingMachine: Am I insane or did Arma used to support `1e99`?

PabstMirror: I think the game is fine with it in normal sqf
but we had to change for sqfc https://github.com/CBATeam/CBA_A3/pull/1618 (this was back on ASC)
`-1e99` is actually `(negative operator) and (1e99)`
part of sqfc optimization is evaluating these into a single number at compile time
and it uses floats and `-1e99` becomes NaN, which it doesn't want to use

(I could be getting something wrong but I think it's close to that)

rautamiekka:
What is PID ?

addons/pid/fnc_setGains.sqf
```diff
@@ -0,0 +1,53 @@
...
+                  `nil`, to keep the current gain <NIL>
+                  (Default: `nil`)
+    _iGain      - the new integral gain <NUMBER>
+                  OR
+                  `nil`, to keep the current gain <NIL>
+                  (Default: `nil`)
+    _dGain      - the new derivative gain <NUMBER>
+                  OR
+                  `nil`, to keep the current gain <NIL>
+                  (Default: `nil`)
```
PabstMirror:
```suggestion
    _pGain      - the new proportional gain <NUMBER><NIL> (Default: `nil`)
                  OR
                  `nil`, to keep the current gain
    _iGain      - the new integral gain <NUMBER><NIL> (Default: `nil`)
                  OR
                  `nil`, to keep the current gain 
    _dGain      - the new derivative gain <NUMBER><NIL> (Default: `nil`)
                  OR
                  `nil`, to keep the current gain
```
I'm not sure if this is more clunky wording, but it's much easier to parse if the types are on the first line

addons/pid/fnc_setGains.sqf
```diff
@@ -0,0 +1,53 @@
...
+---------------------------------------------------------------------------- */
+SCRIPT(setGains);
+params [
+    ["_pid", locationNull, [locationNull]],
+    ["_pGain", nil, [0, nil]],
+    ["_iGain", nil, [0, nil]],
+    ["_dGain", nil, [0, nil]]
+];
+
+if (isNull _pid) exitWith {};
```
PabstMirror, line 38:
```suggestion
if (isNull _pid) exitWith { 0 };
```

PabstMirror (approved):

addons/pid/fnc_setGains.sqf
TheCandianVendingMachine: why this change? is it just to indicate a failure? `setVariable` shouldn't return anything as it is

PabstMirror: opps, wrong func

addons/pid/fnc_update.sqf
```diff
@@ -0,0 +1,97 @@
...
+    tcvm
+---------------------------------------------------------------------------- */
+SCRIPT(update);
+params [
+    ["_pid", locationNull, [locationNull]],
+    ["_value", 0, [0]],
+    ["_time", CBA_missionTime, [0]]
+];
+
+if (isNull _pid) exitWith {};
```
PabstMirror:
```suggestion
if (isNull _pid) exitWith { 0 };
```
this or add <NIL> to returns

TheCandianVendingMachine:
Okay, should be good to go. Here is a video showcasing the PID in action with a simple roll controller.

For context, the only variable we are controlling for this rotation is the current inputted roll acceleration of the plane. This is using a cascading PID controller, and the end PID is controlling _only_ the inputted roll acceleration, there is nothing else to it.

https://youtu.be/NytSU4p3qYY

The code which controls this is simply
```sqf
private _velocityCommand = [_pid_offset, _roll] call CBA_pid_fnc_update;
[_pid_velocity, _velocityCommand] call CBA_pid_fnc_setpoint;
private _accelerationCommand = [_pid_velocity, _rollVelocity] call CBA_pid_fnc_update;

_rollAcceleration = _accelerationCommand;
_rollVelocity = _rollVelocity + _rollAcceleration * _deltaTime;
```

The entire example code is 
```sqf
sleep 5;
ROLL_ACCELERATION = 30;

rollOffset_pid = [1, 0.3, 0.0, 0, CBA_pid_fnc_error_degree, -180, 180, 16] call CBA_pid_fnc_create;
rollAcceleration_pid = [1, 0.05, 0.3, 0, CBA_pid_fnc_error_linear, -ROLL_ACCELERATION, ROLL_ACCELERATION, 32] call CBA_pid_fnc_create;

[{
    params ["_args"];
    _args params ["_pid_offset", "_pid_velocity", "_plane", "_rollVelocity", "_rollAcceleration", "_lastUpdate", "_lastPrint"];
    private _deltaTime = CBA_missionTime - _lastUpdate;
    private _vDir = vectorDirVisual _plane;
    private _vUp = vectorUpVisual _plane;
    private _vRight = _vDir vectorCrossProduct _vUp;
    private _starboard = [
        (_vDir select 1),
        -(_vDir select 0),
        _vDir select 2
    ];

    private _roll = 90 - acos (_vUp vectorCos _starboard);

    private _pointingDown = (_vUp vectorDotProduct [0, 0, 1]) < 0;
    private _pointingRight = (_vUp vectorDotProduct _starboard) > 0;

    if (_pointingDown && _pointingRight) then {
        _roll = 180 - _roll;
    };
    if (_pointingDown && !_pointingRight) then {
        _roll = -180 - _roll;
    };

    private _velocityCommand = [_pid_offset, _roll] call CBA_pid_fnc_update;
    [_pid_velocity, _velocityCommand] call CBA_pid_fnc_setpoint;
    private _accelerationCommand = [_pid_velocity, _rollVelocity] call CBA_pid_fnc_update;

    _rollAcceleration = _accelerationCommand;
    _rollVelocity = _rollVelocity + _rollAcceleration * _deltaTime;

    private _theta = _rollVelocity * _deltaTime;
    _plane setVectorUp ([_vUp, _vDir, _theta] call CBA_fnc_vectRotate3D);

    _args set [3, _rollVelocity];
    _args set [4, _rollAcceleration];
    _args set [5, CBA_missionTime];
    
    if (CBA_missionTime - _lastPrint > 0.2) then {
        systemChat format ["Roll: %1", _roll];
        _args set [6, CBA_missionTime];
    };
}, 0, [rollOffset_pid, rollAcceleration_pid, plane, 0, 0, CBA_missionTime, CBA_missionTime]] call CBA_fnc_addPerFrameHandler;

[{
    [rollOffset_pid, 15] call CBA_pid_fnc_setpoint;
    systemChat "15 degrees";
}, nil, 10] call CBA_fnc_waitAndExecute;
[{
    [rollOffset_pid, -80] call CBA_pid_fnc_setpoint;
    systemChat "-80 degrees";
}, nil, 15] call CBA_fnc_waitAndExecute;
[{
    [rollOffset_pid, -150] call CBA_pid_fnc_setpoint;
    systemChat "-150 degrees";
}, nil, 25] call CBA_fnc_waitAndExecute;
[{
    [rollOffset_pid, 160] call CBA_pid_fnc_setpoint;
    systemChat "160 degrees";
}, nil, 35] call CBA_fnc_waitAndExecute;
[{
    [rollOffset_pid, 0] call CBA_pid_fnc_setpoint;
    systemChat "0 degrees";
}, nil, 45] call CBA_fnc_waitAndExecute;
```

addons/pid/fnc_create.sqf
```diff
@@ -0,0 +1,57 @@
...
+                      (Default: 0)
+    _dGain          - gain for the error rate over time <NUMBER>
+                      (Default: 0)
+    _setpoint       - initial setpoint for the controller <NUMBER>
+                      (Default: 0)
+    _min            - the minimum value the controller can return <NUMBER>
+                      (Default: -1e30)
+    _max            - the maximum value the controller can return <NUMBER>
+                      (Default: 1e30)
+    _errorFunction  - the function which calculates the error which the controller operators <CODE>
```
PabstMirror: error and history are different order from params

addons/pid/fnc_create.sqf
```diff
@@ -0,0 +1,57 @@
...
+                      (Default: <CBA_pid_fnc_error_linear>)
+    _historyLength  - how many past errors are stored and used to calculate the derivative/integral <NUMBER>
+                      (Default: 10)
+
+Returns:
+    _pid            - a PID controller <LOCATION>
+
+Examples:
+    (begin example)
+        private _pid = [1, 0.05, 0.2, 5, CBA_pid_fnc_error_linear, 0, 10, 15] call CBA_pid_fnc_create;
```
PabstMirror:
```suggestion
        private _pid = [1, 0.05, 0.2, 5, 0, 10, 15, CBA_pid_fnc_error_linear] call CBA_pid_fnc_create;
```

addons/pid/fnc_update.sqf
```diff
@@ -0,0 +1,103 @@
...
+};
+
+if (count _history > _maxHistoryLength) then {
+    _history deleteAt 0;
+};
+_pid setVariable [QGVAR(history), _history];
+
+private _derivative = 0;
+switch (true) do {
+    case (count _history == 2): {
```
TheMagnetar, line 56: why do you do that? This is quite strange. There are better ways such as error on measurement etc. Moreover, if you want to prevent dividing by zero check on elapsed time before calculating. In that way you a) prevent dividing by zero, b) prevent updating the PID too fast

addons/pid/fnc_update.sqf
```diff
@@ -0,0 +1,103 @@
...
+private _tn_prev = 0;
+{
+    _x params ["_tn", "_error"];
+    if (_forEachIndex > 0) then {
+        // Trapezoidal rule for numerical integration
+        // (1 / 2) * (b - a) * (f(a) + f(b))
+        // O(h^2) error
+        private _stride = _tn - _tn_prev;
+        private _sum = _error + _error_prev;
+        _integral = _integral + 0.5 * _stride * _sum;
```
TheMagnetar, line 91: There is no need for trapezoidal rule. Additionally this method of calculating the integral prevents from updating the PID parameters.

addons/pid/fnc_update.sqf
```diff
@@ -0,0 +1,103 @@
...
+    };
+    _tn_prev = _tn;
+    _error_prev = _error;
+} forEach _history;
+
+(_pid getVariable [QGVAR(gains), [0, 0, 0]]) params ["_pGain", "_iGain", "_dGain"];
+
+private _delta = _error * _pGain + _integral * _iGain + _derivative * _dGain;
```
TheMagnetar, line 99: You may want to add a method to limit the windup effect (integral part of the PID).

TheCandianVendingMachine: Windup is limited through a small history window. The designer can set the window size, and windup is necessarily prevented by out-of-date values being discarded quickly, meaning you will only ever have relevant error.

I’ve found that, practically, this works to eliminate steady state error as long as the window is large enough. Windup does happen, but it doesn’t cause too much instability in the system

addons/pid/fnc_update.sqf
TheCandianVendingMachine: How does this prevent the parameters from being updated? And although this method of estimating the integral is not needed, it isn’t meaningfully less efficient than a Reimann sum to change. It has a better approximation error, and isn’t meaningfully less efficient due to instructions

addons/pid/fnc_update.sqf
TheCandianVendingMachine: You haven’t read the entire function, this is a fallback to avoid any known unknown DIV0 errors due to some weirdness with floating point. I enforce a monotonic testing variable at the start of this file

Likewise, why do I do what, exactly?

addons/pid/fnc_update.sqf
```diff
@@ -0,0 +1,103 @@
...
+_pid setVariable [QGVAR(history), _history];
+
+private _derivative = 0;
+switch (true) do {
+    case (count _history == 2): {
+        private _x0 = _history select 0;
+        private _x1 = _history select 1;
+
+        private _stride = (_x1 select 0) - (_x0 select 0);
+        if (_stride == 0) then { break };
```
PabstMirror:
```suggestion
        if (_stride == 0) exitWith {};
```
not in a loop so the `break` may hit something unexpected
example
```
[{
    systemChat "A"; if (true) then { break }; systemChat "unreachable";
}] call CBA_fnc_addPerFrameHandler;
[{
    systemChat "B"; // also never reached because `break` hit some upper pfeh-internal loop
}] call CBA_fnc_addPerFrameHandler;
```

addons/pid/fnc_update.sqf
```diff
@@ -0,0 +1,103 @@
...
+        _derivative = ((_x1 select 1) - (_x0 select 1)) / _stride;
+    };
+    case (count _history >= 3): {
+        private _xn0 = _history select -1;
+        private _xn1 = _history select -2;
+        private _xn2 = _history select -3;
+        private _stride0 = (_xn0 select 0) - (_xn1 select 0);
+        private _stride1 = (_xn1 select 0) - (_xn2 select 0);
+        private _stride = _stride0 + _stride1;
+        if (_stride == 0) then { break };
```
PabstMirror:
```suggestion
        if (_stride == 0) exitWith {};
```

addons/pid/fnc_update.sqf
TheCandianVendingMachine: i was thinking about this all day, and here is a proof that windup is not an issue

Integral windup is caused by large initial error being kept in the integral sum, and then not being lost until about the same amount of error is overshot when we reach the set point. 

With the sliding window implemented here, old error is disregarded. Windup still occurs, but initially it acts as a scalar for the proportional term, and is quickly forgotten as time moves on. Observe the function 

image

If we were to take the sum of the error, we would have a lot of windup. This is the integral of `[0, t]`. Now consider `t_win` which is starting at the blue marker. If we take the integral from `[t_win, t]`, we end up with a sum on the function 

image

Now all of the error in the integral is steady state only. This is precisely what the integral term is for. An issue can occur where the total sum of the error is not enough to overcome whatever force which is creating the steady state error, but we can overcome this by introducing a new integral gain `I’` which outscales the degenerate force.

This means that the sliding window is an effective way to remove integral windup while still remaining isomorphic to a traditional infinite integral PID

TheMagnetar: honestly no. Windup is totally dependent on your signal and your tuning. PID is totally simple. What you are doing with optional I' are adding complexity to a control loop that is based on simplicity. And windowing? This is just simply bad design where the I part can change abruptly due to a value not having any effect once its windowing is gone. Things need to be continuous as much as possible in control loops: you want to control a non-linear function by applying a control loop that is non-linear?

Call this implementation whatever but a PID. If CBA needs a PID controller, please make it a PID controller that works  based on standard PID controller design (hello Nyquist? Laplace?). Not some implementation hacked using windowing, additional gain parameters and whatever else that contribute to instability and increase lack of predictability

TheCandianVendingMachine: Firstly, no new gains are added with this method. Secondly, if I introduce a robust method of anti-integral windup (back calculation), a new gain would be required as it is.

That being said, the windowing method should perform as well as the infinite sum variation.

Proof:

The PID controller can be formulated as
```math
K_p e\left(t \right) + K_i\int_{0}^{t}e\left(\tau\right)d\tau + K_d \frac{de\left(t\right)}{dt}
```

Now, the proportional and derivative terms are unaffected by the windowing. These will perform just the same. So we look at the integral term as described in the infinite sum case.
```math
\int_{0}^{t}e\left(\tau\right)d\tau
```
more specifically, consider the integral term split into
```math
\int_{0}^{t_0}e\left(\tau\right)d\tau + \int_{t_0}^{t}e\left(\tau\right)d\tau
```

Given $\epsilon \gt 0$, consider the following cases:
Case 1) $|\int_{0}^{t_0}e\left(\tau\right)d\tau| = \epsilon$, and $\epsilon > 0$ near zero:

In this case the first integral can be approximated as zero, $\int_{0}^{t_0}e\left(\tau\right)d\tau \approx 0$. This means that the integral can be approximated as
```math
\left(\int_{0}^{t_0}e\left(\tau\right)d\tau + \int_{t_0}^{t}e\left(\tau\right)d\tau\right)
= \left(\epsilon + \int_{t_0}^{t}e\left(\tau\right)d\tau\right)
\approx \left(0 + \int_{t_0}^{t}e\left(\tau\right)d\tau\right)
\approx \left(\int_{t_0}^{t}e\left(\tau\right)d\tau\right)
```

So given a window of size $t - t_0$, we appropriately approximate the integral due to prior error being negligible to the current state.

$\square$

Case 2) $|\int_{0}^{t_0}e\left(\tau\right)d\tau| > \epsilon$ where $\epsilon \gt 0$:

In this case, the integral term is affecting the plant considerably. This can either be due to integral windup, large steady state error, or a large disturbance occured. There are two sub-cases to consider.

Case 1) $|\int_{t_0}^{t}e\left(\tau\right)d\tau| = \epsilon$ where $\epsilon \gt 0$ near zero:

This case implies that the process variable is near the set point, so $e\left(t\right)$ is small. If this is occurring because $e\left(t\right)$ is approaching zero, that means that we will soon fall into the $\int_{0}^{t_0}e\left(\tau\right)d\tau = \epsilon$ cases as defined above and we will successfully approximate the controller.

Another this can occur is if a large disturbance happened in the past, but was temporary. In this case, the disturbance is not apart of the steady state error. This means that we don't want the integral term to be considering it as such, and the windowing method will quickly disregard it as we return to stability.

Lastly, if there is large steady state error that is not being corrected, tuning $K_i$ to be larger will eliminate this error and bring the plant to the set point.

$\square$

Case 2) $|\int_{t_0}^{t}e\left(\tau\right)d\tau| > \epsilon$ where $\epsilon \gt 0$

This case can occur if the process variable is quickly approaching the setpoint, i.e. $|\frac{de\left(t\right)}{dt}|$ is high. In this case, the non-windowed controller would only have a higher derivative magnitude. The windowed version of the controller will result in a more subdued process variable change, being able to be controlled with the gains as expected.

$\square$

As proven by the above cases, a controller with a window size of $t - t_0$ is a good approximation of a theoretical PID controller.

$\square$

Of course it is an approximation, however all controllers which introduce integral windup prevention methods are also approximations. If we were to introduce conditional integration based on the saturation of the actuators with the `_min` and `_max` parameters, we would be missing parts of the integral. If we were to introduce back calculation, we would both need a new gain to tune as well as introducing a non-continuous sum of the integral. All of these methods have tradeoffs, and I am arguing that the tradeoff I made is no different. Additionally, in hardware provided PID controllers each manufacturer provides a single way that their engineers determined to be the best way to prevent integral windup. These methods have historically been proprietary and locked down.

If you can argue against the proof as shown above, I will return this to a PID controller with conditional integration for integral windup prevention. But with that said, I've been using this PID model for a lot of control simulations and have not run into any shortcomings that a traditional PID model would not also run into.

b-mayr-1984:
To whom it might concern: 

The sample codes needs some minor change in L4 and L5.

> The entire example code is
> 
> ```sqf
> sleep 5;
> ROLL_ACCELERATION = 30;
> 
> rollOffset_pid = [1, 0.3, 0.0, 0, CBA_pid_fnc_error_degree, -180, 180, 16] call CBA_pid_fnc_create;
> rollAcceleration_pid = [1, 0.05, 0.3, 0, CBA_pid_fnc_error_linear, -ROLL_ACCELERATION, ROLL_ACCELERATION, 32] call CBA_pid_fnc_create;
> [..]
> ```

L4 and L5 needs to be changed to this
```sqf
rollOffset_pid = [1, 0.3, 0.0, 0, -180, 180, 16, CBA_pid_fnc_error_degree] call CBA_pid_fnc_create;
rollAcceleration_pid = [1, 0.05, 0.3, 0, -ROLL_ACCELERATION, ROLL_ACCELERATION, 32, CBA_pid_fnc_error_linear] call CBA_pid_fnc_create;
```
in order for it to work in the latest version.

PR #1809, opened 2026-01-18
Port ACE's Status Effects to CBA

DartRuffian:
- Port ACE `isModLoaded`, `endRadioTransmission`, `toBitMask`, `binarizeNumber` functions to CBA.
- Port `runAtSettingsInitialized` to CBA.
- Port ACE's Status Effect system to CBA.
- Respect the [Submitting Content Guidelines](https://github.com/CBATeam/CBA_A3/wiki/Submitting%20content).


addons/common/fnc_binarizeNumber.sqf
```diff
@@ -0,0 +1,49 @@
...
+    _array set [_index, false];
+};
+
+private _index = 0;
+
+while {_number > 0} do {
+    private _rest = _number mod 2;
+    _number = floor (_number / 2);
+
+    _array set [_index, _rest == 1];
```
rautamiekka:
```suggestion
    _array set [_index, _rest isEqualTo 1];
```

DartRuffian:
~~Statuses for `disableNVEquipment`/`disableTIEquipment` would be nice~~ Doesn't work on units like I thought it did

addons/statuseffects/XEH_postInit.sqf
```diff
@@ -0,0 +1,145 @@
...
+}] call CBA_fnc_addEventHandler;
+
+[QGVAR(disableCollision), { // Name "reversed" from `setPhysicsCollisionFlag` because we want NOR logic
+    params ["_object", "_set"];
+    TRACE_2("disableCollision EH",_object,_set);
+    _object setPhysicsCollisionFlag (_set < 1);
+}] call CBA_fnc_addEventHandler;
+
+// TODO: disableNVG/TIEquipment only work on vehicles, but we could make the status effects work for units as well
+// One possible solution would be hooking into VisionModeChanged and using `player action ["NvGogglesOff", player]`
```
DartRuffian, line 134: I've got too many projects currently, but if someone wants to try to implement this, that'd be nice

addons/statuseffects/fnc_addStatusEffectType.sqf
```diff
@@ -0,0 +1,47 @@
...
+---------------------------------------------------------------------------- */
+SCRIPT(addStatusEffectType);
+params [
+    ["_name", "", [""]],
+    ["_isGlobal", false, [false]],
+    ["_commonReasonsArray", [], [[]]],
+    ["_sendJIP", false, [false]],
+    ["_eventName", "", [""]]
+];
+
```
PabstMirror, line 33: Instead of modifying ace to call new CBA, maybe it's best to just have CBA use ace if it exists

```suggestion

if ("ace_common" call CBA_fnc_isModLoaded) exitWith ace_common_fnc_statusEffect_addType;
```

DartRuffian: But then someone using CBA stuff would have it break with ace loaded, unless every status effect function called ace if loaded and that just seems meh

PabstMirror: we can't have two different status effect systems that aren't compatible
so we have to do something like this in either ace or cba
I think it's safest to just leave ace alone and have CBA use ace if it exists in all relevant funcs

DartRuffian: But why not just update the ace ones to call the CBA functions? The only public ones are get/set, and addType should be kept BWC even though its not public. And *if* someone is manually checking the internal variables / calling stuff like sendEffects, respawnEH, etc. then they're just asking for something to break at some point.

IMO ACE should conform to CBA, not the other way around because its the more widely used mod with more than 2x the users (yes not every one of those people are using CBA stuff directly). If CBA calls ACE if its loaded then you risk incompatibilities / different behavior if the ACE functions are updated and someone doesn't make the same changes to CBA as well.

DartRuffian: @PabstMirror update on this?

LinkIsGrim: > IMO ACE should conform to CBA, not the other way around because its the more widely used mod with more than 2x the users (yes not every one of those people are using CBA stuff directly). If CBA calls ACE if its loaded then you risk incompatibilities / different behavior if the ACE functions are updated and someone doesn't make the same changes to CBA as well.

I tend to agree with Dart on this. Also the idea of CBA calling out to something that depends on it is... weird to me.

addons/bitwise/fnc_toBitMask.sqf
```diff
@@ -0,0 +1,30 @@
+#include "script_component.hpp"
+/* ----------------------------------------------------------------------------
+Function: CBA_fnc_toBitMask
+Description:
+    Convert an array of booleans into a number.
+    Takes Input <ARRAY of BOOLs> (limited to float precision)
+
+Parameters:
+    _input - Boolean (least significant bit) <BOOL>
```
PabstMirror:
```suggestion
    _input - Array of bools (0 index is least significant bit) <ARRAY>
```

addons/common/fnc_binarizeNumber.sqf
```diff
@@ -0,0 +1,49 @@
...
+Author:
+    commy2
+---------------------------------------------------------------------------- */
+SCRIPT(binarizeNumber);
+params [
+    ["_number", nil, [0]],
+    ["_minLength", 8, [0]]
+];
+
+if (isNil "_number") exitWith {};
```
PabstMirror: so return value matches
```suggestion
if (isNil "_number") exitWith { [] };
```
or
```suggestion
if (isNil "_number") then { _number = 0 };
```

addons/statuseffects/CfgEventHandlers.hpp
```diff
@@ -0,0 +1,17 @@
...
+    class ADDON {
+        init = QUOTE(call COMPILE_SCRIPT(XEH_postInit));
+    };
+};
+
+class Extended_Engine_EventHandlers {
+    class All {
+        ADDON = QUOTE(call FUNC(handleEngine));
+    };
+};
```
PabstMirror, line 17:
```suggestion
};
class Extended_Respawn_EventHandlers {
    class All {
        class GVAR(statusEffect) {
            respawn = QUOTE(call FUNC(respawnEH));
        };
    };
};

class Extended_Local_EventHandlers {
    class All {
        class GVAR(statusEffect) {
            local = QUOTE(call FUNC(localEH));
        };
    };
};
```

PabstMirror (approved):

PR #1821, opened 2026-07-12
Common - Isolate PFH errors

### LINKED ISSUE #1602: PerFrameHandler Framework Issue: Stuck Handler Prevents Execution of Subsequent Handlers
**Mods (complete and add to the following information):**
N/A

**Description:**
The perFrameHandler framework in CBA_A3 mod for ArmA 3 is causing a problem where, if one of the perFrameHandlers gets stuck or takes too long to complete, the other handlers in the list do not get called. This results in some event handlers never being executed. https://github.com/CBATeam/CBA_A3/blob/master/addons/common/init_perFrameHandler.sqf

**Steps to reproduce:**
Launch ArmA 3 with the CBA_A3 mod enabled.
Join a game where perFrameHandler event handlers are used (can be a custom mission or mod scenario that heavily relies on event handlers).
Introduce a perFrameHandler that has a long execution time or gets stuck (e.g., infinite loop).
Observe that other perFrameHandlers do not get executed.

**Expected behavior:**
Each perFrameHandler should execute independently of the others, so if one is stuck or takes too long, it should not prevent the others from being executed. Not sure how to fix this, maybe by introducing timeouts or at least better info if one of handlers is taking too long.

- Dedicated / Self-Hosted Multiplayer / Singleplayer / Editor (Singleplayer) / Editor (Multiplayer) / Virtual Arsenal

**Log Files:**
N/A

**Additional context:**
The architecture's current design of iterating through a list to call each handler means that if one gets stuck, the others are prevented from being called. This is particularly problematic in complex scenarios or mods where multiple event handlers are crucial for the mission or gameplay.

`    // Execute per frame handlers
    {
        _x params ["_function", "_delay", "_delta", "", "_args", "_handle"];

        if (diag_tickTime > _delta) then {
            _x set [2, _delta + _delay];
            [_args, _handle] call _function;
        };
    } forEach GVAR(perFrameHandlerArray);`

**Screenshots:**

DarkBall123:
- Keep later PFHs and frame queues running when one PFH errors.

used the isNil suggestion from the issue so one broken PFH no longer stops the rest. ScriptError and the stack trace are still reported normally.

tested this with persistent and one-off errors, nested handlers, self-removal, adding and removing handlers during iteration, return values and delayed handlers. later handlers and frame queues kept running after an error, and the largest measured overhead was about 0.16 ms with 500 PFHs.

image

Fixes [#1602](https://github.com/CBATeam/CBA_A3/issues/1602)

### DIFF
--- a/addons/common/init_perFrameHandler.sqf
+++ b/addons/common/init_perFrameHandler.sqf
@@ -33,7 +33,7 @@ GVAR(waitUntilAndExecArray) = [];
 
         if (diag_tickTime > _delta) then {
             _x set [2, _delta + _delay];
-            [_args, _handle] call _function;
+            isNil {[_args, _handle] call _function};
         };
     } forEach GVAR(perFrameHandlerArray);
 


LinkIsGrim (approved):

DartRuffian:
Comment saying why it's there would probably be good

PabstMirror:
What if we modify `CBA_fnc_addPerFrameHandler` to optionally wrap
that way there is no performance penalty for most users 

```
params [["_function", {}, [{}]], ["_delay", 0, [0]], ["_args", []], ["_iWriteShitCode", false, [false]]];

if (_iWriteShitCode) then {
    _function = compile format ["isNil {%1}", toString _function];
};
```

PR #1822, opened 2026-07-14
XEH - Fix 2.22 pylon tank object XEH

PabstMirror:


PabstMirror:
this would be safe to merge before 2.24
it will just add an empty class that inherits from thingx (scope=0)

LinkIsGrim (approved):

PR #1827, opened 2026-07-29
Settings - Refactor settings menu to use a single controls group

LinkIsGrim:
- Some code cleanup.
- Refactor settings menu to use a single controls group per setting. Makes opening menu faster hopefully.
- Due to above switching tabs no longer re-filters/sorts as well.
- Opening an addon no longer walks and sorts every single setting, pulls from an index.
- configClasses lookup is now done exactly once per row type instead of once per row.
- Fixes a few longstanding bugs with the settings UI (particularly in regards to the overwrite checkboxes and their rendering before saving). TL;DR: changing priority or value of a setting would nil out the other in the temp namespace.
- Because it's a single controls group, switching tabs will preserve the scrollbar position as well now.

Requires #1825.

PR #1825, opened 2026-07-29
Settings - Add Searchbar

LinkIsGrim:
**When merged this pull request will:**
- Add searchbar to settings.

image

also does fuzzy matching:

image

PR #1828, opened 2026-07-29
Settings - UI Layout Redesign

LinkIsGrim:
The table gave up a lot of room to things that didn't need it and had no way to get past a long category other than scrolling.

Folding:
- Clicking a sub-category header folds its settings away and back, and the fold holds for as long as the menu is open. Searching opens a folded group if the search found something in it, so a match is never hidden behind one.
- Settings registered without a sub-category come first and had no header, which made them the one block that couldn't be folded. They get a "General" one, but only in categories that have other groups to fold them against.

Room:
- Rows and sub-category headers are shorter, the strip above the table is smaller, and the buttons below it move down into space that was empty. 13 settings fit where 10 did.
- The two overwrite checkboxes sat far apart to keep their labels from touching. The labels are "Clients" and "Mission" under a shared "Overwrites:" now, so the columns can close up and give the width to the setting name instead.

Reading:
- Setting names are aligned left. Right aligned, a short name sat far from the control it belongs to and the left edge never lined up.
- Whether a setting applies reads down the left edge of the table instead of from the middle of a cluster of four icons.
- Rows are banded, so a name can be followed across to its control. Every row class has to name its own background control for this: an inherited one that is never redeclared doesn't get created, and setting a colour on the nothing it returns does nothing.
- Overwritten by a source that holds the same value has its own icon rather than the same one in a different colour, which is not a difference everyone can see. The value is in effect either way, which is what the new one says.

image

Requires #1825, #1827

Fixes:
- Settings would show as yellow (edited) if they were touched at all even if the end value would be the same.
- ctrlShow the overwrite checkboxes instead of moving them off-screen


LinkIsGrim:
This is what the color banding actually looks like now. I don't want to boot the game to grab an ACE Medical screenshot again.

image

PabstMirror:
For me nothing is clickable, drop down menus don't expand and I can't click checkboxes
The setting itself turns grey
and I can manually hit tab to get to and hit space to check or type so the ctrls aren't locked

https://github.com/user-attachments/assets/afa3733a-f975-42ae-bbf3-074fa74ba348

LinkIsGrim:
background's ctrlEnabled

LinkIsGrim:
Added confirmation pop-up on cancelling/hitting escape with live changes:

image

addons/settings/fnc_gui_confirmDiscard.sqf
```diff
@@ -0,0 +1,98 @@
...
+
+if (_hidden > 0) then {
+    _text = _text + format ["<br/> " + LLSTRING(discardChanges_more), _hidden];
+};
+
+(_dialog displayCtrl IDC_CONFIRM_TEXT) ctrlSetStructuredText parseText _text;
+
+// One line per setting, plus the sentence above them. Nothing in a controls group reflows on
+// its own, so the panel and the buttons are moved to match
+private _lines = 1 + (count _shown) + ([0, 1] select (_hidden > 0));
```
PabstMirror: hemtt suggestion
```suggestion
private _lines = 1 + (count _shown) + (parseNumber (_hidden > 0));
```

PabstMirror (approved):

PR #1832, opened 2026-08-08
Common - Add setter for `hideActions`

PabstMirror:
from https://github.com/acemod/ACE3/pull/11459

req 2.22 to function but mergeable now

### DIFF
--- a/addons/common/fnc_hideAction.sqf
+++ b/addons/common/fnc_hideAction.sqf
@@ -1,0 +1,68 @@
+#include "script_component.hpp"
+/* ----------------------------------------------------------------------------
+Function: CBA_fnc_hideAction
+
+Description:
+    Registers or unregisters a hideActions entry for an action index and updates the current hidden state.
+    See https://community.bistudio.com/wiki/shownAction for CfgAction enums.
+
+Parameters:
+    _index        - The action index. <NUMBER>
+    _key          - The key (case-sensitive). <STRING>
+    _hide         - Whether to hide the action. <BOOL>
+    _forceAdd     - Force add (even if hide is false and action is not hidden). (optional, default: false) <BOOL>
+
+Returns:
+    [hiddenIndices, unhiddenIndices] <ARRAY>
+
+Examples:
+    (begin example)
+        [15, "noEngineAction", true] call CBA_fnc_hideAction;
+    (end)
+
+Author:
+    PabstMirror
+---------------------------------------------------------------------------- */
+
+if (!hasInterface || {(productVersion # 2) < 222}) exitWith { [[], []] };
+
+params [["_index", 0, [0]], ["_key", "", [""]], ["_hide", false, [false]], ["_forceAdd", false, [false]]];
+
+private _addEH = if (isNil QGVAR(hideActionHash)) then {
+    GVAR(hideActionHash) = createHashMap;
+    true
+} else {
+    false
+};
+
+if (_hide || _forceAdd || {_index in GVAR(hideActionHash)}) then {
+    private _actionIndex = GVAR(hideActionHash) getOrDefault [_index, createHashMap, true];
+    if (_hide) then {
+        _actionIndex set [_key, true];
+    } else {
+        _actionIndex deleteAt _key;
+    };
+};
+
+private _fnc_update = {
+    private _hideSelected= [];
+    private _unhideSelected = [];
+    {
+        if (count _y == 0) then {
+            _unhideSelected pushBack _x;
+        } else {
+            _hideSelected pushBack _x;
+        };
+    } forEach GVAR(hideActionHash);
+    call compile "
+    hideActions [false, _unhideSelected];
+    hideActions [true, _hideSelected];
+    "; // adds soft 2.22 req, remove after release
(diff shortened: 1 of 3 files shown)


addons/common/fnc_hideAction.sqf
```diff
@@ -0,0 +1,68 @@
...
+Function: CBA_fnc_hideAction
+
+Description:
+    Registers or unregisters a hideActions entry for an action index and updates the current hidden state.
+    See https://community.bistudio.com/wiki/shownAction for CfgAction enums.
+
+Parameters:
+    _index        - The action index. <NUMBER>
+    _key          - The key (case-sensitive). <STRING>
+    _hide         - Whether to hide the action. <BOOL>
```
rautamiekka, line 12:
```suggestion
    _index        - The action index. (Default: 0) <NUMBER>
    _key          - The key (case-sensitive). <STRING>
    _hide         - Whether to hide the action. (Default: false) <BOOL>
```
At 1st was gonna add the default for `_key` too, but realized it's completely pointless to not specify that param, so it's better as a mandatory one; _damned SQF requiring a default value to allow type-checking_ ...

addons/common/fnc_hideAction.sqf
```diff
@@ -0,0 +1,68 @@
...
+params [["_index", 0, [0]], ["_key", "", [""]], ["_hide", false, [false]], ["_forceAdd", false, [false]]];
+
+private _addEH = if (isNil QGVAR(hideActionHash)) then {
+    GVAR(hideActionHash) = createHashMap;
+    true
+} else {
+    false
+};
+
+if (_hide || _forceAdd || {_index in GVAR(hideActionHash)}) then {
```
rautamiekka, line 38: I can't test in any way, but hashmaps are fast and `in` as well, so an `in` check might be faster than a lazy eval.

addons/common/fnc_hideAction.sqf
```diff
@@ -0,0 +1,68 @@
...
+    } else {
+        _actionIndex deleteAt _key;
+    };
+};
+
+private _fnc_update = {
+    private _hideSelected= [];
+    private _unhideSelected = [];
+    {
+        if (count _y == 0) then {
```
rautamiekka, line 51:
```suggestion
        if (count _y isEqualTo 0) then {
```

addons/common/fnc_hideAction.sqf
```diff
@@ -0,0 +1,68 @@
...
+    private _actionIndex = GVAR(hideActionHash) getOrDefault [_index, createHashMap, true];
+    if (_hide) then {
+        _actionIndex set [_key, true];
+    } else {
+        _actionIndex deleteAt _key;
+    };
+};
+
+private _fnc_update = {
+    private _hideSelected= [];
```
rautamiekka, line 48:
```suggestion
    private _hideSelected = [];
```

rautamiekka (changes requested):

addons/common/fnc_hideAction.sqf
johnb432, line 1: Two things:
- I don't quite understand what the purpose of `_forceAdd` is. In what scenarios would you use this?
- Part of current behaviour: Hide an action (action A), then unhide using the function. Hide action A, but using the `hideActions` command. Now, hide/unhide any other action using the function and action A will be visible again. I don't like that behaviour, it feels very imposing and invasive.
Proposed behavior: when you add a key to the index hashmap, it will hide the action (or keep it hidden if there was another key present with same action index). If all keys of a specific index are removed (i.e. index hashmap is empty), the index hashmap will be removed from the action hashmap and the action will be shown again.

PabstMirror: I'm going to rework `_forceAdd` because I also can't think of a good use
but for 2nd point, we have to think about undoing everything that was done

e.g.
Start as playerA
`["autoHover", "mySetting", true] call cba` // update setting
cba does `hideActions true`
switch to UAV
`["autoHover", "mySetting", false] call cba` // update setting
switch back to playerA
- here we need to undo the hideActions that was originally applied to the player

addons/common/fnc_hideAction.sqf
PabstMirror: yeah, it may technically have a default from `params` 
but I don't think it makes sense to list them, at least for `_index` where the default of 0 won't do anything useful

addons/common/fnc_hideAction.sqf
PabstMirror:
```
[15, "noEngineAction", true] call CBA_fnc_hideAction;  // hides
[15, "noEngineAction", false] call CBA_fnc_hideAction; // shows
[15, "noEngineAction", false, true] call CBA_fnc_hideAction; // shows and then removes from hash
```

johnb432: https://github.com/johnb432/No-Actions/blob/master/addons/main/functions/fnc_hideAction.sqf

Here's something I cooked up that should address the scenario you described.

PabstMirror: that would only work if you ran on every single action enum for every player switch

with this PRs current setup we only touch enums we have previously touched, so it should be the most compatible with any other systems

PR #1833, opened 2026-08-08
Main - Add `RELEASE_DEBUG` macros

PabstMirror:
Only run if `<DEBUG_MODE_FULL>` is defined OR `-debug` command line parameter is used.
This has a small runtime cost but is usable in scripts compiled at build-time (sqfc).

Allows more logging for people with released versions of mods.
Shouldn't be used inside loops or performance code but the cost is pretty small

is `DEBUG` too common of a macro and possibly conflict with something?

### DIFF
--- a/addons/main/script_macros_common.hpp
+++ b/addons/main/script_macros_common.hpp
@@ -298,6 +298,39 @@ Author:
 #endif
 
 /* -------------------------------------------
+Macro: DEBUG()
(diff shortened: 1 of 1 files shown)


Dystopian:
Maybe the name should implicitly state that it does work in released versions to prevent misuse, something like `RELEASE_DEBUG`.

PR #1835, opened 2026-08-09
Common - Check PFH delay before destructuring

LinkIsGrim:
- Check tick time before params.

`(_x select 2)` is faster than full `params` and it's readable enough IMO since the lines are close together.

### DIFF
--- a/addons/common/init_perFrameHandler.sqf
+++ b/addons/common/init_perFrameHandler.sqf
@@ -29,9 +29,9 @@ GVAR(waitUntilAndExecArray) = [];
 
     // Execute per frame handlers
     {
-        _x params ["_function", "_delay", "_delta", "", "_args", "_handle"];
+        if (diag_tickTime > (_x select 2)) then {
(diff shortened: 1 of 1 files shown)


PabstMirror:
10x
before (just `params`):
0.00807582 ms
on skip (just `select`):
0.00308647 ms
on hit (`select` and then `params`):
0.0119874 ms

so if it's mostly misses then it's an improvement
but if it's all 0 delay pfehs then it could be worse

but cba+ace shows mostly big delays
cba_common_perFrameHandlerArray apply {_x # 1}
[0.5,0.5,0.1,1,0.25,1,60,0.05,1,1,0]

so I'm pretty sure this will be faster
possibly recommending people just use `EachFrame` to add 0 delay pfeh

LinkIsGrim:
think we can avoid the check for 0-delay entirely

PabstMirror:
I just mean general advice, not something we need to script

PabstMirror (approved):

PR #1836, opened 2026-08-09
Events - Drop dead nil check and wrapper array from dispatch

LinkIsGrim:
- Remove the `isNil "_x"` check from `CALL_EVENT`, it can never be true
- Replace the `[...] param [0, []]` wrapper with a `getVariable` default value
- Skip the event raise in `startLoadingScreen`, `endLoadingScreen` and the 3DEN preview settings reset when it happens before `cba_events` preInit

The `isNil "_x"` lookup was slow and couldn't be true. `param [0, []]` is slower than a `getVariable` default.

`param [0, []]` did guard against a script error if anything ran before `cba_events`'s preInit created the namespace. Can see that in the `OnMissionPreview` and `OnMissionPreviewEnd` 3DEN Event Handlers. But the event raise would've been a no-op anyway, it was just preventing the script error from showing up.

### Performance

`diag_codePerformance`, 100k cycles:

| handlers | before | after | saved |
| --- | --- | --- | --- |
| 0 | 669ns | 566ns | 103ns |
| 1 | 1559ns | 1094ns | 464ns |
| 3 | 3076ns | 2041ns | 1035ns |

The wrapper saving is per raise (~100-150ns), the `isNil` saving is per handler (~285ns), so it scales with listener count. As a share of a full `CBA_fnc_localEvent` call:

| handlers | saved |
| --- | --- |
| 0 | ~6% |
| 1 | ~15% |
| 3 | ~22% |
| many | approaches ~38% |

For scale, ACE3 has 782 raise sites and 491 handler registrations. `ace_firedPlayer` has 9 listeners and fires on every player shot.

### Testing

Editor and mission preview, RPT clean. Checked that undefined events still don't error, handlers still fire in registration order, a handler removing itself mid-dispatch doesn't break iteration, handlers added mid-dispatch don't fire until the next raise, `localEvent` still returns the last handler's value, and both loading screen events still fire at runtime.

### DIFF
--- a/addons/events/script_component.hpp
+++ b/addons/events/script_component.hpp
@@ -37,11 +37,14 @@
 
 #define SEND_TUEVENT_TO_SERVER(params,name,vehicle,turret) TUEVENT_PVAR = [name, params, vehicle, turret]; publicVariableServer TUEVENT_PVAR_STR
 
+// Handlers are never nil: addEventHandler rejects nil and removeEventHandler compacts with
+// deleteAt, so the isNil check could never be true. It is not free either, isNil on a string
+// looks the variable up by name, which measured ~310ns per handler.
+// getVariable supplies the default directly. The [...] param [0, []] this replaces was added to
+// stop undefined events erroring (79857b37), which the default covers, and it cost ~158ns more.
 #define CALL_EVENT(args,event) {\
-    if !(isNil "_x") then {\
-        args call _x;\
-    };\
-} forEach +([GVAR(eventNamespace) getVariable event] param [0, []]) // copy array so events can be removed while iterating safely
+    args call _x;\
+} forEach +(GVAR(eventNamespace) getVariable [event, []]) // copy array so events can be removed while iterating safely
 
 #define GETOBJ(obj) (if (obj isEqualType grpNull) then {leader obj} else {obj})
 
--- a/addons/xeh/fnc_endLoadingScreen.inc.sqf
+++ b/addons/xeh/fnc_endLoadingScreen.inc.sqf
@@ -5,7 +5,11 @@ private _return = call {
 };
 
 isNil {
-    [QGVAR(LoadingScreenEnded), _this] call CBA_fnc_localEvent;
+    // BIS preload calls this before the events addon preInit has made the registry,
+    // and no handler can be registered before that either, so the raise is a no-op
+    if (!isNil QEGVAR(events,eventNamespace)) then {
+        [QGVAR(LoadingScreenEnded), _this] call CBA_fnc_localEvent;
+    };
 };
 
 RETNIL(_return) //scheduler safe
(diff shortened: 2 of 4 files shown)


PabstMirror:
IIRC the isNil was to handle removing from inside an event, like
```
a1 = ["a", {
  systemChat "A1";
  ["a", a2] call CBA_fnc_removeEventHandler; // remove 2 from inside 1
}] call cba_fnc_addEventHandler;

a2 = ["a", {
  systemChat "A2";
}] call cba_fnc_addEventHandler;

a3 = ["a", {
  systemChat "A3";
}] call cba_fnc_addEventHandler;

["a", 5] call CBA_fnc_localEvent
```

but we copy the array first, which automatically handles the problem with iterating though a modified array
(technically 2 still fires even if removed in 1, but that keeps things simple)

PabstMirror (approved):

addons/xeh/fnc_endLoadingScreen.inc.sqf
```diff
@@ -5,7 +5,11 @@ private _return = call {
 };
 
 isNil {
-    [QGVAR(LoadingScreenEnded), _this] call CBA_fnc_localEvent;
+    // BIS preload calls this before the events addon preInit has made the registry,
+    // and no handler can be registered before that either, so the raise is a no-op
+    if (!isNil QEGVAR(events,eventNamespace)) then {
+        [QGVAR(LoadingScreenEnded), _this] call CBA_fnc_localEvent;
```
PabstMirror, line 11: Would this be any better?
I'm guessing nothing else would have added events, so it probably doesn't matter

```
    if (isNil QEGVAR(events,eventNamespace)) then {
        _this spawn {
            QGVAR(LoadingScreenEnded) call CBA_fnc_localEvent;    
        };
    } else {
        QGVAR(LoadingScreenEnded) call CBA_fnc_localEvent;
    };
```

addons/events/script_component.hpp
```diff
@@ -37,13 +37,16 @@
 
 #define SEND_TUEVENT_TO_SERVER(params,name,vehicle,turret) TUEVENT_PVAR = [name, params, vehicle, turret]; publicVariableServer TUEVENT_PVAR_STR
 
+// Handlers are never nil: addEventHandler rejects nil and removeEventHandler compacts with
+// deleteAt, so the isNil check could never be true. It is not free either, isNil on a string
+// looks the variable up by name, which measured ~310ns per handler.
+// getVariable supplies the default directly. The [...] param [0, []] this replaces was added to
+// stop undefined events erroring (79857b37), which the default covers, and it cost ~158ns more.
```
PabstMirror, line 44: I guess these comments are more for the PR and won't really make sense after merge
but it's not really a problem

PabstMirror (approved):

addons/events/script_component.hpp
rautamiekka: > I guess these comments are more for the PR and won't really make sense after merge but it's not really a problem

I argue it's the total opposite, these are very useful to explain why it's done this way.

LinkIsGrim: I can clean them up if you'd prefer.

PR #1837, opened 2026-08-09
UI - Add setting to auto-expand the Options menu

Dystopian:
- title

### DIFF
--- a/addons/ui/initSettings.inc.sqf
+++ b/addons/ui/initSettings.inc.sqf
@@ -26,3 +26,12 @@
     [1, 10, 4, 1], // default value
     2 // global
 ] call CBA_fnc_addSetting;
+
+[
+    QGVAR(autoExpandOptions),
+    "CHECKBOX",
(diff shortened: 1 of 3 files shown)


PabstMirror (approved):
tested with ace and rhs
ace adds some new options, still showed up fine
rhs adds a button above on the display, showed up fine

addons/ui/initSettings.inc.sqf
```diff
@@ -26,3 +26,12 @@
     [1, 10, 4, 1], // default value
     2 // global
 ] call CBA_fnc_addSetting;
+
+[
+    QGVAR(autoExpandOptions),
+    "CHECKBOX",
+    LLSTRING(AutoExpandOptions),
+    [LELSTRING(main,DisplayName), LLSTRING(Category)],
+    false,
```
PabstMirror: Could be true imho

Dystopian: Can also get rid of setting, imho it is not needed.

PR #1841, opened 2026-08-12
Various - Update headers

PabstMirror:


addons/statemachine/fnc_addEventTransition.sqf
```diff
@@ -39,7 +36,7 @@ params [
     ["_originalState", "", [""]],
     ["_targetState", "", [""]],
     ["_events", [], [[]]],
-    ["_condition", {}, [{}, ""]],
+    ["_condition", {}, [{}]],
```
PabstMirror, line 39: `_condition` has to be code

addons/arrays/fnc_getArrayDiff.sqf
```diff
@@ -6,15 +6,17 @@ Description:
     A function used to return the differences between two arrays.
 
 Parameters:
-    Two Arrays of strings (must not contain scalars)
```
PabstMirror, line 9: but the example uses scalars?

PR #1844, opened 2026-08-15
UI - Add map preview to Create MP Mission dialog

Dystopian:
- title.
s50

### DIFF
--- a/addons/ui/fnc_initDisplayRemoteMissions.sqf
+++ b/addons/ui/fnc_initDisplayRemoteMissions.sqf
@@ -25,18 +25,60 @@ private _fnc_findMissions = {
 
 _display setVariable [QGVAR(stockMissions), _stockMissions];
 
-// Show worldnames as tooltips on map list
 private _worldNames = createHashMap;
+private _worldFeatures = createHashMap;
+private _defaultPicture = getText (configFile >> "display3DENNew" >> "defaultPicture");
+private _defaultAuthor = localize "STR_AUTHOR_UNKNOWN";
+private _authorFullLocalized = localize "STR_FORMAT_AUTHOR_SCRIPTED";
 {
     private _worldName = configName _x;
-    private _description = getText (configFile >> "CfgWorlds" >> _worldName >> "description");
+    private _worldConfig = configFile >> "CfgWorlds" >> _worldName;
+
+    private _description = getText (_worldConfig >> "description");
     _worldNames set [_description, _worldname];
+
+    private _pictureMap = getText (_worldConfig >> "pictureMap");
+    if (_pictureMap == "") then {_pictureMap = _defaultPicture}; // can be empty
+    private _author = getText (_worldConfig >> "author");
+    if (_author == "") then {_author = _defaultAuthor};
+    _author = format [_authorFullLocalized, _author];
+
+    _worldFeatures set [_worldName, [_pictureMap, _author]];
 } forEach (configProperties [configFile >> "CfgWorldList", "isClass _x"]);
+
+// Show worldnames as tooltips on map list
 for "_index" from 0 to ((lbSize _ctrlMaps) - 1) do {
     private _description = _ctrlMaps lbText _index;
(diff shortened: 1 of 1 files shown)


addons/ui/fnc_initDisplayRemoteMissions.sqf
```diff
@@ -25,19 +25,61 @@ private _fnc_findMissions = {
...
 for "_index" from 0 to ((lbSize _ctrlMaps) - 1) do {
     private _description = _ctrlMaps lbText _index;
     private _worldName = _worldNames getOrDefault [_description, ""];
     _ctrlMaps lbSetTooltip [_index, _worldName];
 };
 
+_ctrlMaps setVariable [QGVAR(worldFeatures), _worldFeatures];
+
+ctrlPosition _ctrlMaps params ["_mapsLeft", "_mapsTop", "_mapsWidth", "_mapsHeight"];
+private _squareWidth = _mapsHeight * 3 / 4;
```
Dystopian, line 59: Please check this expression, because I completely do not understand how to correctly calculate it,
`3/4` works for me, as well as `safeZoneH / safeZoneW`.

rautamiekka: I wonder if the `3/4` refers to the 4:3 aspect ratio, which, as I understand it, `safeZone{H,W}` were designed to help abstract away.

addons/ui/fnc_initDisplayRemoteMissions.sqf
```diff
@@ -25,19 +25,61 @@ private _fnc_findMissions = {
...
 {
     private _worldName = configName _x;
-    private _description = getText (configFile >> "CfgWorlds" >> _worldName >> "description");
+    private _worldConfig = configFile >> "CfgWorlds" >> _worldName;
+
+    private _description = getText (_worldConfig >> "description");
     _worldNames set [_description, _worldname];
+
+    private _pictureMap = getText (_worldConfig >> "pictureMap");
+    if (_pictureMap == "") then {_pictureMap = _defaultPicture}; // can be empty
```
rautamiekka:
```suggestion
    if (_pictureMap isEqualTo "") then {_pictureMap = _defaultPicture}; // can be empty
```

addons/ui/fnc_initDisplayRemoteMissions.sqf
```diff
@@ -25,19 +25,61 @@ private _fnc_findMissions = {
...
+    private _author = getText (_worldConfig >> "author");
+    if (_author == "") then {_author = _defaultAuthor};
```
rautamiekka:
```suggestion
    if (_author isEqualTo "") then {_author = _defaultAuthor};
```

rautamiekka (changes requested):

addons/ui/fnc_initDisplayRemoteMissions.sqf
```diff
@@ -25,19 +25,61 @@ private _fnc_findMissions = {
...
     private _description = _ctrlMaps lbText _index;
     private _worldName = _worldNames getOrDefault [_description, ""];
     _ctrlMaps lbSetTooltip [_index, _worldName];
 };
 
+_ctrlMaps setVariable [QGVAR(worldFeatures), _worldFeatures];
+
+ctrlPosition _ctrlMaps params ["_mapsLeft", "_mapsTop", "_mapsWidth", "_mapsHeight"];
+private _squareWidth = _mapsHeight * 3 / 4;
+private _mapsWidthNew = _mapsWidth - _squareWidth;
```
PabstMirror, line 60: On square-ish aspect ratios this could possibly start to get cut-off
but I really don't think it will be a problem

we could do something like
`if (_squareWidth > 0.5 * _mapsWidth) exitWith {}`

to prevent the pic from taking too much of the horizontal,
but like I said I don't really think we need it unless someone has a problem

PabstMirror (approved):

PR #1846, opened 2026-08-17
UI - Add setting to hide world suffix from missions in Create MP Mission dialog

Dystopian:
- title.

### World + Missions:
overall

### MP Host before/after
MP Host off
MP Host on

### Dedicated before/after
dedi def off
dedi def on

### Dedicated with `SkipDescriptionParsing` before/after
dedi skip off
dedi skip on

### DIFF
--- a/addons/ui/fnc_initDisplayRemoteMissions.sqf
+++ b/addons/ui/fnc_initDisplayRemoteMissions.sqf
@@ -102,12 +102,19 @@ private _fnc_storeMapMissions = {_this spawn {isNil { // delay a frame
     params ["_ctrlMaps"];
     private _display = ctrlParent _ctrlMaps;
     private _ctrlMissions = _display displayCtrl IDC_SERVER_MISSION;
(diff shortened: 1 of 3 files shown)


BrettMayson:
I don't see the reason why though

PabstMirror:
yeah I have to agree

e.g. there are 3 versions of saharni, 
so seeing full filename lets you make sure you have selected the right mission on the right map

Dystopian:
I hope setting (disabled by default) suits more.

Dystopian:
Just to add, in case it matters: the selected world is always shown next to the mission list (to the left), so I find duplicating it in the mission name unnecessary. I realize some people may prefer to have it there, which is why I added a setting.

BrettMayson:
Personally I'd rather see us either accept or reject this, I don't think a setting is needed

Can you provide screenshots of the before and after?

Dystopian:
Added screenshots to the first post.

PR #1848, opened 2026-08-19
Common - cba_fnc_getCfgDataHashmap: support for subclasses

OverlordZorn:
Previously, a subclass would return with `any` as its value. With this fix they will return a further hashmap of its content.

### DIFF
--- a/addons/common/fnc_getCfgDataHashmap.sqf
+++ b/addons/common/fnc_getCfgDataHashmap.sqf
@@ -4,6 +4,7 @@ Function: CBA_fnc_getCfgDataHashmap
 
 Description:
     This function extracts all config properties of a class and returns them as a Hashmap with the properties as keys (string).
+    SubClasses will be returned as their own Hashmap.
 
     The following Values will be converted:
     Boolean as String ("true", "false") will be converted to boolean data-type.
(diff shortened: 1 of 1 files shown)


PabstMirror (approved):

addons/common/fnc_getCfgDataHashmap.sqf
```diff
@@ -76,10 +75,16 @@ private _convertCall = [
...
 
 
+private _returnHashMap = createHashMap;
 {
     private _cfg = _x;
     private _key = if (_toLower) then { toLowerANSI configName _cfg } else { configName _cfg };
-    private _value = _cfg call BIS_fnc_getCfgData call _convertCall;
+
+    private _value = if ( isClass _cfg ) then {
+        [_cfg, _condition, _inherit, _convert, _toLower] call FUNC(getCfgDataHashmap)
```
PabstMirror: warning[L-S29]: Undefined Function: cba_common_fnc_getcfgdatahashmap
   ┌─ addons/common/fnc_getCfgDataHashmap.sqf:84:63
   │
84 │         [_cfg, _condition, _inherit, _convert, _toLower] call FUNC(getCfgDataHashmap)
   │                                                               ^^^^^^^^^^^^^^^^^^^^^^^ undefined function

OverlordZorn: removed the macro again and it works without on my side

OverlordZorn: grafik

PR #1851, opened 2026-08-23
Settings - Fix Settings UI checkbox display, code cleanup

LinkIsGrim:
- Replace #1839.
- Fix the greyed out "Mission" overwrite checkboxes appearing on the Mission and Client tabs after changing category, by putting the checkboxes back after the row's controls group is shown, rather than by skipping the `ctrlShow`.
- Handle other edge cases:
  - Clearing the search bar, or unfolding a sub-category, on the Mission or Client tab. Both bring a row back with `ctrlShow`, which shows every control inside it again, so the checkboxes the source doesn't use reappeared. 
  - Toggling "Configure Base Game" and back to "Configure Addons" from the vanilla options menu. `ctrlShow true` on the addons group shows every category built inside it, so all of them were drawn on top of each other, and every row a search or a folded sub-category had hidden came back. Only the selected category is shown now, and its rows are re-filtered once the source is known again.
  - The "Overwrite mission" checkbox enabling itself on a row that can't be edited. `gui_setOverwriteVisible` used to `ctrlEnable _show` unconditionally, undoing `gui_setRowEnabled` for a setting locked by userconfig, the mission or the server config. It now only ever takes the enable away, so it can't fight `gui_setRowEnabled` or the row's `updateUI_priority` when it is re-applied.

- Cleanup:
  - `gui_setOverwriteVisible` takes the row instead of a single checkbox and works both out from the row's own source and `isGlobal`, so it is repeatable and there is one place that knows which checkboxes a source uses. Removes the duplicated condition in `gui_retargetRow`.
  - `gui_addonChanged` iterates the options groups hashmap directly instead of `keys` + `get`.
  - Trailing whitespace in `fnc_check.sqf` and `fnc_gui_addonChanged.sqf`.

### DIFF
--- a/addons/settings/fnc_check.sqf
+++ b/addons/settings/fnc_check.sqf
@@ -42,7 +42,7 @@ switch (toUpper _settingType) do {
     };
     case "TIME": {
         _settingData params ["_min", "_max"];
-        _value isEqualType 0 && {_value >= _min} && {_value <= _max} && {round _value == _value} 
+        _value isEqualType 0 && {_value >= _min} && {_value <= _max} && {round _value == _value}
     };
     default {false};
 };
--- a/addons/settings/fnc_gui_setOverwriteVisible.sqf
+++ b/addons/settings/fnc_gui_setOverwriteVisible.sqf
@@ -27,13 +27,13 @@ Author:
 
 params ["_controlsGroup"];
 
-// a global setting always overwrites the clients and a local one never leaves
-// the client it is set on, neither of them has anything to point anywhere
-private _isGlobal = _controlsGroup getVariable [QGVAR(isGlobal), 0];
+// a local setting never leaves the client it is set on, so it has nothing to
+// point anywhere
+private _isLocalOnly = ROW_IS_LOCAL_ONLY(_controlsGroup);
 private _source = ROW_SOURCE(_controlsGroup);
 
-private _showClient = _source isNotEqualTo "client" && _isGlobal < 2;
-private _showMission = _source isEqualTo "server" && _isGlobal < 2;
+private _showClient = _source isNotEqualTo "client" && !_isLocalOnly;
+private _showMission = _source isEqualTo "server" && !_isLocalOnly;
 
 private _ctrlOverwriteClient = _controlsGroup controlsGroupCtrl IDC_SETTING_OVERWRITE_CLIENT;
 private _ctrlOverwriteMission = _controlsGroup controlsGroupCtrl IDC_SETTING_OVERWRITE_MISSION;
--- a/addons/settings/fnc_gui_setOverwriteVisible.sqf
+++ b/addons/settings/fnc_gui_setOverwriteVisible.sqf
@@ -42,7 +42,7 @@ _ctrlOverwriteClient ctrlShow _showClient;
 _ctrlOverwriteMission ctrlShow _showMission;
 
 // a checkbox that isn't there can't be ticked either. Whether the ones that are
-// can be is not this function's to answer, so it only ever takes that away -
+// can be is not this function's task to answer, so it only ever takes that away -
 // FUNC(gui_setRowEnabled) and the row's updateUI_priority own the other half.
 _ctrlOverwriteClient ctrlEnable (_showClient && ctrlEnabled _ctrlOverwriteClient);
 _ctrlOverwriteMission ctrlEnable (_showMission && ctrlEnabled _ctrlOverwriteMission);
--- a/addons/settings/fnc_gui_setOverwriteVisible.sqf
+++ b/addons/settings/fnc_gui_setOverwriteVisible.sqf
@@ -3,28 +3,46 @@
 Internal Function: CBA_settings_fnc_gui_setOverwriteVisible
 
 Description:
-    Shows or hides one of the "overwrite" checkboxes of a settings menu row.
+    Shows the "overwrite" checkboxes that the source a settings menu row is
+    pointed at can use, and hides the rest.
 
-    Which of them a row has depends on the source it is showing, so this has to
-    be reversible.
+    Which of them a row has depends on that source, so this has to be reversible.
+    It also has to be repeatable: showing a row shows every control inside it,
+    the hidden checkboxes included, so whoever shows one puts them back.
 
 Parameters:
-    _ctrl - Overwrite checkbox <CONTROL>
-    _show - Show the checkbox <BOOL>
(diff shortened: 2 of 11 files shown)


PabstMirror (approved):

addons/settings/fnc_gui_setOverwriteVisible.sqf
```diff
@@ -3,28 +3,46 @@
...
+private _showMission = _source isEqualTo "server" && _isGlobal < 2;
+
+private _ctrlOverwriteClient = _controlsGroup controlsGroupCtrl IDC_SETTING_OVERWRITE_CLIENT;
+private _ctrlOverwriteMission = _controlsGroup controlsGroupCtrl IDC_SETTING_OVERWRITE_MISSION;
+
+_ctrlOverwriteClient ctrlShow _showClient;
+_ctrlOverwriteMission ctrlShow _showMission;
+
+// a checkbox that isn't there can't be ticked either. Whether the ones that are
+// can be is not this function's to answer, so it only ever takes that away -
```
PabstMirror:
```suggestion
// can be is not this function's task to answer, so it only ever takes that away -
```

addons/settings/fnc_gui_settingOverwrite.sqf
```diff
@@ -29,7 +29,9 @@ _ctrlOverwriteClient setVariable [QFUNC(event), {
 _controlsGroup setVariable [QFUNC(auto_check_overwrite), {
     params ["_controlsGroup", "_source"];
 
-    if (_source isEqualTo "mission") then {
+    // a local only setting has no "overwrite clients" to tick, and a priority
+    // ticked in here would be sanitized away again the moment it is saved
+    if (_source isEqualTo "mission" && !ROW_IS_LOCAL_ONLY(_controlsGroup)) then {
```
PabstMirror: hemtt complains about `help[L-S19]: Unneeded Not`


```suggestion
    private _isLocalOnly = ROW_IS_LOCAL_ONLY(_controlsGroup);
    if (_source isEqualTo "mission" && !_isLocalOnly) then {
```

addons/settings/fnc_check.sqf
```diff
@@ -42,7 +42,7 @@ switch (toUpper _settingType) do {
     };
     case "TIME": {
         _settingData params ["_min", "_max"];
-        _value isEqualType 0 && {_value >= _min} && {_value <= _max} && {round _value == _value} 
+        _value isEqualType 0 && {_value >= _min} && {_value <= _max} && {round _value == _value}
```
rautamiekka, line 45:
```suggestion
        _value isEqualType 0 && {_value >= _min} && {_value <= _max} && {round _value isEqualTo _value}
```

addons/settings/fnc_gui_setRowEnabled.sqf
```diff
@@ -30,9 +30,14 @@ params ["_controlsGroup"];
...
 
+// a local setting is never overwritten by the mission, so there is nothing a
+// mission maker could set from here. The server keeps it: a server is a client
+// too, and FUNC(set) writes both when it is set there.
+private _isLocalOnly = ROW_IS_LOCAL_ONLY(_controlsGroup);
+
 private _enabled = switch (_source) do {
     case "client": {CAN_SET_CLIENT_SETTINGS && {isNil {GVAR(userconfig) getVariable _setting}}};
-    case "mission": {CAN_SET_MISSION_SETTINGS && {isNil {GVAR(missionConfig) getVariable _setting}}};
+    case "mission": {CAN_SET_MISSION_SETTINGS && !_isLocalOnly && {isNil {GVAR(missionConfig) getVariable _setting}}};
```
rautamiekka, line 40:
```suggestion
    case "mission": {CAN_SET_MISSION_SETTINGS && !_isLocalOnly && {GVAR(missionConfig) isNil _setting}};
```
I do see the others use the older way.

addons/settings/script_component.hpp
```diff
@@ -237,8 +237,18 @@
...
+// always overwritten for everyone, LOCAL_ONLY can't be overwritten at all.
+#define SETTING_LOCAL_OVERRIDABLE 0
+#define SETTING_GLOBAL_ONLY 1
+#define SETTING_LOCAL_ONLY 2
+
+#define IS_GLOBAL_SETTING(setting) (GVAR(default) getVariable [setting, []] param [7, 0] == SETTING_GLOBAL_ONLY)
+#define IS_LOCAL_SETTING(setting)  (GVAR(default) getVariable [setting, []] param [7, 0] == SETTING_LOCAL_ONLY)
+
+// the same question asked of a row, which keeps its setting's _isGlobal
+#define ROW_IS_LOCAL_ONLY(group) ((group getVariable [ARR_2(QGVAR(isGlobal),SETTING_LOCAL_OVERRIDABLE)]) == SETTING_LOCAL_ONLY)
```
rautamiekka, line 251:
```suggestion
#define IS_GLOBAL_SETTING(setting) (GVAR(default) getVariable [setting, []] param [7, 0] isEqualTo SETTING_GLOBAL_ONLY)
#define IS_LOCAL_SETTING(setting)  (GVAR(default) getVariable [setting, []] param [7, 0] isEqualTo SETTING_LOCAL_ONLY)

// the same question asked of a row, which keeps its setting's _isGlobal
#define ROW_IS_LOCAL_ONLY(group) ((group getVariable [ARR_2(QGVAR(isGlobal),SETTING_LOCAL_OVERRIDABLE)]) isEqualTo SETTING_LOCAL_ONLY)
```

addons/settings/test_gui.sqf
```diff
@@ -0,0 +1,229 @@
...
+
+// ----- folding a sub-category away hides its rows, unfolding brings them back
+[_ctrlAddonList, _testIndices param [0, 0]] call FUNC(gui_addonChanged);
+
+private _category = uiNamespace getVariable [QGVAR(addon), ""];
+private _ctrlOptionsGroup = (_display getVariable [QGVAR(optionsGroups), createHashMap]) getOrDefault [_category, controlNull];
+private _rowOrder = _ctrlOptionsGroup getVariable [QGVAR(rowOrder), []];
+private _headerIndex = _rowOrder findIf {!isNil {_x getVariable QGVAR(members)}};
+
+if (_headerIndex == -1) then {
```
rautamiekka, line 181:
```suggestion
if (_headerIndex isEqualTo -1) then {
```

rautamiekka (changes requested):

PR #1854, opened 2026-08-27
Common - Add alive check to `turretDir`

Mike-MF:
- When using OCAP, it runs `CBA_fnc_turretDir` on a loop using `vehicles` as it's forEach, the function works fine for alive vehicles, but with dead it will always return:
> [CBA] (common) WARNING: Vehicle I_APC_Wheeled_03_cannon_F has invalid gun configs on turret [0]

### DIFF
--- a/addons/common/fnc_turretDir.sqf
+++ b/addons/common/fnc_turretDir.sqf
@@ -27,6 +27,8 @@ SCRIPT(turretDir);
 
 params [["_vehicle", objNull, [objNull]], ["_turret", [-1], [[]]], ["_relativeToModel", false, [false]]];
 
+if (!alive _vehicle) exitWith {};
+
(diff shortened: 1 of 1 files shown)


addons/common/fnc_turretDir.sqf
```diff
@@ -27,6 +27,8 @@ SCRIPT(turretDir);
 
 params [["_vehicle", objNull, [objNull]], ["_turret", [-1], [[]]], ["_relativeToModel", false, [false]]];
 
+if (!alive _vehicle) exitWith {};
```
PabstMirror:
```suggestion
if (!alive _vehicle) exitWith { [0,0] };
```

PabstMirror (approved):

PR #1856, opened 2026-09-09
Events - Restore using `param` in event calls

PabstMirror:
Partial revert #1836 
Handles event being called too early in preInit
This bug breaks XEH chain

```
Debug: Local event [event = cba_keybinding_registerKeybind],  [cba_events_eventNamespace = any]
Error in expression <ntName, []])];
{    _params call _x;} forEach +(cba_events_eventNamespace getV>
 Error position: <forEach +(cba_events_eventNamespace getV>
 Error foreach: Type Number,Not a Number, expected Array,HashMap
File x\cba\addons\events\fnc_localEvent.sqf..., line 31
 âž¥ Context: 	[] L6 (x\cba\addons\xeh\fnc_initDisplay3DEN.sqf)
	[] L8 (x\cba\addons\xeh\fnc_initDisplay3DEN.sqf)
	[] L160 (x\cba\addons\xeh\fnc_preInit.sqf)
	[] L119 (x\cba\addons\xeh\fnc_preInit.sqf)
	[] L120 (x\cba\addons\xeh\fnc_preInit.sqf)
	[] L121 (x\cba\addons\xeh\fnc_preInit.sqf)
	[] L124 (x\cba\addons\xeh\fnc_preInit.sqf)
	[] L1 ()
	[] L5 (WBK_Zombies\bootstrap\XEH_preInit.sqf)
	[] L195 (x\cba\addons\keybinding\fnc_addKeybind.sqf)
	[] L31 (x\cba\addons\events\fnc_localEvent.sqf)
```

The `Type Number` error is confusing but can be replicated with `{} forEach +nil`

PR #1857, opened 2026-09-09
Events - Guard raises made before the events addon preInit

LinkIsGrim:
CBA_fnc_addKeybind and CBA_fnc_addSetting can both be called from another addon's own preInit, which is not guaranteed to run after the events addon's preInit has created the event registry. When that registry is nil, CALL_EVENT's forEach +(getVariable [event, []]) does not fall back to [] (the default only applies when the variable is missing, not when the namespace itself is nil), and the raise crashes with a 'Not a Number, expected Array,HashMap' error.

Guard both raises the same way the three existing early-raise sites already do.

### DIFF
--- a/addons/keybinding/fnc_addKeybind.sqf
+++ b/addons/keybinding/fnc_addKeybind.sqf
@@ -194,6 +194,7 @@ GVAR(actions) setVariable [_action, [_displayName, _tooltip, _keybinds, _default
 // this can run before the events addon preInit has made the registry; no handler can be
 // listening yet either, so the raise is a no-op
 if (!isNil QEGVAR(events,eventNamespace)) then {
+    // Emit an event that a key has been registered.
     [QGVAR(registerKeybind), _this] call CBA_fnc_localEvent;
 };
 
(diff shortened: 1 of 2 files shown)


PabstMirror (approved):

addons/keybinding/fnc_addKeybind.sqf
```diff
@@ -191,7 +191,10 @@ GVAR(actions) setVariable [_action, [_displayName, _tooltip, _keybinds, _default
     };
 } forEach _keybinds;
 
-// Emit an event that a key has been registered.
-[QGVAR(registerKeybind), _this] call CBA_fnc_localEvent;
+// this can run before the events addon preInit has made the registry; no handler can be
+// listening yet either, so the raise is a no-op
+if (!isNil QEGVAR(events,eventNamespace)) then {
+    [QGVAR(registerKeybind), _this] call CBA_fnc_localEvent;
```
PabstMirror, line 198: i guess keep original comment
```suggestion
    // Emit an event that a key has been registered.
    [QGVAR(registerKeybind), _this] call CBA_fnc_localEvent;
```

LinkIsGrim:
Prefer https://github.com/CBATeam/CBA_A3/pull/1856

PabstMirror:
there is no harm in also merging this
could add warning like `Add CBA to required addons or call in postInit` but probably not needed

rautamiekka:
> could add warning like `Add CBA to required addons or call in postInit` but probably not needed

Sounds like a pretty damn important piece of documentation to me ...

